PR #31, opened 2024-03-06
Login error updated

arashveysi3:
added basic information on wrong credential entrance.


jelveh:
Hi there, thank you for the contribution. 
Quick question, auth errors are received from the backend and printed in `login-error-msg`, what is the reasoning behind the default error message?

PR #53, opened 2024-03-08
style: fix tooltip shadow on taskBarItem

LynanBreeze:
Fixed the shadow style of tooltip popover on taskBarItem, mostly about not showing shadow on the arrow, also removed the border of tooltip body to make it looks more natural with an arrow.

Before: 
before

After: 
after


KernelDeimos:
Tested this and everything looks good!

jelveh:
Beautiful. This was bothering me for a long time.

PR #69, opened 2024-03-10
delaying progress window for quick move/copy operations

vineeth-vk11:
Issues covered - #6 
1. Updated code to show progress dialogue only if copy/move operation takes more than 2 seconds
2. Increased copy_progress_hide_delay to 2000 so that if the operation takes more than 2 seconds, the progress dialogue will be shown for at least 2 seconds


jelveh:
Looks great, Vineeth! Thank you for your second contribution.

PR #70, opened 2024-03-10
Issue #5 -Fixes direct upload on trash

8577shivam:
#5


jelveh:
Trash must be checked using the destination path of the item not Window title.

PR #71, opened 2024-03-10
Adding ability to undo actions

vineeth-vk11:
### Task
Add ability to undo below actions via context menu or ctrl/cmd + z
- [x] Copy
- [x] Move
- [x] Delete
- [x] Rename
- [x] New file/folder
- [x] Upload


vineeth-vk11:
@jelveh I have implemented undo for the create file/folder action and rename action. Looking for validation on the approach to proceed with implementation for other actions.

jelveh:
Thank you @vineeth-vk11 ! this seems to be going in the right direction, let's continue. I'll do more, deep manual testing along the way as you add more actions. 💪

jelveh:
Thanks for letting me know. I can take care of refactoring the copy/move funcs. I need some time to look through the code though. I'll update you soon here. In the meantime let me know if there are other issues blocking dev on this.

vineeth-vk11:
@jelveh other than that one case which needs to be looked into is the over write case when files are being copied or moved, I don't think the old file is available in the file system after overwriting, so how we implement undo for this?

jelveh:
This is a good question. I take blame for not writing up a good issue for such an important feature request! I just added [CONTRIBUTING.md](https://github.com/HeyPuter/puter/blob/master/CONTRIBUTING.md) and I'm gonna follow it just like everybody else.

Anyway, I just checked how Mac OS handles this. If you move/copy file to a new location and overwrite an existing file with the same name, once you undo, the new file is put back in the original location and the overwritten file is gone forever i.e. simply perma-deleted. This makes sense, because there are potential security issues with bringing back an overwritten file. The security issue could be caused both by implementation and by the users taking advantage of it. So let's go with the Mac OS model.

Also, I haven't forgotten about this PR, but I need to finish a few things before I get back to it. Please continue and let me know if anything comes up, I'll get to this soon.

vineeth-vk11:
@jelveh makes sense. I will keep working on this. I will also try writing alternate functions for copy and move, which work on path rather than dom elements. Let's see where it goes.

jelveh:
Thank you. Sounds good. You could refactor the functions, commit them, and I'll review it with you. Or if there are issues I'll help resolve them together. 💪

jelveh:
This is coming along very nicely! Great job @vineeth-vk11 💪

jelveh:
@vineeth-vk11 🔥🔥🔥 the big PR is here 💪
testing...

vineeth-vk11:
@jelveh, the draft implementation for undoing all actions is complete. Could you do a review once when you get time? I want to spend a little more time testing different scenarios.

Also, I did not have to refactor any function. I found there was already a way in the code to create the HTML element from options in UIItem.js, so I created a function in helper.js that creates the HTML element and passes it to the existing functions.

jelveh:
Thank you for this awesome PR @vineeth-vk11 many people will benefit from this feature!
I couldn't find any issues so far, it seems to be working exactly as expected. I'm going to do some more manual testing, but seems ready for merge 🔥

jelveh:
@vineeth-vk11 I think you forgot to commit the function `get_html_from_options` , am I right?

src/helpers.js
```diff
@@ -3313,3 +3380,314 @@ window.unzipItem = async function(itemPath) {
...
+window.undo_delete = async(items)=>{
+    for (const item of items) {
+        const el = await get_html_from_options(item.options);
+        let metadata = $(el).attr('data-metadata') === '' ? {} : JSON.parse($(el).attr('data-metadata'))
+        move_items([el], path.dirname(metadata.original_path), true);
+    }
+}
+
+
+window.get_html_element_from_options = async function(options){
```
vineeth-vk11, line 3570: @jelveh this is the function

vineeth-vk11: My bad. I had renamed it at the last moment but forgot to rename the references. I have made a commit to fix it.

jelveh: All fixed now, thank you.

jelveh:
Great work @vineeth-vk11 , a very useful feature that Puter needed for a long time. Thanks to you many people will enjoy it now.

vineeth-vk11:
Thanks, @jelveh; I am glad I can contribute. 

On a different note, please let me know which formatter you use to format the code. I am using prettier, and the formatting is creating too many changes.

jelveh:
I'm using VSCode's default formatter for JS. I'll take a look to see what happens if I move to prettier. Some other devs have suggested I give it a try as well.

PR #88, opened 2024-03-13
Implement env support for PORT property

### LINKED ISSUE #86: Use environment variables for setting the dev server port
title.
I was thinking obviously we can pull from process.env but it's also nice when it automatically updates from the .env file - this would mean we use the dotenv module

stretch07:
This includes .env file-defined environment variables.
Also added a .env.example file if people want to see how it's done.

Closes #86


jelveh:
Works like a charm! Thank you for the your contribution.

PR #99, opened 2024-03-14
update: use any port if all attempts fail

Axorax:


jelveh:
Thank you for your contribution! merging :)

PR #105, opened 2024-03-15
adding ability to customise desktop by arranging items

vineeth-vk11:
Added an option in the desktop context menu to disable/enable auto-arrange

On disabling auto arrange, the user can arrange items on the desktop by dragging items around.

When the auto arrange is enabled, the items return to sorted positions, but they can be brought back to previously customised positions by disabling the auto arrange again.

Screenshot 2024-03-15 at 8 33 35 PM

Pending Task - Handle auto arrange when multiple tabs of puter are open


jelveh:
Thank you @vineeth-vk11, this is coming along very nicely too! It works very well.

One thing to consider. I noticed that when in "manual arrangement" mode, each item's position is individually saved in the key-value store. This can cause some performance issues when loading the desktop later on since there will be one KV request made for each item on the desktop. My desktop has about 40 items on it and it takes a few seconds to load. An alternative approach might be to store all positions in one json and store that in the KV store. Let me know your thoughts :)

vineeth-vk11:
Makes sense, let me do that.

vineeth-vk11:
@jelveh I have moved the desktop items positions to a map now

jelveh:
Thank you @vineeth-vk11! after a much needed 10-hour sleep, I'm back and testing 😎

jelveh (commented):
seems like when I re-arrange items an empty array is passed to `puter.kv.set()` resulting in positions not being saved.

jelveh:
Merged this awesome feature 🎉
We probably need to add a lot more (live syncing across devices, usability testing, etc) but it's already in a very good place. Thank you @vineeth-vk11 for all your contributions.

Issue #12, opened 2024-03-04
Automatic Adjustment of Puter Windows to Stay Within Viewport on Browser Resize


jelveh:
Users often adjust the size of their browser window for various reasons, such as multitasking or changing display settings. Currently, in Puter, resizing the browser can lead to open windows moving out of the viewport, making them inaccessible or partially hidden. This issue can disrupt the user's workflow and requires manual adjustment to bring these windows back into view. Implementing an automatic adjustment feature that ensures all windows remain within the viewport during and after a browser resize would improve usability and UX.

Expected Behavior:

- **Automatic Re-positioning**: When the browser window is resized, any Puter window (or windows) that ends up partially or fully outside of the viewport should automatically be re-positioned to ensure it remains visible within the new viewport dimensions.
- **Intelligent Placement**: The adjustment should not merely bring the window back into view but should do so intelligently to minimize disruption. For instance, if only a small portion of the window is out of view, the system should adjust it minimally to make it fully visible again.
- **Respect User Layout**: While ensuring window visibility, the system should strive to maintain the user's original layout as much as possible, avoiding drastic changes that could disorient the user.
- **Notification/Option to Revert**: Optionally, after an automatic adjustment, a subtle notification could be provided to inform the user of the change, with an option to undo the adjustment if desired.

meetqy:
![image](https://github.com/HeyPuter/puter/assets/18411315/297dd811-39c2-4214-9e42-9c5e565b40c8)

I want to solve this problem by listening to the browser zoom, recalculating the position, is that possible?

PR #120, opened 2024-03-17
fix: Automatic Adjustment of Puter Windows to Stay Within Viewport on Browser Resize #12

meetqy:
`window-login` can also use this code.


jelveh:
Thank you for the PR! 
While testing for usability, I realized that it would be a good idea that open windows "remember" their initial position before resizing so that if the browser is expanded again, open windows can be put back in their initial position. Additionally, once the browser is shrunken, if the user moves a window, it might be a good idea to "forget" about initial positions since the user has signalled their intention to modify the state of their desktop. 

The following displays the usability issue with not remembering the initial positions.

https://github.com/HeyPuter/puter/assets/1715019/02e07aa7-8d6a-4b50-ae7a-63b8079a9625

This will take some work, so let me know if you need help.

meetqy:
okey, I can try it. I think we can also add that when the window is larger than the initial position, use the initial position to calculate the `radio`. Without waiting for the window to return to its original position.

meetqy:
https://github.com/HeyPuter/puter/assets/18411315/efa4046a-ab43-4fa1-be5a-36e9a21d47cb

add original_window_position in global.js

jelveh:
@meetqy this is looking great! I need to test a little more (it's an important feature) and I should be able to merge today. Thanks again! 💪

jelveh:
Thank you @meetqy for your contribution. I couldn't find any issues. Merging 💪

PR #122, opened 2024-03-18
Add Norwegian Bokmål

SondreNjaastad:


SondreNjaastad:
I'm close to having this, but with each language separated into individual files + language selector, so closing this as this way of handling the language will not match the new one

PR #123, opened 2024-03-18
Romanian Translation

EndreEndi:
Hi, here is Romanian translation, i'd say it's 95-98% Translated, 
The words that aren''t translated i use the English version of them so i don't know the translation for them :)


jelveh:
Thank you for your contribution! It's great to have Puter in Romanian :)

PR #127, opened 2024-03-18
i18n: Finnish translation

lsileoni:
## What

This change adds i18n for the Finnish language.

## Ambiguities

The change contains some unsolved ambiguities relating to the meaning of the words. The ambiguities are marked as `TODO` comments with suggestions on how to proceed. As the maintainers have a greater contextual understanding of the context behind words, most of the cases are just a matter of picking the most appropriate ones.


jelveh:
Awesome! thank you :)

PR #132, opened 2024-03-18
More translation

SondreNjaastad:


src/UI/UIDesktop.js
```diff
@@ -1197,18 +1197,19 @@ $(document).on('click', '.user-options-menu-btn', async function(e){
                     // see if there are any open windows, if yes notify user
                     if($('.window-app').length > 0){
                         const alert_resp = await UIAlert({
-                            message: `<p>You have open apps. Are you sure you want to log out?</p>`,
+                            message: `<p></p>`,
```
jelveh: I think this was removed unintentionally?

src/UI/UIItem.js
```diff
@@ -1245,14 +1245,14 @@ function UIItem(options){
                     html: i18n('delete_permanently'),
                     onClick: async function(){
                         const alert_resp = await UIAlert({
-                            message: `Are you sure you want to permanently delete this item?`,
+                            message: ` you want to permanently delete this item?`,
```
jelveh: This one looks wrong.

jelveh (changes requested):

src/UI/UIDesktop.js
SondreNjaastad: Yes, it should have bern replaced with a i18n replacement. Will fix

src/UI/UIItem.js
SondreNjaastad: It is wrong , for sure 😅

jelveh:
Another awesome PR, thank you!

Issue #101, opened 2024-03-14
Implement Warning Message in Browser Console to Prevent Security Exploits


jelveh:
## Description:
This issue proposes the implementation of a security warning message within Puter's browser console. Malicious parties may attempt to trick users into pasting harmful scripts into their browser console, which can lead to account compromise or other security breaches.

## Problem:
Some users might be unaware of the dangers of pasting untrusted scripts into their browser console. Scammers often disguise these scripts as hacks or features to entice users, leading to potential security vulnerabilities and unauthorized access to user accounts.

## Proposed Solution:
Introduce a clear and prominent warning message that appears in the console as soon as the developer tools are opened. This message will alert users that pasting code into the console can compromise their account security and should only be done when they understand the implications and trust the source of the script. The message could be styled with attention-grabbing colors (e.g., bold red text) and should include a link to a page with more information about the risks involved in such actions.

The following is an example of Facebook's notice to users when they open the browser console:

<img width="687" alt="Screenshot 2024-03-13 at 11 31 39 PM" src="https://github.com/HeyPuter/puter/assets/1715019/95180b1c-cbf1-4f45-8858-eff7689e1afc">

SaiUddisa:
On it. 
@jelveh mean while can you please answer my question posted on issue #116

SaiUddisa:
@jelveh checkout my new PR and let me know if  you need any more changes.

PR #137, opened 2024-03-19
Warning about self xss accounts added to console

SaiUddisa:
#101  
created a warning message in the console about the self-xss attacks.
also mentioned Google's blog post about these types of attacks.
wrote the console message in the index.js file to make it visible immediately before all other console messages.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,12 +7,12 @@
  * it under the terms of the GNU Affero General Public License as published
  * by the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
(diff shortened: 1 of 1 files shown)


jelveh:
@SaiUddisa Thank you very much for the contribution! It looks pretty nice.
One thing before I merge though, could you use tabs instead of double-spaces for indentation? I take blame for not having standards published yet, but if you could use tabs I can merge it. Thanks again!

SaiUddisa:
Absolutely I will do it by EOD

PR #138, opened 2024-03-19
Presist language user_preference

### LINKED ISSUE #133: Persist Language Preference Across Devices
Suggested by @SondreNjaastad 

## Description
Puter can now be used in multiple languages. Either Puter picks the language based on locale or the user can manually set their preferred language. Either way, at this point the preference is only stored locally using `localStorage`, so if the user accesses Puter on another device they need to set the language again.

## Proposed Solution
Using [`puter.kv`](https://docs.puter.com/KV/set/) we can set the preference in the user's cloud account in order to persist it across devices. There is already logic in the code that handles user preferences in KV, which could be used to store language preferences as well.

SondreNjaastad:
Resolves #133


jelveh:
Works like a charm. Merged.

PR #142, opened 2024-03-19
Add Armenian translation

1ilit:
Added Armenian translation :)


jelveh:
Thank you for your contribution @1ilit. It's great to have Puter available in Armenian 🔥

PR #145, opened 2024-03-20
Add Thai translation

athachai29:


jelveh:
Thank you for your contribution @athachai29 ! Thai is an awesome addition to the supported languages 💪

PR #149, opened 2024-03-20
Add a CI script to check translations are valid

AtkinsSJ:
Checks the following:
- Translation files are valid JS
- Each translation file is registered in translations.mjs
- Each translation's code matches its name
- Translation dictionaries only contain keys that exist in the English translation.

An example output for when things are right:
image

...and when various things are wrong:
image

On second thought I might have gone overboard with exclamation marks. :laughing:

### DIFF
--- a/.github/workflows/check-translations.yml
+++ b/.github/workflows/check-translations.yml
@@ -1,0 +1,28 @@
+# This workflow runs the tools/check-translations.js script to make sure that the translation data is valid.
+
+name: Check Translations
+
+env:
+  NODE_VERSION: 21.x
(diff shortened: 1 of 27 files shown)


AtkinsSJ:
Decided to try making the project use JS modules instead, and that was pretty easy, so I've done that in #151 instead of the file-renaming hacks I had done here. That'll need merging first.

jelveh:
> gone overboard with exclamation marks.

There is no such a thing !!!!!!

!!

AtkinsSJ:
Rebased and undrafted!!!

jelveh:
Very useful, thanks Sam!

PR #151, opened 2024-03-20
Migrate to JavaScript modules

AtkinsSJ:
This turned out to not be difficult actually!

Discovered when working on #149. We already use modules for most of the code, but not the parts that Node deals with. Making the whole project use modules, means that tests and other scripts that interact with module code will work nicely, without having to rename things.


jelveh:
Good idea, Sam. Merged

PR #156, opened 2024-03-21
handling context submenus on mobile devices

vineeth-vk11:
This approach opens the submenu on top of the main menu in mobile devices making sure sub menus don't get cropped out

### DIFF
--- a/src/UI/UIContextMenu.js
+++ b/src/UI/UIContextMenu.js
@@ -185,6 +185,17 @@ function UIContextMenu(options){
                 if($(`.context-menu[data-id="${menu_id}-${$(e).attr('data-action')}"]`).length === 0){
                     // close other submenus
                     $(`.context-menu[parent-element-id="${menu_id}"]`).remove();
+
+                    // Calculate the position for the submenu
+                    let submenu_x_pos, submenu_y_pos;
+                    if (isMobile.phone || isMobile.tablet) {
+                        submenu_y_pos = y_pos;
+                        submenu_x_pos = x_pos;
+                    } else {
+                        submenu_y_pos = item_rect_box.top - 5; 
+                        submenu_x_pos = x_pos + item_rect_box.width + 15;
+                    }
+
                     // open the new submenu
                     UIContextMenu({ 
                         items: options.items[parseInt($(e).attr('data-action'))].items,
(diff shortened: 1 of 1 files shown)


jelveh:
Thanks for this @vineeth-vk11. This works pretty well. The reason it took me long to respond is that I'm still thinking about whether we could change the UI of the ctx menu on mobile. For example maybe we could convert submenus into accordions on mobile. Thoughts?

o0101:
I probably shouldn't chime in I have no great design, but one cool idea that just emerged reading this thread was like

_The sub menu slides in from the right?_

Idk. Could be crazy, but would solve the space problem at least. Could be disorienting? A little back button or a micro crumb at the top could help orient?

I'm sure you folks have better ideas so just throwin' it in, in case. 😹

jelveh:
This won't work :/ 
Extra animation or far jumps from main menu to submenu bothers users. On the web, generally, this works because the website menu is not opened 100s of times, maybe 1-2 times. But on a desktop it has to be quick and in the same vicinity of the action.

o0101:
Hmm, @jelveh I see what you mean. Too many animations and surprising appearances and long distances would definitely be jarring. 

You probably already get what I'm saying but there's a chance you think I mean, *slides in from the right of the viewport*. What I really mean (sorry was not clear!) was:

- context menu (C1) opens in central box, the size of context menu, called 'A'
- on clicking menu item in C1, the submenu, C2, slides in from the right side of 'A' 
- this keeps happening when new items are clicked in subsequent submenus, they slide in from right of A, not right of screen. 
- 'A' remains fixed centered in screen the entire time.
- back button at top of 'A' returns to ancestor submenu.
- No submenus grow beyond the bounds of 'A', which could have vertical scrollbar to keep it clearer for longuns.

Call it a "wipe" transition between context sub menus? 

Is this already what you thought I meant?

Sorry I probably shouldn't impose my crazy ideas here! I'm sure you guys have already considered this and decided it doesn't work. But just in case! 

Anyway, I'm learning something' 😃 😹

vineeth-vk11:
@jelveh accordions can be an option but I am not confident because when the sub menus are long enough there is still a chance of overflow.

vineeth-vk11:
@jelveh how should we move forward on this

jelveh:
Hi @vineeth-vk11, I'll get back to this very soon, I need some time to finish Settings which I think is going to affect this as well.

jelveh:
Great fix! Thanks @vineeth-vk11

Issue #114, opened 2024-03-16
Icons under taskbar


KernelDeimos:
Chrome 122, Android 14

I can't see the text on this icon when I open on mobile.

![Screenshot_20240316-111408](https://github.com/HeyPuter/puter/assets/7225168/3cea5708-e3a7-45b2-8208-8111fdf53ccf)

SaiUddisa:
can you give me much more details about this issue.. I did not catch your point.

meetqy:
![image](https://github.com/HeyPuter/puter/assets/18411315/9045bc6d-6f69-4fb6-95c7-b8d76b64d0b9)

The Problem also exits on computers.

meetqy:
@jelveh Can I add a removable range to solve this problem?

jelveh:
@meetqy could you elaborate? I'm not sure what that is.

meetqy:
> @meetqy could you elaborate? I'm not sure what that is.

1. When the app icon moves, it detects whether it is outside the desired area.
2. If it is out of area, the app icon is fixed to the edge of the viewable area.

jelveh:
This wouldn't work because the item should be able to move over the Taskbar. This is because we have a feature allowing users to drop files on apps in taskbar and opening them. A better alternative would be to revert a drop if it's outside the viewport.  (e.g. under the taskbar and not on any taskbar item).

meetqy:
https://github.com/HeyPuter/puter/assets/18411315/c2d6ae28-725a-4928-8378-1b98a5219692

`New File.txt` also on taskbar?

jelveh:
It would have to go back to its original position unless it's dropped on an app.

jelveh:
In the long term we may want to allow for pinning files to the Taskbar, but this needs some user research. So for now, files should just return to their original position.

PR #158, opened 2024-03-22
fix: Icons under taskbar #114

meetqy:
https://github.com/HeyPuter/puter/assets/18411315/44f9ef8c-20c9-43cd-bb90-6f69233f9800


jelveh:
Thank you @meetqy for this! Definitely improves UX :)

vineeth-vk11:
@meetqy @jelveh This solution has a bug, 
Screenshot 2024-03-23 at 9 10 22 AM

With the current code, the item is rearranged and the position is stored and then goes to the original position because of the new condition. When you refresh it comes back to the position it was arranged to.

A PR which fixes this - #162

PR #162, opened 2024-03-23
minor fix in disallowing items arrange over task bar

vineeth-vk11:


jelveh:
Thank you for fixing the issue, @vineeth-vk11

PR #163, opened 2024-03-23
fixing html encoding issue in i18n

vineeth-vk11:
There was a bug which was causing html_encoding to always be set to true resulting in HTML elements not rendering properly

Screenshot 2024-03-23 at 11 45 52 AM
Screenshot 2024-03-23 at 11 49 10 AM
Screenshot 2024-03-23 at 1 55 50 PM


jelveh:
Thank you for this. It is indeed fixed!

PR #164, opened 2024-03-23
removing html inside translation strings

### LINKED ISSUE #141: Don't allow HTML inside translation strings
A few strings currently have links in. That's awkward both because it means if a URL changes then all the translations need to be updated; but mainly because allowing raw HTML there is an XSS injection risk.

Perhaps instead of:

```js
{
	// ...
	powered_by_puter_js: `Powered by <a href="https://docs.puter.com/" target="_blank">Puter.js</a>`,
	// ...
}
```

we could have something like:

```js
{
	// ...
	powered_by_puter_js: `Powered by {{link=docs}}Puter.js{{/link}}`,
	// ...
}
```

and then a fixed set of URLs defined somewhere, which translations can use. That `{{foo}}` syntax is arbitrary, we could also use a subset of something more common, like BBCode.

vineeth-vk11:
This PR resolves #141


vineeth-vk11:
@AtkinsSJ do check if this going in the right direction, I will move all URLs from all languages

vineeth-vk11:
@jelveh this is ready for review

PR #167, opened 2024-03-24
Fixes #148 - Added functionalities for i18n module

### LINKED ISSUE #148: Changing language does not update name of "Trash" until reloading the page
To reproduce:

1. Open Puter locally
2. See that there's a Trash icon on the desktop, and a Trash on the taskbar
3. Change language to Dansk (or anything really)
4. See that Trash is still "Trash" in both places
5. Reload the page
6. Trash is now "Papirkurv" (or the name in whatever language you chose) in both places

This might be a general problem with text not updating, I don't know, but this was the most obvious example.

ehsan-ashik:
Fixes #148. Added functionalities for i18n module to return html in order to grab elements and update relevant attributes based on the new language translation.

Added class `i18n` to the elements that need update at the time of changing language. In addition, a `data-i18n-key` attribute is added for elements that contains the i18n keys in order to make the correct translation.

`i18n()` function has been updated to include a new parameter `return_html = true` to return html element to be used in the application. 

For calls of `i18n()` to the title of  UI elements, a new attribute has been introduced to the UI modules (e.g., UIWindow, UIAlert) to include `i18n_key` field and set to the UI elements' `data-i18n-key` attribute.

### DIFF
--- a/src/i18n/i18nChangeLanguage.js
+++ b/src/i18n/i18nChangeLanguage.js
@@ -22,6 +22,40 @@ function ChangeLanguage(lang) {
     window.mutate_user_preferences({
         language : lang,
     });
+
+    // -------------------------------------------
+            // Get all 'i18n' elements and update their text, title, and data name attributes based on the new language
+    // -------------------------------------------
+
+    // find all element with the 'i18n' class to replace their values based on the new translation
+    var elements = Array.prototype.slice.call(document.getElementsByClassName('i18n'))
+
+    //iterate over all the elements and change attribute values
+    elements.forEach(element => {
+        if (element.getAttribute('data-i18n-key') !== undefined && element.getAttribute('data-i18n-key') !== null && element.getAttribute('data-i18n-key') !== ""){
+            let i18n_key = element.getAttribute('data-i18n-key');
+            let i18n_text = i18n(element.getAttribute('data-i18n-key'), false);
+
+            // if there is no translation for this key, do not proceed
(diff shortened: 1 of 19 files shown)


jelveh:
Thank you for your contribution @ehsan-ashik, this will require thorough review, I'm on it.

jelveh:
Thanks again for this PR. Preliminary tests are looking good! I have some comments but testing a little more first :)

ehsan-ashik:
Thank you for the update. Please take your time testing the changes and let me know your feedback when you are done testing.

src/i18n/i18nChangeLanguage.js
```diff
@@ -22,6 +22,40 @@ function ChangeLanguage(lang) {
...
+    // -------------------------------------------
+
+    // find all element with the 'i18n' class to replace their values based on the new translation
+    var elements = Array.prototype.slice.call(document.getElementsByClassName('i18n'))
+
+    //iterate over all the elements and change attribute values
+    elements.forEach(element => {
+        if (element.getAttribute('data-i18n-key') !== undefined && element.getAttribute('data-i18n-key') !== null && element.getAttribute('data-i18n-key') !== ""){
+            let i18n_key = element.getAttribute('data-i18n-key');
+            let i18n_text = i18n(element.getAttribute('data-i18n-key'), false);
```
KernelDeimos, line 37:
```suggestion
            let i18n_text = i18n(i18n_key, false);
```

PR #170, opened 2024-03-24
UIWindowRefer.js bugfix

sebe324:
Changes:
- Changed copy link button text to the i18n translation.
- Translated an untranslated word in the polish translation.

## Before:

img

## After:
img

Notes:
 - I had to add three styling changes to the button, because the translated text was overflowing.


jelveh:
Thank you for this!

PR #184, opened 2024-03-26
Fix typo in parameter name

tom-delalande:
Randomly stumbled across this instance which calls `isMobile.table` instead of `isMobile.tablet`. Using the code from `isMobile.min.js` there was no parameter for `table` so I'm assuming this is unintentional. Apologies if it is


jelveh:
Thank you @tom-delalande! Nice catch :)

PR #191, opened 2024-03-30
Make showing/hiding the clock configurable

meetqy:
https://github.com/HeyPuter/puter/assets/18411315/be24280f-91fd-40e3-83b9-2171d8fd24e4


jelveh:
Thank you @meetqy for the contribution! I've made a few small changes but now merging.

PR #195, opened 2024-03-31
feat: Add support for running Puter self-hosted using the automatically built Docker image

shuguet:
Hello team Puter!

First of all, thank you for releasing the backend of Puter into the world!
Been playing with it for a few hours, and I really like it :)

As promised, I've updated a bunch of things to support running Puter self-hosted without having to clone the entire repo locally.

In this PR you'll find:
1) Updated instructions on how to run Puter using Docker
2) Updated instructions on how to run Puter using Docker Compose
3) Updated Dockerfile to add missing `git` command (used by Puter to check if there's a new version available, as far as I can tell)
4) Updated Dockerfile to reflect the current app version
5) Updated ignore file to ignore the data & config folder used to persist files & config between runs, when running from the main repo
6) Placeholders for a healthcheck

Regarding that last point, to the folks at Puter: What's a good URL docker can check (from inside the container's context) to see if Puter is running healthy?
I've included, commented, an example of what that could/should look like, feel free to update accordingly.

Happy Puter'ing folks :)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,9 +17,9 @@ services:
     volumes:
       - ./config:/opt/puter/app/volatile/config
       - ./data:/opt/puter/app/volatile/runtime
-    healthcheck:
-      test: wget --no-verbose --tries=1 --spider http://localhost:4100 || exit 1
-      interval: 5m
-      timeout: 3s
-      retries: 3
-      start_period: 2m
+    # healthcheck:
+    #   test: wget --no-verbose --tries=1 --spider http://localhost:4100 || exit 1
+    #   interval: 5m
+    #   timeout: 3s
+    #   retries: 3
+    #   start_period: 2m
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,25 @@
-version: '3'
-
+---
+version: "3.8"
 services:
-  app:
-    build: ./
+  puter:
+    container_name: puter
+    image: ghcr.io/heyputer/puter:latest
+    pull_policy: always
+    # build: ./
+    restart: unless-stopped
     ports:
-      - 4000:4000    
+      - '4100:4100'
+    environment:
+      # TZ: Europe/Paris
+      # CONFIG_PATH: /etc/puter
+      PUID: 1000
+      PGID: 1000
+    volumes:
+      - ./config:/opt/puter/app/volatile/config
+      - ./data:/opt/puter/app/volatile/runtime
+    healthcheck:
+      test: wget --no-verbose --tries=1 --spider http://localhost:4100 || exit 1
+      interval: 5m
+      timeout: 3s
+      retries: 3
+      start_period: 2m
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,5 @@
 .dockerignore
 Dockerfile
 node_modules
+config
+data
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ dist/
 .env
 # this is for jetbrain IDEs
 .idea/
+config
+data
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,10 @@ FROM node:21-alpine
 # Set labels
 LABEL repo="https://github.com/HeyPuter/puter"
 LABEL license="AGPL-3.0,https://github.com/HeyPuter/puter/blob/master/LICENSE.txt"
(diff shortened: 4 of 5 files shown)


shuguet:
Tagging @KernelDeimos for review as well as @jelveh for information.

KernelDeimos:
> Regarding that last point, to the folks at Puter: What's a good URL docker can check (from inside the container's context) to see if Puter is running healthy?

I think `http://puter.localhost:4000/test` is a good choice. That endpoint just sends a response saying "It works!" but we could perform more detailed checks in the future.

docker-compose.yml
```diff
@@ -1,7 +1,25 @@
...
+      PGID: 1000
+    volumes:
+      - ./config:/opt/puter/app/volatile/config
+      - ./data:/opt/puter/app/volatile/runtime
+    # healthcheck:
+    #   test: wget --no-verbose --tries=1 --spider http://localhost:4100 || exit 1
+    #   interval: 5m
+    #   timeout: 3s
+    #   retries: 3
+    #   start_period: 2m
```
KernelDeimos:
```suggestion
    healthcheck:
      test: wget --no-verbose --tries=1 --spider http://puter.localhost:4100/test || exit 1
      interval: 5m
      timeout: 3s
      retries: 3
      start_period: 2m
```

shuguet: Happy to commit that, although have you tested it? Otherwise I'll get to it tomorrow morning my time (CEST).

Easiest way to check is to run that wget command from inside the container as it is running.
Open a new shell and run:
```
docker ps
docker exec -it <container id or name from "docker ps"> ash
```
We need to use `ash` as shell instead of `bash`, as the Node base image for this puter container image is built on Alpine, which uses `ash` as its main shell.

MohamedElashri:
I think for avoiding `CORS` browser policy violations, puter on anything other than localhost will need a way to pass the URL. 

You will get 
```
puter  | ┃  Puter is now live at: http://puter.localhost:4100  ┃
```

So if you try to run this on `foo.example.com` you will get errors like these.

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://api.puter.localhost:4100/socket.io/?auth_token=null&EIO=4&transport=polling&t=OwMpxtc. (Reason: CORS request did not succeed). Status code: (null).
```

KernelDeimos:
The domain can be configured in `volatile/config/config.json` (or `/etc/puter/config.json` if that directory exists)

MohamedElashri:
> The domain can be configured in `volatile/config/config.json` (or `/etc/puter/config.json` if that directory exists)

Thanks, I would suggest making it possible to pass those and env variables. Also you should add an option to specify which port the user would choose. because now it assumes that it will always be `foo.example.com:$PORT` where port is the one that the image is build with. But I might just expose another port or don't expose a port and just use reverse proxy and refer to the internal port.

MohamedElashri:
Also this will be extended to your api url hard coded. So we can change this hard coded to something suitable but then with each pull this need to be repeated. I think there should more flexibility into passing these argument during initial run. Maybe add all of these inside `config/config.json` would be a solution.

shuguet:
There's already an `http_port` parameter in the same `config.json` file.

As for environment variables, feel free to contribute support for those.
Since they do not exists right now, this PR adds support for all the features that *do* exists, including static config files and persistent storage.

docker-compose.yml
shuguet: I tested it, it works perfecty.

Notice the `(healthy)` in the `STATUS` column.

Thanks @KernelDeimos !

```$ docker ps
CONTAINER ID   IMAGE     COMMAND                  CREATED         STATUS                   PORTS                    NAMES
4736a94580a3   puter     "docker-entrypoint.s…"   3 minutes ago   Up 3 minutes (healthy)   0.0.0.0:4100->4100/tcp   unruffled_solomon```

shuguet:
@KernelDeimos : Updated (and re-enabled) the healthcheck both on the Docker image itself, as well as in the docker-compose file service spec directly.

I've also reduced the delayed check at start time to 30s, and the interval between checks to 30s as well.
It's better now, as it transition to a healthy status quicker (puter does not take 2min to start up), and will report any health issue quicker too.

In the future, as you guys add more checks behind the `/test` endpoint, docker (or any other monitoring tool) will also benefit from those.

Do you see anything missing, in that PR's scope, that you'd want me to add before approving/merging?

PR #198, opened 2024-03-31
Update pt.js

berto:
Correct many instances of PT-BR to PT-PT. The original had some typos too. Kept “Password” as “Password” instead of a mix of “Senha” and “Palavra Chave”.


jelveh:
Thank you for the contribution @berto! 🙌

PR #205, opened 2024-04-01
Fix issues with translation keys, and check these in CI

AtkinsSJ:
Expands our existing `check-translations.js` script to also find every use of `i18n()` and ensure that the string passed to it is an existing translation key. This has already helped me find a few errors, which I've fixed here, but should also help prevent similar mistakes from appearing in the future.

Example output:

image


jelveh:
Beautiful addition! 💪

Issue #208, opened 2024-04-02
Getting 404 error responses when running new release of self-hosted


Braingix:
**Issue Description:**
I am currently facing an issue while trying to run and access the new release of the run-self-hosted implementation for my local development. The problem arises when the server successfully returns the `index.html` file, but it encounters a 404 error when attempting to fetch other JavaScript (JS) and Cascading Style Sheets (CSS) files. I have verified that these JS and CSS files are indeed available in the repository.

**Steps to Reproduce:**
1. Clone or fetch the webserver implementation repository to a local directory.
2. Start the webserver according to the provided instructions or scripts.
3. Access the server using a web browser or HTTP client.
4. Observe that the `index.html` file is successfully returned by the server.
5. Attempt to fetch other resources such as JS and CSS files using relative paths from `index.html`.
6. Notice that the server returns a 404 error for these JS and CSS files, even though they exist in the repository.

**Expected Behavior:**
I expect the webserver implementation to serve all necessary resources, including JS and CSS files, without encountering 404 errors. Accessing `index.html` should also result in the correct loading of associated assets.

**Actual Behavior:**
The server successfully returns `index.html` but returns 404 errors when attempting to fetch other JS and CSS files, despite their presence in the repository.

**Additional Information:**
- I have double-checked the paths used to fetch JS and CSS files from `index.html` to ensure they match the actual paths in the repository.
- I have also reviewed the configuration files and settings of the webserver implementation to confirm that it is configured to serve static files correctly.
- Additionally, I have checked file permissions and the server directory structure to ensure that the JS and CSS files are accessible to the webserver.

**Environment:**
- Operating System: Windows 10
- Browser: Chrome

**Logs:**
Here is a screenshot of the log from one of the runs to provide more context
![Screenshot (84)](https://github.com/HeyPuter/puter/assets/78648115/85b24c84-90f0-43e9-82fb-11a2d68986d3)

KernelDeimos:
Does the gui still load successfully or does it load a blank/blue page?

developmentkw:
the same problem , I can see the register and login page, when I register , the page is blank and it seems index is not working or the wallpaper is missing from dist? idk 
you can check here http://app.zamanie.com/

Braingix:
> Does the gui still load successfully or does it load a blank/blue page?

Thank you for replying. The loads a blank page as seen here:

[...]

PR #215, opened 2024-04-02
Fixed issue #208 and added dev-runner script for selfhosting

Braingix:


Braingix:
> Thank you for your contribution, Odah. Could you look into the CLA, it's needed before we can merge.

I have signed it.

KernelDeimos:
The check isn't accepting it because the commit metadata differs from your GitHub account

KernelDeimos:
This commit has a large amount of style changes. Is this your editor doing that? We'd like to keep style changes separate from bug fixes and features, so these will need to be removed or PR'd separately

Braingix:
> This commit has a large amount of style changes. Is this your editor doing that? We'd like to keep style changes separate from bug fixes and features, so these will need to be removed or PR'd separately

I think I have format on save turned on in my editor. Didn't check for that.

Braingix:
> The check isn't accepting it because the commit metadata differs from your GitHub account

So I have to make another PR?

packages/backend/src/temp/puter_page_loader.js
```diff
@@ -16,68 +16,66 @@
...
+  gui_params = {
+    ...meta,
+    ...gui_params,
     app_origin,
     api_origin,
+    gui_origin: app_origin,
+  };
 
-    meta,
+  const asset_dir = env === "dev" ? "" : "";
```
KernelDeimos, line 49: We need the switch between dev using /src and production deploy using /dist

puter-gui.json
```diff
@@ -22,10 +22,10 @@
...
-        "/src/initgui.js",
-        "/src/helpers.js",
-        "/src/IPC.js",
-        "/src/globals.js",
-        "/src/i18n/i18n.js"
+        "/initgui.js",
+        "/helpers.js",
+        "/IPC.js",
+        "/globals.js",
+        "/i18n/i18n.js"
```
KernelDeimos, line 29: these changes appear to "un-do" the change on 54>49

KernelDeimos (commented):
The addition to `package.json` and missing imports for `APIError` references look good to me. Would you be willing to move these to a separate PR?

The changes for asset loading break deployment with the bundled version of gui (something we haven't documented yet, but we use it to deploy to puter.com as well), so this needs further evaluation and changes.

Do you still see 404 messages in the console on the latest commit? This is something I'm not able to produce, so my understanding was that #208 was fixed by my recent changes.

KernelDeimos:
> > The check isn't accepting it because the commit metadata differs from your GitHub account
> 
> So I have to make another PR?

For style changes I'm now thinking it's better to leave them out entirely. If we don't merge style changes it makes it more difficult for someone in the future to hide malicious changes in a massive "style PR" for example (like imagine something with 10x the changes of your PR, and then one line is changed somewhere; I think we don't want to set that precedent)

packages/backend/src/temp/puter_page_loader.js
Braingix: With the current server implementation it gives a 404 because the url is src/** instead of /**. When I removed the /src it worked. Maybe someone should look at the server code.

Braingix:
> > > The check isn't accepting it because the commit metadata differs from your GitHub account
> > 
> > 
> > So I have to make another PR?
> 
> For style changes I'm now thinking it's better to leave them out entirely. If we don't merge style changes it makes it more difficult for someone in the future to hide malicious changes in a massive "style PR" for example (like imagine something with 10x the changes of your PR, and then one line is changed somewhere; I think we don't want to set that precedent)

Got it.

packages/backend/src/temp/puter_page_loader.js
Braingix: I didn't remove the asset_dir declaration because I understood what it meant (it is easier to make future changes in the json file, then all will still work). In the dev-server implementation there are two middlwwares for serving static assets (./ and ./src or ./dist) that's why it doesn't break.

AtkinsSJ:
> So I have to make another PR?

No, you should be able to add whichever email address you used for that commit, to your GitHub account. I think that will work.

And depending on how comfortable you are with git, you could:
1. Do a soft reset of the commit. (`git reset --soft HEAD~1`) This undoes the commit but keeps the file changes.
2. Discard the code-formatting changes.
3. Test it still works.
4. Create a new commit of the changes.
5. Force-push to update it here. (`git push --force`)

And in future I'd recommend using a separate branch for each PR, it makes things a bit easier to manage.

KernelDeimos:
Closing as this is no longer active

PR #218, opened 2024-04-02
WIP: Barebones x86emu skeleton

lsileoni:
Work in progress PR for experimenting with x86 emulation, potentially to integrate with Puter using V86. Once this is done it should be enough for people to get started and contributing on bringing Linux to Puter in an easily deployable manner. The primary goal is to get a baseline going so everyone can get up and running and create something awesome!

**Warning**: This is in an intermediate stage. No guarantees of build working on your machine, nor guarantees about stability or security. There's a bunch of unnecessary files and files that are in the middle of being worked on and files that haven't been pushed.

TODO:

- [x] Easily reproducible build process
- [ ] Design document
- [ ] Finalize design of Instance/InstanceManager
- [ ] Figure out licensing concerns(BSD 2.0 for V86, fair attribution required, Xterm already attributed)
- [ ] Document build process
- [ ] Usage tutorial
- [ ] FIFO shell muxing(?)
- [x] ZSTD compression for delivery
- [ ] Security model
- [ ] Define constraints for implementation(how much memory can it use, network bandwidth etc)


KernelDeimos:
Hi,

First of all, this PR is fantastic. Not only in the effort already put toward this but also the checklist you provided for next-steps is comprehensive, and the disclaimer about security and stability is very responsible.

I'm attempting to run this right now and I'm running into an issue. I think you might have files already generated which I don't, but I'm not sure: (3rd line down)
```sh
mkdir -p "$ROOT_DIR/vm"
cd "$ROOT_DIR/vm" || exit
docker build . -t v86builder # <- here
docker run --name v86build v86builder
```

There's no Dockerfile in this directory, so I copied over the one from the `make_container` directory and it seems to have gone smoothly for a while

As a side-note, seeing this in the console output is very exciting
image

So now, after all the scary initramfs messages and compiler warnings were successfully passed by, guess what failed: a javascript import mechanism mismatch. Yep... I think this is evidence that it's more difficult to get that right than low-level dependencies. It's easier to do a task like "fix sound in linux" than to "use a javascript package" because of how this ecosystem evolved (yay).
image

Anyhow, jokes about that aside - any idea why I might be getting this error?

lsileoni:
> There's no Dockerfile in this directory, so I copied over the one from the `make_container` directory and it seems to have gone smoothly for a while

Yes you've got it correct, I've now reflected that in the starter script. Probably shouldn't make rushed changes to the structure right before a trip :sweat_smile:.

> So now, after all the scary initramfs messages and compiler warnings were successfully passed by, guess what failed: a javascript import mechanism mismatch. Anyhow, jokes about that aside - any idea why I might be getting this error?

The error was caused by the node script referencing to the `package.json` at the root of the repository and then failing somehow. I've moved that part of the build process to the `/tmp/` directory so it's isolated as a temporary measure, it should build the state image just fine now.

So, does it build for you now? The examples should run as follows: when you run `main.js` at `www`, it should hijack your terminal I/O with Node running debian on V86 and should be exitable with EOF/CTRL+D. And when you run a web server for the `/www/` directory and visit `index.html`, it should load two terminals with separate V86 emulators.

As a side note Zstd compression could reduce the main binary image from ~150MB -> ~30MB and compression level is configurable in the starter script now. Thank you for the help and kind words!

KernelDeimos:
That's pretty cool

image

KernelDeimos:
btw since this PR is in the `incubator` directory, which is meant for works-in-progress, you can take this out of draft and I'll merge it.

jelveh:
This is... quiet these days.

PR #234, opened 2024-04-05
fixing issues in copy while overwriting

### LINKED ISSUE #80: Api call for copy with overwrite set to true is failing with 500
### Steps to reproduce
Copy a file to a directory which already has a file with the same name

Screenshot 2024-03-12 at 9 46 33 AM

vineeth-vk11:
This PR closes #80 

options.user.id was causing the issue since there is no variable named options, has to change it to values

The other changes are related to handling the removal of overwritten item from DOM on the frontend.

Followed a similar approach to move operation, returned the overwritten item id in the api response and removed the item with that uid from DOM.

### DIFF
--- a/packages/backend/src/filesystem/hl_operations/hl_copy.js
+++ b/packages/backend/src/filesystem/hl_operations/hl_copy.js
@@ -163,6 +163,7 @@ class HLCopy extends HLFilesystemOperation {
             throw APIError('cannot_copy_item_into_itself');
         }
 
+        let overwritten;
         if ( await dest.exists() ) {
             // condition: no overwrite behaviour specified
             if ( ! values.overwrite && ! values.dedupe_name ) {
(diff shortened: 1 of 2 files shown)


packages/backend/src/filesystem/hl_operations/hl_copy.js
```diff
@@ -189,12 +190,13 @@ class HLCopy extends HLFilesystemOperation {
                 dest = await parent.getChild(target_name);
             }
             else if ( values.overwrite ) {
-                if ( ! await chkperm(dest.entry, options.user.id, 'rm') ) {
+                if ( ! await chkperm(dest.entry, values.user.id, 'rm') ) {
```
KernelDeimos, line 193: yep

packages/backend/src/filesystem/hl_operations/hl_copy.js
```diff
@@ -189,12 +190,13 @@ class HLCopy extends HLFilesystemOperation {
...
                     throw APIError.create('forbidden');
                 }
 
                 // TODO: This will be LLRemove
                 // TODO: what to do with parent_operation?
+                overwritten = await dest.getSafeEntry();
```
KernelDeimos, line 199: definitely

packages/backend/src/filesystem/hl_operations/hl_copy.js
```diff
@@ -213,7 +215,10 @@ class HLCopy extends HLFilesystemOperation {
 
         await this.copied.awaitStableEntry();
         const response = await this.copied.getSafeEntry({ thumbnail: true });
-        return response;
+        return {
+            copied : response,
+            overwritten
```
KernelDeimos, line 220: oh right 😅

KernelDeimos (approved):

Issue #220, opened 2024-04-02
parent-child relationships for opened apps, signals, and streams


KernelDeimos:
When an app is opened:
- `launch_app` is called in `helpers.js`
- A UIWindow is created
- via `postMessage`, the app has a channel to communicate with the GUI

Apps can open other apps with the `puter.js` SDK, so we should allow a mechanism through which an app can communicate with its "parent app" (the one which opened it) and vice-versa. We should provide an API similar to `postMessage` where arbitrary data can be sent, and we should also provide an API similar to the standard IO streams in POSIX so that apps can work with the terminal.

Apps should also be able to be "headless". I like the idea that apps communicate this intent to the OS on launch; an invisible "UIWindow" is still created and the app iframe is loaded, but the app iframe hooks into an event (`onBeforeAppLoad`, it could be called) and tells Puter "I want to stay invisible" so the UIWindow is never displayed.

Additionally, GUI could track the associations between apps to enable a "Task Manager" app that can help power-users and developers see what communication channels have been established between apps.

PR #238, opened 2024-04-05
Implement support for launching child apps, and IPC between a parent and its children

AtkinsSJ:
Details in commit messages.

See also https://github.com/HeyPuter/puter.js/pull/10 which implements the code that uses this!

There are a few questions to figure out:
- Should this functionality be part of `launchApp` instead of a special `launchChildApp`?
- What limitations do we want to place on launching child apps?
- What limitations do we want to place on apps communicating with each other?

Progress towards https://github.com/HeyPuter/puter/issues/220 :tada:

### DIFF
--- a/src/IPC.js
+++ b/src/IPC.js
@@ -92,6 +92,19 @@ window.addEventListener('message', async (event) => {
     //-------------------------------------------------
     if(event.data.msg === 'READY'){
         $(target_iframe).attr('data-appUsesSDK', 'true');
+
+        // If we were waiting to launch this as a child app, report to the parent that it succeeded.
+        const child_launch_callback = window.child_launch_callbacks[event.data.appInstanceID];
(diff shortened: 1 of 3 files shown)


KernelDeimos:
I think having them as two separate methods is fine for now because it's easy to merge that functionality later. I don't think we need any limitations for now; I see that as more of a tuning thing for heavy workloads, so more like "what gets priority" than "what's the limit"

KernelDeimos:
Actually let's have just `launchApp` doing it; since this is going in the SDK we need to support legacy function signatures so we'd be stuck with `launchChildApp` forever - better to just have one method

AtkinsSJ:
Merged `launchChildApp()` into `launchApp()`. Launching an app is always treated as launching a child, and returns an AppConnection to it.

PR #240, opened 2024-04-06
Puter Theme Color Setting

KernelDeimos:
|  |  |  |
| -- | -- | -- |
| image | image | image |

#### Update: Settings and Saving

image


src/services/ThemeService.js
```diff
@@ -0,0 +1,34 @@
...
+
+    get (key) { return this.state[key]; }
+
+    reload_ () {
+        // debugger;
+        const s = this.state;
+        this.ss.replace(`
+            .taskbar, .window-head, .window-sidebar {
+                background-color: hsla(${s.hue}, ${s.sat}%, ${s.lig}%, ${s.alpha});
+            }
```
AtkinsSJ: This could make use of CSS variables, so that you don't have to directly set `background-color` on the different classes that need it. For example:
```css
::root {
    --puter-window-background: hsla(${s.hue}, ${s.sat}%, ${s.lig}%, ${s.alpha});
}
```
...and then the regular `.taskbar` CSS etc would set `background-color: var(--puter-window-background, /* maybe some fallback value here */);` (There are some techniques for making sure there's a fallback without having to specify it everywhere too.)

And going on a tangent, if we did have a message-broadcasting system as some have suggested, we could then tell all open apps what the theme is when it changes, so they could make use of it. :thinking:

AtkinsSJ:
Looks like you need to add `'ui_colors'` to the English translation. (Yay, my thing was useful!)

PR #241, opened 2024-04-06
Implementing HostDiskUsageService removing diskusage dependency

vineeth-vk11:


vineeth-vk11:
@KernelDeimos I gave HostDiskUsageService a shot.

KernelDeimos:
PR looks good! I'm giving it a quick test right now

KernelDeimos:
This default of 90TB for unsupported platforms is _perfect_ - large enough to cover all practical cases, small enough that it won't confuse a UI, and also something about `90.95 TB` is just fun to look at.

image

It looks like the same `df` approach works just the same in Linux. I would still keep linux and darwin as separate methods; I'll add my implementation after this is merged.

One question - you multiply the output by `512`. Doesn't `df` output the number of 1K blocks?

Issue #247, opened 2024-04-08
Broadcast to apps when certain settings change


AtkinsSJ:
It would be useful for apps to be able to match Puter's colour scheme, language, and such. To make this possible, Puter could broadcast this information to apps whenever it changes, and also send it when an app first reports that it is `READY`.

For example, the theme broadcast could be something like:
```js
{
  name: 'theme',
  palette: {
    windowHue: '200',
    // ...
  },
  darkMode: true,
}
```
and then apps can do what they want with this. Eventually, fonts could be included in there once we have that as a setting.

- [x] Broadcast messages
- [x] Puter.js event listeners to make broadcasts easier to deal with
- [x] Send broadcast messages to newly launched apps
- [x] Theme broadcast
- [x] Language broadcast

jelveh:
Done! :)

PR #252, opened 2024-04-08
Add a message broadcasting system

AtkinsSJ:
A broadcast is a message sent to every running app that uses Puter.js. Broadcasts have a name and a data payload, and are sent as a 'broadcast' message.

Send a broadcast using `window.broadcast_manager.sendBroadcast()`. When doing so, you have the option to keep the broadcast message around, so that it can be sent to any newly-launched apps. Sending another broadcast with the same name will overwrite the previous one, so you don't have to worry about flooding a new app with duplicates.

Then, add broadcasts when either the theme, or locale change.

Contributes to #247.

[Peek 2024-04-08 13-49 puter theme broadcast noaudio.webm](https://github.com/HeyPuter/puter/assets/222642/952f2741-96f5-4c5d-9823-69bc4b3b76af)

### DIFF
--- a/src/IPC.js
+++ b/src/IPC.js
@@ -105,6 +105,9 @@ window.addEventListener('message', async (event) => {
             }, '*');
             delete window.child_launch_callbacks[event.data.appInstanceID];
         }
+
+        // Send any saved broadcasts to the new app
+        globalThis.services.get('broadcast').sendSavedBroadcastsTo(event.data.appInstanceID);
     }
     //-------------------------------------------------
     // windowFocused
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -722,7 +722,14 @@ window.mutate_user_preferences = function(user_preferences_delta) {
 window.update_user_preferences = function(user_preferences) {
     window.user_preferences = user_preferences;
     localStorage.setItem('user_preferences', JSON.stringify(user_preferences));
-    window.locale = user_preferences.language ?? 'en';
+    const language = user_preferences.language ?? 'en';
+    window.locale = language;
+
+    // Broadcast locale change to apps
+    const broadcastService = globalThis.services.get('broadcast');
+    broadcastService.sendBroadcast('localeChanged', {
+        language: language,
+    }, { sendToNewAppInstances: true });
 }
 
 window.sendWindowWillCloseMsg = function(iframe_element) {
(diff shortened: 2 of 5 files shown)


AtkinsSJ:
The dummy app seen in the video is:

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Dummy App</title>
    <script src="http://localhost:8081/dist/puter.dev.js"></script>
    <script>
        document.addEventListener('DOMContentLoaded', async () => {
            console.log('%cHello from dummy app!', 'font-size:200%;color:orange');

            puter.ui.on('themeChanged', theme => {
                const { primaryHue, primarySaturation, primaryLightness, primaryAlpha } = theme.palette;
                document.body.style.setProperty('background-color', `hsla(${primaryHue}, ${primarySaturation}, ${primaryLightness}, ${primaryAlpha})`);
            });
            puter.ui.on('localeChanged', locale => {
                document.getElementById('language').innerText = locale.language;
            });

            const parentApp = puter.ui.parentApp();
            if (parentApp) {
                document.body.innerHTML += `<p>I am the child</p>`;
                parentApp.on('message', message => {
                    console.log('%cDummy got message from parent', 'color:orange', message);
                    parentApp.postMessage('I love you too, parent!');
                    document.body.innerHTML += `<p><i>${new Date().toString()}</i> - Parent says: ${message}</p>`;
                });
            } else {
                document.body.innerHTML += `<p>I am the parent!</p>`;
                console.log('%cDummy has no parents :^(', 'color:orange');
                const childApp = await puter.ui.launchApp('dummy');
                childApp.on('message', message => {
                    console.log('%cDummy got message from child', 'color:orange', message);
                    document.body.innerHTML += `<p><i>${new Date().toString()}</i> - Child says: ${message}</p>`;
                });
                childApp.postMessage('I love you, child');
            }
        });
    </script>
</head>
<body>
This is a dummy app for testing!
<p>Language is <span id="language">unknown</span>.</p>
</body>
</html>
```

AtkinsSJ:
I wasn't entirely sure where to put this. It feels like a large amount of the code has ended up in `helpers.js` which isn't ideal, so I made this a separate class in its own file. Would it be worth moving more code out?

KernelDeimos:
I would register this to the `services` object like I did with ThemeService (that way everything stays as "flat" as possible architecturally).

`helpers.js` can be thought of as "functions without a file". Moving things out as needed seems to be good enough. Backend started with (and still has, partly) the same pattern and I've found moving helpers.js functions into the "registry of tiny process-local services" pattern had the most benefit

AtkinsSJ:
> I would register this to the services object like I did with ThemeService (that way everything stays as "flat" as possible architecturally).

Yup, that's much nicer! I wasn't sure what counted as a service but this makes sense.

Changes: Converted BroadcastManager to BroadcastService, and updated its users.

Issue #185, opened 2024-03-26
[Request] Unraid Support


al3x1337:
A Community App compatible version of Putty would be great for Unraid users. Currently can't figure out how to install this manually with Docker Compose Manger and can't find it on Dockerhub either.

AtkinsSJ:
Hi there! 

I'm not familiar with Unraid so just trying to make sure I understand what you're asking for: You want to control a remote server running Unraid, by using a Putty app (or similar) inside Puter?

Wdrussell1:
Unraid is a type of hypervisor that runs on baremetal. What they are requesting is that you build a docker container that can be imported into Unraid easily using their Community Apps plugin. 

https://unraid.net/community/apps

https://forums.unraid.net/topic/87144-ca-application-policies-notes/

KernelDeimos:
Is this url what you're looking for? ghcr.io/heyputer/puter:latest

al3x1337:
> Hi there!
> 
> I'm not familiar with Unraid so just trying to make sure I understand what you're asking for: You want to control a remote server running Unraid, by using a Putty app (or similar) inside Puter?

Self hosting Puter on Unraid. Having a Community App template would open up ease of installation for users on Unraid.

Using Docker Compose Manager (app), I was able to paste the following docker-compose template with the volumes modified to match. I ran New Permissions from the Tools menu on Unraid on the share that has my volume I modified from the Docker-Compose and was able to get to the Login Screen after bringing it up.

```
version: "3.8"
services:
  puter:
    container_name: puter
    image: ghcr.io/heyputer/puter:latest
    pull_policy: always
    # build: ./
    restart: unless-stopped
    ports:
      - '4100:4100'
    environment:
      # TZ: Europe/Paris
      # CONFIG_PATH: /etc/puter
      PUID: 1000
      PGID: 1000
    volumes:
      - ./puter/config:/etc/puter
      - ./puter/data:/var/puter
    healthcheck:
      test: wget --no-verbose --tries=1 --spider http://puter.localhost:4100/test || exit 1
      interval: 30s
      timeout: 3s
      retries: 3
      start_period: 30s
```

![image](https://github.com/HeyPuter/puter/assets/44815197/2c7149c5-908e-4cfc-963b-6da08ef79f3d)

I viewed the logs and have gotten the password for the default_user account but receive a red empty error box above the username field when attempting to login. The logs don't show any errors or provide helpful insight.

[...]

PR #253, opened 2024-04-08
Add an error message when login fails because of domain misconfiguration

### LINKED ISSUE #235: Display a proper error message when login fails because of misconfiguration
We keep getting bugs reported where the user gets an empty error box when trying to log in, like this:
image
(Taken from [#185](https://github.com/HeyPuter/puter/issues/185#issuecomment-2037916041))

Also from that issue is KernelDeimos's explanation:
>  This tends to happen when there's a configuration mismatch between domains or ports.
> 
> Make sure of the following:
> - `domain` in config.json is set to the domain you're using to access puter
> - DNS resolves for the domain, and the `api.` subdomain on that domain
> - `http_port` is set to the port Puter is listening on (`auto` will use `4100` unless that port is in use)
> - `pub_port` is set to the external port (ex: `443` if you're using a reverse proxy that serves over https)
> 
> We recently updated the [configuration instructions](https://github.com/HeyPuter/puter?tab=readme-ov-file#configuration) so be sure to check that out as well. For local DNS configuration you can look at [this reference](https://github.com/HeyPuter/puter/blob/main/doc/self-hosters/domains.md) that we also added recently.
> 
> _Originally posted by @KernelDeimos in https://github.com/HeyPuter/puter/issues/185#issuecomment-2037977592_

We really should display an actual error message in this case!

AtkinsSJ:
Previously, we just output whatever err.responseText was. However, that has some downsides:

- If there's no responseText (as for when we can't find the domain) then the user gets a blank message.
- If it's a 404 error, the responseText includes the full HTML for Puter's 404 page, which we don't want to dump in the error box! This is unlikely to happen in practice, but was easy enough to cater for.

So, add a nicer message in those cases.

The misconfiguration message is taken from here: https://github.com/HeyPuter/puter/issues/185#issuecomment-2037977592

Resolves #235.

image

Screenshot is taken from when I manually changed the domain to something nonsensical, but should solve the problem for users too.

### DIFF
--- a/src/UI/UIWindowLogin.js
+++ b/src/UI/UIWindowLogin.js
@@ -162,7 +162,30 @@ async function UIWindowLogin(options){
                     $(el_window).close();
                 },
                 error: function (err){
-                    $(el_window).find('.login-error-msg').html(err.responseText);
+                    const $errorMessage = $(el_window).find('.login-error-msg');
+                    if (err.status === 404) {
+                        // Don't include the whole 404 page
+                        $errorMessage.html(`Error 404: "${gui_origin}/login" not found`);
+                    } else if (err.responseText) {
+                        $errorMessage.html(err.responseText);
+                    } else {
+                        // No message was returned. *Probably* this means we couldn't reach the server.
+                        // If this is a self-hosted instance, it's probably a configuration issue.
+                        if (app_domain !== 'puter.com') {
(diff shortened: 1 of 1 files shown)


JeckerWen:
some wrong. Have you solved it?

AtkinsSJ:
> some wrong. Have you solved it?

Hi @JeckerWen, if you're having problems can you open an issue please?

PR #255, opened 2024-04-09
Notify apps when a child app closes, and let them close their children manually

AtkinsSJ:
Two parts to this, both related to closing. Further details in commit messages. See also https://github.com/HeyPuter/puter.js/pull/12 which is the puter.js part.

### Send an `appClosed` event to parent/child apps when an app is closed.

For now, the only close notifications are from child -> parent, and parent -> child, because that relationship is represented in the DOM. I think we need a proper representation in JS for what connections apps have between each other. It's been suggested for apps to be able to discover other running apps and communicate with them, and `AppConnection` can reasonably support this, but Puter itself would need to know that those connections exist to be able to send `appClosed` events, and whatever else.

### Listen to the 'closeApp' message, allowing apps to request to close other apps.

Currently, this is limited to apps being able to close their children, and godmode apps being able to close anything.

### DIFF
--- a/src/IPC.js
+++ b/src/IPC.js
@@ -1100,6 +1100,43 @@ window.addEventListener('message', async (event) => {
             contents,
         }, targetAppOrigin);
     }
+    //--------------------------------------------------------
+    // closeApp
+    //--------------------------------------------------------
+    else if (event.data.msg === 'closeApp') {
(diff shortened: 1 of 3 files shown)


KernelDeimos:
This looks good! Merging it

PR #261, opened 2024-04-11
WIP: Allow apps to be headless

AtkinsSJ:
Make app UIWindows initially hidden, and then show them when they send their READY message, if they didn't request to be headless.

Still remaining:
- Some way to ensure that non-SDK apps still show (probably with a timeout that automatically makes them visible)


KernelDeimos:
closing this as it's now handled by the `background` property on apps

PR #269, opened 2024-04-12
Update German translation

secondtruth:


jelveh:
Thank you for your contribution, @secondtruth!

Issue #254, opened 2024-04-09
White text for darker UI Theme colors and dark mode


Hiya-its-Aya:
Extention of issue #140 since @KernelDeimos merged a working color theme changer in PR #240. Some issues discussed:

- @AtkinsSJ added AppConnection; maybe `ThemeService` exposes an API to apps
- the foreground color (text) doesn't change accordingly
- dark mode

I'm currently working on the second issue adding white text to darker-colored themes, and then towards a full UI dark mode. 

The first point should be facilitated by @AtkinsSJ 's PR #252.

jelveh:
Really cool addition. Thank you, Aya!

jelveh:
@Hiya-its-Aya I know it's been a long time since you opened this issue (very sorry about this!), but starting with new issues we will respond in less than 24 hours. We'll never let issues hanging again.

With that out of the way, this seems to be complete aside from the dark mode part. I think it would be better if I closed this as complete and create a dark mode issue later on as that will probably be quite extensive.

Hiya-its-Aya:
Hey thank you for closing this. Sorry to go MIA. I got a job in May (yay!) and recently got my AWS S03 cert so I had little free time. I'm back to try taking another crack at dark mode or possibly another issue I can do. I'll have to take a look at the new architecture. Thanks and congrats on all the success!

jelveh:
Hi @Hiya-its-Aya. Congratulations on the new job and the certification, that's great news!
It's great to have you back and thanks for taking a stab at this. Let me know if you need any help.

PR #273, opened 2024-04-12
White text for themes

Hiya-its-Aya:
A part of a solution for #254.

Makes text white on below 60% lightness. 

Not perfect but it works! 

Thanks.


jelveh:
Really cool addition. Thank you, Aya!

PR #280, opened 2024-04-14
fixing on conflict issue in sqlite kv write

### LINKED ISSUE #206: Issues with local-hosted KV store
I'm noticing a couple of problems with `puter.kv` when running locally. At least, I *assume* they are problems; I want to check before messing with them.

Firstly, `puter.kv.get()`'s Promise seems to be returning the wrong type of data, when locally hosted. Specifically, it's returning `{ "value": "FOO" }` instead of just `"FOO"`. This means that after refreshing the page, the `language` and `clock_visible` settings get corrupted.

Second, each time a KV is saved, it appends to the database table instead of updating the existing record.

The steps to reproduce both of these are:

1. Launch a local Puter instance
2. Open settings
3. Go to the "Language" tab
4. Click on a few different languages
5. Refresh the page

The things to notice:
- Whichever language you had, the UI is now in English again. (The default when nothing is set.)
- Opening the Language settings again, nothing is selected.
- Opening the Local Storage browser inspector, the `user_preferences` value has something like `{"show_hidden_files":null,"language":{"value":"de"},"clock_visible":{"value":"show"}}`, with the `language` wrapped in `{ "value": "..." }`.
- If you open the sqlite database up, the same key has multiple rows, as below: (Second issue above.)
image
(Side note, https://sqlitebrowser.org/ is pretty good for browsing sqlite files.)

vineeth-vk11:
Closes #206

Currently, the UPSERT query on kv table in sqlite is not properly handled causing new rows to get added for the same key instead of the existing row getting updated. 

This was causing issues when updating languages and others preferences.

### DIFF
--- a/packages/backend/src/services/database/sqlite_setup/0001_create-tables.sql
+++ b/packages/backend/src/services/database/sqlite_setup/0001_create-tables.sql
@@ -120,7 +120,10 @@ CREATE TABLE `kv` (
   `value` text,
 
   -- 0016
-    `migrated` tinyint(1) DEFAULT '0'
+    `migrated` tinyint(1) DEFAULT '0',
+  
+  -- 0019
+    UNIQUE (user_id, app, kkey_hash)
 );
 
 CREATE TABLE `subdomains` (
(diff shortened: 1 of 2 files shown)


vineeth-vk11:
@KernelDeimos

PR #283, opened 2024-04-15
Re-add Puter platform support for phoenix

AtkinsSJ:
This was accidentally dropped by ignoring all directories named 'puter' instead of just the top-level one.

Note that I'm still having trouble getting phoenix or terminal to run, so this doesn't solve everything.


KernelDeimos:
Did this on my end (`main` branch) to make sure it has the exact state of the contents where it's working for me

PR #291, opened 2024-04-16
Re-enable Phoenix tests

AtkinsSJ:
This also runs the tests for contextlink, which I think we previously never ran. I don't see any other packages that define an `npm test` command. Having to manually list all the sub-package test directories in the top-level package.json is a bit unfortunate, but it works.

For Puter itself we support Node 16.x IIRC, but Phoenix requires at least 20.x currently.


KernelDeimos:
What makes the version requirement so high?

AtkinsSJ:
> What makes the version requirement so high?

Investigating in https://github.com/HeyPuter/puter/pull/297 - so far it's just one function used in the test harness.

PR #293, opened 2024-04-16
Implement stdio bridge for running apps from the terminal

AtkinsSJ:
After launching an app, if successful, we connect stdio streams to it, and wait for it to exit before we return to the prompt.

stdio is implemented as regular AppConnection messages:
- stdin:  `{ $: 'stdin',  data: Uint8Array }` from phoenix -> child
- stdout: `{ $: 'stdout', data: Uint8Array }` from child -> phoenix

SIGINT currently is not sent.

### DIFF
--- a/packages/phoenix/packages/pty/exports.js
+++ b/packages/phoenix/packages/pty/exports.js
@@ -16,6 +16,8 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
+import { raceCase } from '../../src/promise.js';
+
 const encoder = new TextEncoder();
 
 const CHAR_LF = '\n'.charCodeAt(0);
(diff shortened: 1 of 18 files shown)


AtkinsSJ:
**Changes:**
- All apps are now awaitable, and so we wait for them to exit before returning to the prompt. This is slightly hacky still, but should allow for a future solution without requiring an API change.
- SIGINT works! ...sometimes. It's limited to SDK apps. As far as I can tell this is a limitation of `SignalReader`, but I'm not certain.
- Terminal and Phoenix now use the same `stdin`/`stdout` messages that executed apps do.

AtkinsSJ:
**Changes:**
- Fixed the annoying bug where we consumed another chunk of data from stdin on the app's behalf, after it has exited.
- That broke the few cases where SIGINT was working before, so I've removed the commits enabling SIGINT. `SignalReader` needs a rethink before we can use it.

I think fixing the bug is more important than enabling the feature.

AtkinsSJ:
Copying the last commit message here because it explains the state we're in:

----

What's done here:
- Call releaseLock() on the stdin reader, which immediately cancels any
  active reads and prevents new ones.
- Adjust BetterReader to pass releaseLock() on, and handle the input
  ending, and expose how much input was read into a buffer if it's
  partially filled.
- Add an error message when the one user of read(opt_buffer) gets too
  small a buffer.

That last one is readline trying to read 1 byte, and we currently
infinitely loop on that after exiting a process. So this is definitely
not correct but it feels like it's the right direction.

AtkinsSJ:
...that readline problem sounds like it's caused by using the same Reader in multiple places. I think we might need to instead use multiple. Get one for readline, then release the lock on it when we pass a new stdin reader to a child process, then release that and obtain a new Reader for readline again. Something for me to look into tomorrow though, unless you want to try that @KernelDeimos?

AtkinsSJ:
Fixed the stdin problem, but found a new stdin problem: stdin is only readable for the first app launched in the terminal. Closing it and running another command, means that command doesn't get any stdin input. Investigating that now.

In the meantime, here's a demo app that interacts with stdin and stdout. It takes in text, then sends back the reversed text.
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Reverse</title>
    <script src="http://puter.localhost:4100/sdk/puter.dev.js"></script>
    <script>
        document.addEventListener('DOMContentLoaded', async () => {
            const parentApp = puter.ui.parentApp();
            if (parentApp) {
                const encoder = new TextEncoder();
                const decoder = new TextDecoder();
                let input = '';

                parentApp.postMessage({ $: 'stdout', data: encoder.encode('Type in text to reverse, and hit return\n') });
                parentApp.on('message', message => {
                    if (message.$ === 'stdin') {
                        const textIn = decoder.decode(message.data);
                        input += textIn;
                        if (input.includes('\n')) {
                            do {
                                const [ line, remainder ] = input.split('\n', 2);
                                input = remainder;

                                const reversed = line.split('').reverse().join('');
                                parentApp.postMessage({
                                    $: 'stdout',
                                    data: encoder.encode(`"${line}" reversed is "${reversed}"\n`),
                                });
                            } while (input.length > 0);
                            puter.exit();
                        }
                    }
                });
            }
        });
    </script>
</head>
<body>
You should not see this, this app should be headless
</body>
</html>
```

KernelDeimos:
Merging https://github.com/HeyPuter/puter/pull/304 which is the same PR but with two additional commits

Issue #232, opened 2024-04-05
Development console flicker


KernelDeimos:
### The issue

The prompt/widget area of the development console flickers when it re-draws. While on the surface this appears to be purely aesthetic, this flicker makes the console look a bit less stable, which might cause users to worry about missing log messages, which is not fun.

### Possible solutions

The following solutions are not mutually exclusive, and are in the order that it makes sense to apply them in:

- don't update if nothing changed (relatively easy, highest gain)
- compute more before drawing

#### Most of the time
![image](https://github.com/HeyPuter/puter/assets/7225168/2407e299-3015-4b15-b246-f6f362d77b67)
#### Briefly, during a re-draw
![image](https://github.com/HeyPuter/puter/assets/7225168/a6fdf9e8-baf7-4426-8892-b2ab71f2734d)

krauzX:
Can you assign me this?

KernelDeimos:
@krauzX sure thing! You've been assigned. Let me know if you run into any hurdles.

jelveh:
hey @krauzX, any updates on this? Let us know if you need help with anything.

guilhermegodoydev:
Hi! I'd like to work on this issue. Could you please assign it to me?

krauzX:
@jelveh pls assign @guilhermegodoydev to this I have a heavy workload and deadline to reach so I can't work on it

jelveh:
@krauzX, thanks for letting me know. @guilhermegodoydev, assigned! Let us know if you need any help.

guilhermegodoydev:
I’m opening the terminal from the home screen menu, but I don’t see the flickering effect mentioned.
Is this terminal the same as the “dev terminal” referenced in the issue?
If not, how do I access that terminal to reproduce the problem?

jelveh:
This would be the self-hosted puter and what it prints to the console. On a separate note, this seems to be resolved now! But I'mt testing a bit more anyway

PR #299, opened 2024-04-19
Optimize console redraw by tracking widget changes

Eric-Lighthall:
## Summary
Partial fix for issue #232, which relates to the prompt/widget area of the development console flickering.

## Details
This implements a tracking system that flags updates to the widget area, allowing redraws to be triggered upon widget changes instead of the previous redraw trigger only on a 2-second timer.

The previous 2-second redraw may only be apparent on slower devices.

## Visual Comparisons
Before (2 second flicker): 
https://github.com/HeyPuter/puter/assets/55029902/bb433ba4-da39-4b8b-be5a-821f9a95ffdd

After (no flicker) :
https://github.com/HeyPuter/puter/assets/55029902/0cf62b26-9eaf-402b-a46b-68e48c3d2e08

## Current Impact
This PR improves issue #232 by optimizing the 2-second redraw timer, reducing but not entirely eliminating flickering.

### DIFF
--- a/packages/backend/src/services/DevConsoleService.js
+++ b/packages/backend/src/services/DevConsoleService.js
@@ -24,6 +24,7 @@ class DevConsoleService extends BaseService {
         this.static_lines = [];
         this.widgets = [];
         this.identifiers = {};
+        this.has_updates = false;
     }
 
     turn_on_the_warning_lights () {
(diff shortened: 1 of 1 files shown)


KernelDeimos:
Thanks for submitting this fix! This little detail makes a big difference; the console looks a lot nicer.

PR #306, opened 2024-04-21
Do not show the "Change Password" option when user is temporary

### LINKED ISSUE #300: Guest account username/password modification while un-logged in
I discovered that it is possible to modify the guest's account password and username from the settings :
image
Because this a **guest account**, this should be disabled while in guest

zinreq:
fixes #300


jelveh:
Thank you for your contribution @zinreq! merged :)

PR #325, opened 2024-04-22
Update README.md

Jhonso7393:
Fixing a small annoying typo


KernelDeimos:
This is not a typo. The text "`protocol`" here functions as a proper noun, despite not being one, which is common in technical contexts. The change you suggested - "the protocol `protocol`" - conveys a different meaning that implies the `protocol` property is itself a protocol, which is not the case.

PR #341, opened 2024-04-24
Pass locale to the iframe url whenever accessing an app

### LINKED ISSUE #135: Pass Locale Preferences to Apps
Suggested by @SondreNjaastad 

## Description
Apps running within Puter should be able to retrieve the user's locale settings. 

## Proposed solution
For apps running within Puter, we could pass the locale preferences in the iframe URL of the app. An example would be the following query string appended to the URL `puter.local= en-US `.

Xiayucheng1212:
Fixes #135 

Add locale info to the iframe url.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -1756,6 +1756,9 @@ window.launch_app = async (options)=>{
             }
         }
 
+        // Add locale to URL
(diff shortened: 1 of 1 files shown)


jelveh:
Thank you @Xiayucheng1212!

PR #342, opened 2024-04-24
Update Docker CI

brhahlen:
Pull request to have Docker run on tags instead of just on the Release branch and on Pull Requests to Main (can add more branches there if needed).

This should automatically create the corresponding Docker image when releasing a new version (as a tag).
The push to GHCR only happens when a tag is created, so that should only leave "stable" versions to be pushed to GHCR.

Also removed a bunch of metadata that was generated, as I'm not sure if that's needed per se.

### DIFF
--- a/.github/workflows/docker-image.yaml
+++ b/.github/workflows/docker-image.yaml
@@ -5,8 +5,13 @@ name: Docker Image CI
 # branch called `main`.
 on:
   push:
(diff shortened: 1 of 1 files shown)


KernelDeimos:
Oh this is fantastic, we can get rid of the `release` branch and do this instead. Thanks!

PR #343, opened 2024-04-24
fix: Add missing imports for APIError

AtkinsSJ:
This specifically was causing a crash when trying to stat a non-existent path, but I found all the other places that use APIError and added imports to those too.


AtkinsSJ:
Rolling this into a larger ESLint-based PR

PR #346, opened 2024-04-25
fix(security): Prevent email enumeration

youngsiiimba:
This prevents a email enumeration vulnerability on the recover password page.


jelveh:
Hi there,
Thank you for your contribution. We still maintain that this is not a security issue per-se and enumeration should be prevented using rate-limiting. However, I'd like to give A/B testing this a try since our primary concern has been UX. So I'm going to merge this for now.

KernelDeimos:
We reverted this - it does **not** solve email enumeration and actually leaks emails for usernames

PR #348, opened 2024-04-25
Set up eslint, and try to resolve the issues it finds

AtkinsSJ:
eslint is a static analysis tool that tries to identify bugs and mistakes in the code. We have quite a variety of code in this repo so the config is a little complicated, but I've tried to make it clear. It's found a *lot* of issues (over 1700), and while many of which will not be actual problems, (it doesn't like us using window properties without `window.`,) but some definitely are.

This (unfortunately large but mostly straightforward) PR gets the number of issues down to two, because I couldn't figure out what to do with them.

Firstly:
```
/puter/packages/backend/src/routers/kvstore/clearItems.js
  41:32  error  'DB_MODE_WRITE' is not defined  no-undef
```
Other users of this do `const { DB_MODE_WRITE } = require("../MysqlAccessService");`, but that file doesn't seem to actually exist?

And then:
```
/puter/packages/backend/src/routers/whoami.js
  101:35  error  'db' is not defined  no-undef
```
There's no other database access nearby and I'm a bit confused by where it should come from. :sweat_smile:


AtkinsSJ:
This should be ready for review. Sorry for the immense size. Most of this is trivial changes but some of the genuine issues require some looking at to make sure I'm not making a silly mistake. I've tried to document things in the commit messages to hopefully be helpful.

As noted above, there are a couple of issues remaining because I am completely stumped. But after that, it might be worth enabling eslint on CI to stop some of these "accessing an undeclared variable" errors and such.

PR #349, opened 2024-04-26
fix(security): Prevent user enumeration.

youngsiiimba:
Quick fix to prevent user enumeration on the login page.


KernelDeimos:
Hello, we appreciate the concern addressed by this PR. However, we also understand that other services - for example Amazon and Google - do in fact inform you when an email does not exist at the login page. For UX reasons, we would like to keep this as-is.

We would, however, accept a PR which makes this an optional change for deployments of Puter.

Issue #356, opened 2024-04-29
Window blank scroll


Bashamega:
Hello:)
I noticed that the scroll bar that it is in all the windows is visible even if there is not items.
![image](https://github.com/HeyPuter/puter/assets/110662505/a23db7f1-54c8-4370-9b9a-e7a5330ef1ea)
Which is occurring because if the `overflow-y: scroll;`.
I would like to work on this issue.

jelveh:
It would be nice to hide scrollbars when they are not used. Is this happening on Windows?

Bashamega:
> It would be nice to hide scrollbars when they are not used. Is this happening on Windows?

Yes. My os is windows 11, and i am using MS Edge.

jelveh:
Just assigned this to you. Let me know if you need help.

PR #360, opened 2024-04-30
Fixes the scroll bar

Bashamega:
#356


jelveh:
Thank you for the fix!

Issue #362, opened 2024-04-30
Error installing Puter


marco00petrucci:
### Discussed in https://github.com/HeyPuter/puter/discussions/358

<div type='discussions-op-text'>

<sup>Originally posted by **marco00petrucci** April 29, 2024</sup>
Hi! I don't understand how to install puter. I've followed the instructions but it gives me this error:
![image](https://github.com/HeyPuter/puter/assets/100141045/4dc9b112-f930-42ed-8a04-e66c340bca00)

Can somebody help me please?</div>

AtkinsSJ:
Hi, can you try applying the changes I made in #364 and let me know if that solves this?

PR #364, opened 2024-05-01
fix: Make postinstall script work on Windows

AtkinsSJ:
Split each `cd` call into one per directory so that we don't have any directory separators, because separators are different on Windows than Mac/Linux.

This is intended to fix the issue in #362 but I'm waiting for confirmation on that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "start": "node run-selfhosted.js",
     "build": "node ./build.js",
     "check-translations": "node tools/check-translations.js",
-    "postinstall": "cd packages/phoenix && cd packages/contextlink && npm install && cd - && cd packages/strataparse && npm install && cd - && cd packages/pty && npm install"
+    "postinstall": "cd packages && cd phoenix && cd packages && cd contextlink && npm install && cd .. && cd strataparse && npm install && cd .. && cd pty && npm install"
   },
   "workspaces": [
     "packages/*"


marco00petrucci:
I replaced the line 30 as you said from:
`"postinstall": "cd packages/phoenix && cd packages/contextlink && npm install && cd - && cd packages/strataparse && npm install && cd - && cd packages/pty && npm install"`
to:
`"postinstall": "cd packages && cd phoenix && cd packages && cd contextlink && npm install && cd - && cd strataparse && npm install && cd - && cd pty && npm install"`

but it seems it doesn't work.
I've typed `npm start` and it says:
image

Thank you a lot for the answer! ♥️

AtkinsSJ:
Thanks for trying it out, did you run `npm install` after changing the line? I should have mentioned that.

AtkinsSJ:
Ah... I think `cd -` isn't a thing on Windows. I've changed it again, hopefully it'll work now. Can you try applying my changes here and then running `npm install` and `npm start` again?

marco00petrucci:
Now `npm install` worked.
But if I make a `npm start` it says:

> "export" is not recognized as an internal or external command, operable program or batch file

image

AtkinsSJ:
Thanks again. Seems like making it work on Windows is going to be more work than I anticipated. :sweat_smile: Until then, you could try using the Docker build. Setting up Docker on Windows might not be straightforward depending on your system though. ([Link](https://docs.docker.com/desktop/install/windows-install/))

marco00petrucci:
Thank you for now, I really appreciate your help.
I already tried using Docker but it gives me an error:
image
Can you help me with it please? Thank again.

AtkinsSJ:
Oh whoops, I forgot our instructions for that assume Linux. I don't actually use Docker so I can't help you unfortunately.

AtkinsSJ:
No longer relevant, we don't have a postinstall script any more!

PR #380, opened 2024-05-08
Run ESLint on CI

AtkinsSJ:
Run ESLint on PRs and pushes to the main branch, as we already do for the check-translations script (which is basically a linter too and should maybe become an eslint plugin). This should hopefully stop certain kinds of bugs from creeping in.

This PR first fixes up some lint issues with recently-merged code, and then suppresses the errors on a couple of lines I couldn't correct before, with FIXME comments that we need to come back to them later. Those are genuine bugs from what I can tell, I just don't know how to fix them as they both (AFAIK) involve the distributed backend code, which isn't in this repo.

We're mostly using the default ESLint configuration, with a few adjustments I made previously. We will probably want to adjust the settings over time, but I feel strongly that this is better than nothing. (The main annoyance is requiring `window.` everywhere, but that's just more motivation to pull things off the global object and into es modules or services, which seems like a good idea.)


packages/backend/src/routers/kvstore/clearItems.js
```diff
@@ -38,6 +38,8 @@ module.exports = eggspress('/clearItems', {
     }
 
     const svc_mysql = req.services.get('mysql');
+    // FIXME: MySQL service doesn't exist in self-host. Should this use the `database` service?
```
KernelDeimos:
```suggestion
    // TODO: check if used anywhere, maybe remove
```

packages/backend/src/routers/whoami.js
```diff
@@ -101,6 +101,8 @@ WHOAMI_POST.post('/whoami', auth, fs, express.json(), async (req, response, next
     if(req.query.return_desktop_items === 1 || req.query.return_desktop_items === '1' || req.query.return_desktop_items === 'true'){
         // by cached desktop id
         if(req.user.desktop_id){
+            // FIXME: Where does `db` come from? Is this on the window object, or a bug?
```
KernelDeimos:
```suggestion
            // TODO: check if used anywhere, maybe remove
```

KernelDeimos (changes requested):

PR #385, opened 2024-05-09
Add password requirement to: disable 2FA, change email

KernelDeimos:
To improve security we would like to define and configure middleware used by each route in a small number of groups or "access levels". Right now we only have system-wide middlewares and middlewares that are defined per-route.

Access levels might be:
- Puter Public - login, signup, etc
- Puter User - session token is required
- Puter User Dangerous - session token + password are required
- Puter API (the `api.puter.com` endpoints which need an authorization token)

The scope of this PR is prerequisite refactoring and addition of the "Puter User Dangerous" access level.

### DIFF
--- a/packages/backend/doc/contributors/boot-sequence.md
+++ b/packages/backend/doc/contributors/boot-sequence.md
@@ -1,0 +1,87 @@
+# Puter Backend Boot Sequence
+
+This document describes the boot sequence of Puter's backend.
+
+**Constriction**
+  - Data structures are created
+
+**Initialization**
+  - Registries are populated
+  - Services prepare for next phase
+
+**Consolidation**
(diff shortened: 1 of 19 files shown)


packages/backend/doc/contributors/boot-sequence.md
```diff
@@ -0,0 +1,87 @@
...
+  - Service event bus receives first event (`boot.consolidation`)
+  - Services perform coordinated setup behaviors
+  - Services prepare for next phase
+
+**Activation**
+  - Blocking listeners of `boot.consolidation` have resolved
+  - HTTP servers start listening
+
+**Ready**
+  - Services are infomred that Puter is providing service
```
AtkinsSJ: Typo: informed

PR #389, opened 2024-05-10
Update README.md docker-compose

fredbcode:
Fix docker-compose syntax


fredbcode:
I'm not in my office for now, but I'm using debian 12 and Ubuntu 22.04 (apt packages)

You are right the "new" syntax is docker compose, maybe this is not by default

fredbcode:
Docker version 20.10.24+dfsg1, build 297e128

docker-ce is provided by docker.com, docker.io is provided by Debian 

The plugin "compose" is not provided with docker.io

PR #390, opened 2024-05-10
refactor: Convert Task Manager to Components

AtkinsSJ:
## Add a Table component, based on the TaskManager table

For now, this is quite basic. The main feature is the sticky header.

## refactor: Convert Task Manager to use Components

This currently behaves the same as it did before: It still recreates the
table contents every half a second. It should also look identical,
though it's possible I missed some small differences.

The component structure is:

- TaskManagerTable
  - Table
    - TaskManagerRow
    - TaskManagerRow
    - TaskManagerRow
    - ...

TaskManagerRow is implemented so that we can later move to modifying
them in place as the process tree changes, instead of having to replace
them all.

Otherwise, most of the code is just moved around, and not changed much.

## fix: Only run Component initialization functions once

If the Component gets removed from the DOM and then re-added, it already
has contents, and we don't need to create them again. It also has
already had on_ready called, so that doesn't need to happen again
either.

This fix stops Components duplicating their content elements and
listener callbacks whenever they're moved around the document.

## Ignore setting a ValueHolder's value to its current value

This saves unnecessary work when nothing has changed.

## Re-use existing Task Manager rows if possible instead of recreating them

Most of the time, we'll already have a TaskManagerRow for the given
process, so we can just update its properties.

Iterating the results from #iter_tasks means we also insert the rows in
the correct order, regardless of their previous order.


KernelDeimos:
> Ignore setting a ValueHolder's value to its current value

re-testing the 2FA flow is necessary for this optimization. It's very possible this change will break it, as I believe I'm using `this.set('value', this.get('value'))` to perform a forced-update. Maybe `set` can have a `{ force: true }` option.

AtkinsSJ:
> > Ignore setting a ValueHolder's value to its current value
> 
> re-testing the 2FA flow is necessary for this optimization. It's very possible this change will break it, as I believe I'm using `this.set('value', this.get('value'))` to perform a forced-update. Maybe `set` can have a `{ force: true }` option.

Ah, that's unfortunate. I'll drop that commit for now then, it's not vital.

AtkinsSJ:
**Changes:**
- Rebased to fix conflict
- Dropped the "Ignore setting a ValueHolder's value to its current value" commit

PR #391, opened 2024-05-10
refactor: Put truncate_filename() helper in its own file

AtkinsSJ:
Every user previously set the max_length as window.TRUNCATE_LENGTH, so I've moved that constant into the truncate_filename file and set it as the default if max_length is not specified.


AtkinsSJ:
**Changes:**
- Resolved conflict

PR #396, opened 2024-05-14
fix(Terminal): Accept input from Chrome on Android 

### LINKED ISSUE #217: no keyboard typing in terminal from Android
just open up terminal on brave on Pixel 7

keyboard should work, but it is not acknowledged

cool system.

AtkinsSJ:
Fixes #217.

Xterm.js produces two kinds of events: `onKey` and `onData`. On a desktop, these are effectively the same, but on mobile, IME inputs produce data but not key presses. By listening to `onData` instead of `onKey`, we get that input.

With some experimentation, I also found that we don't need the code to handle enter, home, end, or Ctrl-Shift-V. All of these function as expected without that code, so we can remove it and simplify this further. :^)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4071,6 +4071,19 @@
         }
       }
     },
+    "node_modules/@xterm/addon-fit": {
+      "version": "0.10.0",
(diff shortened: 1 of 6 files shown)


strusty:
I received a notification that this issue was closed out. however the problem still persists.

PR #400, opened 2024-05-15
fix: Correct APIError imports

AtkinsSJ:
APIError is the only thing exported from its file, so we must not wrap it in {}.

When I "corrected" these in the ESLint PR, I never had errors come up so I didn't realise that it was still broken. :sweat_smile: Works now though.


KernelDeimos:
Additionally it is also possible for other open PRs to still be using the outdated imports when changes like this are made. For these reasons I avoid changes like this; there's no benefit (editor auto-imports in either case; if it didn't you'd probably have referenced some other file for APIError's usage and then already know about this) but there are a number of ways it can introduce bugs.

PR #403, opened 2024-05-16
Update fr.js

Zac0511:
I translated the new 2FA things (since they were in the `en.js` file), so they are in french.


Zac0511:
Huh those `'` caracters seems to be broking the code...

Zac0511:
Hmmm i think i broken the code... i'l try to repair it

Zac0511:
Okay, the code is **completly broken** for the moment. I'l try to repair it later. Its because i was there was a lot of new words missing, so i tried updating from `en.js`, but i broke the code by trying to do this.

Zac0511:
I think i'l work on this later, my head is starting to hurt

Zac0511:
I'm patching more and more bugs, it will be ready soon...

lucasgames8957 (approved):
Stinky man has arrived!

Zac0511:
All bugs patched ! Its ready to merge, and i have intergrated **every string** from [en.js](https://github.com/HeyPuter/puter/blob/main/src/i18n/translations/en.js), so it should work !

PR #408, opened 2024-05-17
Add a git client

AtkinsSJ:
Builds on #413.

A git app, that can be run inside Puter's terminal.

This is limited to the following commands, which are all implemented to some extent:
- `git help`
- `git version`
- `git init`
- `git status`
- `git add`
- `git commit`
- `git log`
- `git config`
- `git show`

It still lacks:
- Branches
- Tags
- Anything that interacts with a remote server (eg cloning, pulling, or pushing)
- Rebasing
- Stashes
- A lot of functionality with those implemented commands. To say that git is very configurable is an understatement. :sweat_smile:


lucasgames8957:
Cant wait... I want it, I want it, I want it, I want it, I want it, I want it, I want it, I want it, I want it, I want it, I want it, I want it, I want it, I want it,, I want it, I JUST WANT IT

triptych:
Yes, this is by far my number 1 request for puter.

AtkinsSJ:
I think this is in a decent enough state to be worth reviewing.

A couple of things that have come up:
- It'd be nice to have a better API for stdin/stdout/stderr, that just take strings, and all the encoding is kept internal
- Passing an exit code to `puter.exit()` would be convenient
- Being able to launch a background app from the start menu is a bit odd. For git, I don't think you can usefully do anything, but for some console apps like a python interpreter, you'd want a way of forcing it to have a parent terminal and shell.

PR #418, opened 2024-05-22
Git client part 2: HTTP

AtkinsSJ:
Initially this was just about HTTP but it's ended up being many things. Enjoy!

New commands:
- `git clone`
- `git remote`
- `git branch`
- `git fetch`
- `git pull`
- `git push`
- `git checkout`
- `git diff`
- `git restore`
- `git cherry-pick`

Other noteworthy features:
- Coloured output
- Diff display
- Ref names show up when looking at commits
- Bug fixes


AtkinsSJ:
`git checkout <from> -- <pathspec>...` is... interesting. I have a stash where I added the functionality, which isogit actually does for you as part of their `git.checkout()` function... except it just does a bunch of processing, after which the files haven't changed. So I might be missing something.

AtkinsSJ:
Marking this as ready, so people have a chance to ask for changes! :sweat_smile:

There's more I'd like to do, particularly making commit references like `HEAD~2`, or `master^1`, work, but that doesn't have to delay this PR. I'll just add that if this isn't merged, or submit another PR with it. :^)

`git` is far more pleasant to use with the changes in #498, too.

jelveh:
Can't wait to merge this 💙

AtkinsSJ:
Parsing `foo^` and `HEAD~3` turned out to be relatively straightforward, so that's added here. :^)

PR #420, opened 2024-05-23
feat: Display app icon and description in embeds

### LINKED ISSUE #398: Show app descriptions in oembed
When linking to a Puter app in a place that supports oembed, currently we show a description of Puter itself:

That's valuable when linking to puter.com, but when linking to a specific app, it would be nice to show a description and picture of that specific app instead. Perhaps it would still mention running that app on Puter, but I don't know what the text limit is for these.

AtkinsSJ:
**I have NOT been able to test this** because the only tester tool I could find requires HTTPS and no `:port` in the URL, which I haven't figured out how to set up. :sweat_smile: 

Fixes #398 - turns out we don't actually do oembed, it's just OpenGraph and Twitter metadata that other sites like Discord also support. Maybe we should support oembed too?

I'm not sure what the distinction between `description` and `short_description` is in this code.

We might want to also insert a mention of what Puter is into the description somewhere.

As a side note we have ~~three~~*two* different places that generate the HTML for the Puter GUI:
- `generateDevHtml()` in `utils.js`, which is used by the old "run only the GUI locally" server, `npm run start=gui`. That seems to be broken.
- `html_head()` in `backend/src/html_head.js`, which is unused
- ~~`generate_puter_page_html()` in `backend/src/temp/puter_page_loader.js` which is what actually runs, but is in a `temp` directory~~

So that could use some figuring out. I got a bit confused. :sweat_smile:

### DIFF
--- a/packages/backend/src/routers/_default.js
+++ b/packages/backend/src/routers/_default.js
@@ -265,7 +265,7 @@ router.all('*', async function(req, res, next) {
         // ------------------------
         else{
             let canonical_url = config.origin + path;
-            let app_name, app_title, description;
+            let app_name, app_title, app_description, app_icon;
             let launch_options = {
                 on_initialized: []
             };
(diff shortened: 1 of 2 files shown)


AtkinsSJ:
Fixed the conflicts.

PR #424, opened 2024-05-24
Actually parse sed scripts

AtkinsSJ:
Finally, `sed` is actually useful! :tada:

In the end, I did manage to parse things like the `s` and `y` commands using the grammar DSL instead of a custom Parser class for them. For `s` it's about half as much code, too. Pretty nice! Thought that's not including the code for the `DelimiterParser`.


AtkinsSJ:
I think this is ready now. `sed` still has some missing features, but the majority of it is there.

PR #442, opened 2024-05-28
Eric/extensibility

KernelDeimos:


KernelDeimos:
linter errors here are because of something I would like to allow, so ideally we should update linter configuration to allow this.

Issue #453, opened 2024-06-04
Terminal loses focus when running a background app


AtkinsSJ:
This is a general problem, but most obvious in the terminal:

When an app is launched, it gets the keyboard focus. That makes sense for most apps, but for background apps, they're invisible and can't take focus. All it does is remove the focus from the app the user is actually using instead, which is usually the terminal.

I expect this will be quite a simple fix, I'm just in the middle of something else right now.

jelveh:
I'll wait for somebody to pick this up as a first issue, if nobody does I'll fix it

Juhi-Lakdawala:
Hey can I work on this issue? Can you assign it to me

KernelDeimos:
> Hey can I work on this issue? Can you assign it to me

Sure thing! Let me know if you need help finding anything - I was pretty close to that code recently and have an idea on my head about how it would be fixed.

KernelDeimos:
@Juhi-Lakdawala did you run into a hurdle when trying to implement this? Let me know if you need help.

jhan10jr:
Hi, Can I work on this issue please?

KernelDeimos:
@jhan10jr Hi, I just noticed now (GitHub's UI makes it a little difficult to catch everything) - I'll assign you right now, let me know if you're still interested.

hewansirak:
Hello @KernelDeimos , can I work on this issue, could you please assign me?

gautamkapoor1709:
hey @jelveh , can i work on this issue??

KernelDeimos:
Hi, anyone can take on this issue if they like. If there's duplicate effort, that's not all that bad because the more people who understand this part of the source the better. I'm going to try solving it this using an agentic AI tool because I'm curious if it cal pull it off.

KernelDeimos:
@gautamkapoor1709 Great timing, I tried to get AI to do it yesterday and Claude is rate limiting everyone hard right now so it was a futile attempt. I just pushed the branch for that as `fix/terminal-focus-loss`, so you might want to check that for a general idea of where to look based on how Claude approached it. I've just assigned you.

Hetsavani:
is this issue still open?
if so i would like to work on it
And i will require some more explanation or steps to reproduce issue

bhumijain19:
Hey, I’d like to work on this issue. Could you please assign it to me if its still open

braintrue:
[...]

PR #463, opened 2024-06-07
refactor: conditionally focus terminal if phoenix is active

kumaryogesh17:
Issue #453
Added conditional statement that checks whether the variable phoenix exists and whether the phoenix application is active. If both conditions are met, it sets focus to the terminal window.

### DIFF
--- a/packages/terminal/src/main.js
+++ b/packages/terminal/src/main.js
@@ -117,5 +117,7 @@ window.main_term = async () => {
     const ioController = new XTermIO({ term, pty });
     ioController.bind();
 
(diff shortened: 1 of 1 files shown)


jelveh:
Thank you for the PR. Could look into the CLA before I can review/merge this.

KernelDeimos:
What is this for? If this is suppressing an error we should determine the root cause instead.

jelveh:
I'm closing this due to no response

Skandiakiwi13 (approved):
Help

PR #485, opened 2024-06-20
feat: add protected apps

KernelDeimos:


KernelDeimos:
eslint is wrong here, `mime` on main in the specified file **is** defined - it's imported at the top of the file!

AtkinsSJ:
> eslint is wrong here, `mime` on main in the specified file **is** defined - it's imported at the top of the file!

Actually no, it got removed in 7006dcc31c26504d01e79b0db3f73c7989c8acfb apparently by accident. Fixed in #487.

At least that answers my question of whether CI checks the PR merged against main!

PR #491, opened 2024-06-24
/sharelink/request

KernelDeimos:


KernelDeimos:
The overhead of managing eslint false positives is becoming _a bit much_ so I'm electing to not do that this time; eslint fails but I've carefully examined the issues it has reported and they're all irrelevant to this PR.

Issue #417, opened 2024-05-22
Error when calling `puter.fs.readdir()` on a file path


AtkinsSJ:
**Issue Description**
When calling `readdir()` on a path that is a file, not a directory, we get an internal error on the server, triggering an alarm.

**Steps to Reproduce**
In the Terminal:
- `touch foo.txt`
- `ls foo.txt`

Otherwise, if `/username/foo.txt` exists and is a file, run `puter.fs.readdir('/username/foo.txt')`.

**Expected & Actual Behavior**
The expected behaviour is that an error is returned to the user, but it should be something like `ENOTDIR`, not an internal server error.

**Addition Information or Screenshots (if applicable)**

Error details:
```
TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    --- property 'win32' closes the circle
    at JSON.stringify (<anonymous>)
    at stringify (/home/sam/Projects/HeyPuter/puter/node_modules/express/lib/response.js:1159:12)
    at ServerResponse.json (/home/sam/Projects/HeyPuter/puter/node_modules/express/lib/response.js:272:14)
    at ServerResponse.send (/home/sam/Projects/HeyPuter/puter/node_modules/express/lib/response.js:163:21)
    at /home/sam/Projects/HeyPuter/puter/packages/backend/src/routers/filesystem_api/readdir.js:116:9
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async errorHandledHandler (/home/sam/Projects/HeyPuter/puter/packages/backend/src/api/eggspress.js:183:14)
- error: TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    --- property 'win32' closes the circle
    at JSON.stringify (<anonymous>)
    at stringify (/home/sam/Projects/HeyPuter/puter/node_modules/express/lib/response.js:1159:12)
    at ServerResponse.json (/home/sam/Projects/HeyPuter/puter/node_modules/express/lib/response.js:272:14)
    at ServerResponse.send (/home/sam/Projects/HeyPuter/puter/node_modules/express/lib/response.js:163:21)
    at /home/sam/Projects/HeyPuter/puter/packages/backend/src/routers/filesystem_api/readdir.js:116:9
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async errorHandledHandler (/home/sam/Projects/HeyPuter/puter/packages/backend/src/api/eggspress.js:183:14)
- url: '/readdir'
[...]

PR #492, opened 2024-06-24
fix: Throw an error when readdir is called on a non-directory

AtkinsSJ:
This replaces a72ec9799ac3bd76ceafa22cce149e373a13f3b9 as a fix for #417.

That change meant that calling readdir on a file path would return a single stat entry, for that file. The problem is, to callers that looks like the file is a directory containing itself. This was causing problems in isomorphic-git, as it would readdir `foo.txt`, receive the listing of `[ {path: foo.txt', ... } ]`, and then try to read `foo.txt/foo.txt` and get confused that the file doesn't exist.

This change instead returns an error equivalent to ENOTDIR to the caller, which matches the behaviour of Node.js and Posix.

cc @KernelDeimos

### DIFF
--- a/packages/backend/src/api/APIError.js
+++ b/packages/backend/src/api/APIError.js
@@ -236,6 +236,10 @@ module.exports = class APIError {
             status: 422,
             message: 'Directory is not empty.',
         },
+        'readdir_of_non_directory': {
+            status: 422,
+            message: 'Readdir target must be a directory.',
+        },
 
         // Write
         'offset_without_existing_file': {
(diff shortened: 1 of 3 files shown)


KernelDeimos:
> This was causing problems in isomorphic-git, as it would readdir foo.txt, receive the listing of [ {path: foo.txt', ... } ], and then try to read foo.txt/foo.txt and get confused that the file doesn't exist.

I was really hoping I could come to the conclusion that we don't need to endpoints that stat filesystem nodes, but it looks like this elegance breaks other software so I'm going to merge this fix.

AtkinsSJ:
> I was really hoping I could come to the conclusion that we don't need two endpoints that stat filesystem nodes, but it looks like this elegance breaks other software so I'm going to merge this fix.

Yeah that would definitely be convenient! Maybe there's some other way of signalling that it's not a directory.

PR #493, opened 2024-06-24
Fix eslint issues

AtkinsSJ:
The issues with `mods/` were just because that directory wasn't registered in the eslint config, so it didn't know what globals were allowed.

The others were just missing `window.` in places.

It's probably worth either using an IDE plugin for eslint so these show up while writing code, or disabling the `no-undef` lint error. I think it's useful but I seem to be in a minority, while you two just get frustrated by it. :sweat_smile: 

cc @jelveh @KernelDeimos

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -28,6 +28,7 @@ export default [
         // Back end
         files: [
             "packages/backend/**/*.js",
+            "mods/**/*.js",
             "dev-server.js",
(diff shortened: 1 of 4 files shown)


KernelDeimos:
I actually like the no-undef errors, but not so much the maintenance required. Unfortunately it seems that we can't have one without the other so I appreciate this fix.

PR #498, opened 2024-06-26
(Partly) stop background apps taking focus from the terminal

AtkinsSJ:
This is a slightly hacky solution which doesn't work 100% of the time, but it's a lot nicer to use the terminal with git with these changes. (Previously, the terminal would lose focus after *every git command* which gets annoying fast.)

What stops this being a full fix is that the windowing code doesn't know if a window is a background window or not. They still go into the `window_stack`, and so when a window closes (like `git`'s) it focuses the next window on the stack, and if that's `phoenix` then the terminal loses focus.

I started playing with a Service for managing windows, but that's a much bigger task and I didn't want to get *too* distracted. :sweat_smile: I think that can be a lot neater though.

### DIFF
--- a/packages/terminal/src/main.js
+++ b/packages/terminal/src/main.js
@@ -117,5 +117,6 @@ window.main_term = async () => {
     const ioController = new XTermIO({ term, pty });
     ioController.bind();
 
+    window.addEventListener('focus', () => term.focus() );
     term.focus();
 };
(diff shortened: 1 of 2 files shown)


KernelDeimos:
This is pretty deep in gui logic for something that doesn't work 100% of the time; I don't understand the full ramifications of this so I'll need @jelveh 's input. I think this should probably be fixed properly without any intermediate fixes.

Skandiakiwi13 (approved):

Issue #512, opened 2024-07-01
Having problems when try close Start Menu on mobile devices


guicazaroto:
### Description
When I try to close the navigation menu on my smartphone (Xiaomi Redmi 13 Pro), it doesn't work because the search input is overlapping the X button.

### How to reproduce
Open Puter using a mobile device.
In the taskbar, click on "Start".
Try to open the start modal.

### Snapshot
<img width="349" alt="Screenshot 2024-07-01 at 7 24 40 PM" src="https://github.com/HeyPuter/puter/assets/9921030/8dc8b0c0-8258-43a5-aafb-0d7e45dd4adb">

guicazaroto:
I'm adding a small fix to solve this issue here: https://github.com/HeyPuter/puter/pull/513/files

As future improvement we could use relative units of measurement like (rem, em, %, vw, vh, ch, ex), instead of just use px. In this way we could make it work for different devices easily

jelveh:
Oops! We fixed it at the same time!

PR #513, opened 2024-07-01
fix:  close start menu on mobile devices

guicazaroto:
It fixes [Issue 512](https://github.com/HeyPuter/puter/issues/512)

### DIFF
--- a/src/UI/UITaskbar.js
+++ b/src/UI/UITaskbar.js
@@ -90,7 +90,7 @@ async function UITaskbar(options){
             
             let apps_str = '';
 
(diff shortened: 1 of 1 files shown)


jelveh:
Thank you for fixing the issue. We did it at the same time!
The fix I submitted takes into account whether the device is a phone or not by creating a wrapper class around the input and referencing it in the CSS.: 
```css
.device-phone .launch-search-wrapper{
    margin-top: 15px;
}
```
Basically preventing an unnecessary margin when it's not needed (e.g. on desktops).

Thank you for reporting the issue and submitting a PR to address it. This time, I need to keep my changes, but please do continue to submit PRs. They are much appreciated!

PR #598, opened 2024-07-16
Translate en.js as of 2024-07-10

ohm69:
Translated to Polish from English version. Apart from adding more than a hundred new translations, I have modified a few of existing ones to be more uniform with the rest of translations.


jelveh:
Merged! Thanks again 🎉

PR #611, opened 2024-07-19
Fixed Tamil translation

Alfika07:
The Tamil translation was in the wrong place and wasn't added to the translation list.


jelveh:
Thank you for fixing this issue! 😇

Alfika07:
No problem!

BTW I found this by accident. I don't even speak Tamil.

jelveh:
Much appreciated. We used to have a CI action to catch these, Github acted so unstable with the actions that we had to remove it :/

PR #618, opened 2024-07-20
Add issue template for app ideas

secondtruth:
This should make it easier to track and organize ideas for potential Puter apps.

**Important:** This issue template requires an "app" label to be present.

Maybe "potential app" is more specific and clearer? Just let me know and I'll update this PR.


jelveh:
This is a good idea, thanks for the suggestion!

PR #621, opened 2024-07-20
i18n: Retain 'confirm' keyword for account deletion across languages

pvhuwung:
## Changes

- Reviewed and updated all language files to ensure the word `'confirm'` is used consistently for account deletion confirmation.
- Updated the `type_confirm_to_delete_account` string to retain `'confirm'` in all languages.
- Ensured that the surrounding text is properly translated while keeping `'confirm'` as is.

## Rationale
- Keeping `'confirm'` as a universal keyword across all languages serves several purposes:
- Maintains a consistent user experience across different locales.
- Reduces the risk of accidental account deletions due to misunderstandings in translation.
- Aligns with common practices in software internationalization for critical actions.

### DIFF
--- a/src/gui/src/i18n/translations/ar.js
+++ b/src/gui/src/i18n/translations/ar.js
@@ -295,7 +295,7 @@ const ar = {
 	    two_factor_disabled: "تم تعطيل المصادقة الثنائية",
 	    two_factor_enabled: "تم تمكين المصادقة الثنائية",
         type: "نوع",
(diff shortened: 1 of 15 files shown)


jelveh:
Hi @pvhuwung, we think it might be difficult for a non-English speaking end-user to use an English word (or English keyboard to type in a word they may not understand) to confirm a very dramatic action like deleting their account. So we decided to make the account deletion work with `confirm` translated into different languages. That said, if a user types in `confirm` right now but their Puter language is not English, the action will still take place.

Thanks again for all your help. Let me know if you have any questions.

PR #631, opened 2024-07-24
Create config.yml to add Puter Discord to the issue screen

Zac0511:
This adds a config file which will show the Puter Discord Server in the issue menu, allowing users to quickly ask help on the Puter Discord.

Also, the if `true` of `blank_issues_enabled: true` is replaced with `false`, users will be forced to use a template, and won't be able to create a custom issue.


KernelDeimos:
Great idea, thanks!

PR #632, opened 2024-07-25
Update README.md

Zac0511:
I added 2 shields from [shields.io](https://shields.io/).
Those will display the repository size and the latest version number in the readme.
They automatically update themself, so there is no need to edit README every time 1 byte is added to the repo.

I thinked this would be useful, so i made this pull request


Zac0511:
I just updated it to also show the license at the top, which also updates if the licence gets modified.

jelveh:
Pretty cool! Thanks @Zac0511.

PR #633, opened 2024-07-25
Add badges

Zac0511:
I added badges, used for when an app is **available on Puter**, or is **powered by Puter.js**.
The source Figma files are also included.


jelveh:
Thank you @Zac0511, looking amazing!

Issue #541, opened 2024-07-06
Complete Korean translation based on English file


jelveh:
The Korean translation file (`ko.js`) needs to be updated to include all translations present in the English file (`en.js`). Currently, the Korean translation may be incomplete or outdated compared to the English version.

### Steps to resolve
1. Compare the contents of `src/gui/src/i18n/translations/en.js` and `src/gui/src/i18n/translations/ko.js`
2. For each translation key in `en.js`, ensure a corresponding translation exists in `ko.js`
3. If a translation is missing in `ko.js`, add it with an appropriate Korean translation
4. If any translations in `ko.js` are outdated or incorrect, update them to match the current English version

### Additional notes
- Please ensure that the translations are culturally appropriate and maintain the intended meaning of the English text
- If certain terms or phrases don't have a direct Korean equivalent, consider adding translator notes or using transliteration where appropriate

### Files to be updated
- `src/gui/src/i18n/translations/ko.js`

### Related files
- `src/gui/src/i18n/translations/en.js` (reference file)

Thank you for helping to improve the Korean localization of our project!

jelveh:
closed via #648

PR #650, opened 2024-08-01
Updated Korean translation file

de1eb:
Completed Korean translation file
#541

By the way, while I check texts by using puter, I found some texts doesn't exists on en.dictionary (so, ko.dictionary as well).

Texts / Location where I found texts
1. Found no suitable apps to open this file with. Would you like to download it instead? / when I try to open unknown file
2. Enter Full Screen / Button on Upper right side of Desktop
3. QR Code / Button on Upper right side of Desktop
4. My Websites / My Websites Window Title
5. used of / Below (Settings - Storage Usage)
6. Save account before logging out! / Window contents when I try to log out temporary account
7. You are using a temporary account and logging out will erase all your data. / Window contents when I try to log out temporary account
and..
contents in mail for email verification

I also found some bugs
1. I can't use 'Contact US' right after my email address is verified. I should relogin to use it.
2. Storage usage limitation doesn't work when I copy existing file on cloud.


jelveh:
Thank you for this and for the bug reports/feedback. I'll open separate issues for them ✌️

PR #685, opened 2024-08-14
translate README.md to spanish

juank0205:
Created a new file ```./doc/i18n/README.es.md``` with the translated content while keeping the original format and content.


jelveh:
Thank you very much for your contribution. Merging now 🎉

Issue #696, opened 2024-08-17
Add configuration to disable new user sign ups


stevenkampen:
_**Note:** this issue has been edited by Puter staff_

### Good First Issue

This issue is good for new contributors.
- Carefully read the instructions in this report to learn how to solve this issue.
- Be sure to follow the [PR Standards](https://github.com/HeyPuter/puter/blob/main/CONTRIBUTING.md#pr-standards).

<hr>

Original report by [stevenkampen](https://github.com/stevenkampen)

> I would like a config parameter to disable new signups so I can prevent random people creating accounts on my server if I happen to deploy it publicly.

<hr>

Instructions by [KernelDeimos](https://github.com/KernelDeimos)

### Changes Required
- Add default config value for `temp_users` in [config.js](https://github.com/HeyPuter/puter/blob/main/src/backend/src/config.js) (maybe `false`?)
- [AuthService](https://github.com/HeyPuter/puter/blob/main/src/backend/src/services/auth/AuthService.js) should register a config feature flag [as ACLService does](https://github.com/HeyPuter/puter/commit/4561b8937de025471c2dfb1771465d779cefab5d#diff-655355b9f2519b04d9f3f049ad84e3a57a0f75f9b768b7c35ea283de3d174d8fR28-R34), called `temp-users`
  - this is so the config is available to gui and not just backend
- Backend should disallow temporary user creation when `config.temp_users` is `false.`
  - temporary users are created [when no email or password is provided in the request](https://github.com/HeyPuter/puter/blob/cdc394758517f0958002ce66d405fa093b1a8897/src/backend/src/routers/signup.js#L136-L139) (also `req.body.is_temp` has to be true, which is a little redundant so this could be cleaned up)
- GUI should not attempt to create temporary users when the `temp-users` feature flag is disabled
  - There's an example of a feature flag being checked [in UIItem.js](https://github.com/HeyPuter/puter/blob/cdc394758517f0958002ce66d405fa093b1a8897/src/gui/src/UI/UIItem.js#L785)
  - [This is the condition in GUI that makes temp users](https://github.com/HeyPuter/puter/blob/cdc394758517f0958002ce66d405fa093b1a8897/src/gui/src/initgui.js#L813-L826)

KernelDeimos:
This has been requested before, I'll add this soon if nobody beats me to it

OluwaDraco:
HI, I would love to help out with this!

OluwaDraco:
Hi,ill love to help with this

[...]

PR #701, opened 2024-08-19
issue#696 solved

iam-subho:
in line no 39 of  src/gui/src/UI/UIWindowLogin.js  a boolean variable added  showSignupButton
true means signup button will be visible and false means signup button will be disable

### DIFF
--- a/src/gui/src/UI/UIWindowLogin.js
+++ b/src/gui/src/UI/UIWindowLogin.js
@@ -7,12 +7,12 @@
  * it under the terms of the GNU Affero General Public License as published
  * by the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


KernelDeimos:
Hello, thanks for this contribution. There are a few changes that need to be made before we can merge this:

1. **Configuration:** end-users won't know how to edit the code to change this property, so it needs to be a configuration value instead. A configuration value can be passed to gui using something called a "feature flag". Take a look at [the one recently added to ACLService](https://github.com/HeyPuter/puter/commit/4561b8937de025471c2dfb1771465d779cefab5d#diff-655355b9f2519b04d9f3f049ad84e3a57a0f75f9b768b7c35ea283de3d174d8fR28-R34) - this would be similar, except it would probably be registered by AuthService instead.
2. **Backend functionality:** in its current state, this change might give users a false sense of security if they don't realize someone can still create accounts on their server by manually sending HTTP requests. The `signup.js` file will need to be updated to not create temporary users on this condition.

KernelDeimos:
Hello, I'm going to close this PR since we would like to implement this differently. I added instructions to #696 describing the ideal implementation.

PR #702, opened 2024-08-20
add Chinese, add language tag

Sam-SSTU:
add Chinese, add language tag
Issue: #665


AJcompilenet:
Yes please do help

jelveh:
Thank you for your contribution, @Sam-SSTU.
Just a quick note: you may have noticed that I removed the language badges. We've decided not to use badges and use text links instead so that users can find the language link using the search utility of the browser (badges are images and won't show up if user types in the search box of the browser)

Thanks again, merging now 😇

PR #708, opened 2024-08-23
Add README.id.md

fauzipadlaw:
Solved #676


jelveh:
Thank you for another great contribution and for keeping the Indonesian version of Puter updated. Merging 🫡

PR #715, opened 2024-08-25
Added installation instructions for public hosting to readme

decipher2k:
Added installation instructions for public hosting to readme

### DIFF
--- a/src/backend/src/routers/signup.js
+++ b/src/backend/src/routers/signup.js
@@ -65,273 +65,281 @@ module.exports = eggspress(['/signup'], {
         body: req.body,
     });
 
-    // check bot trap, if `p102xyzname` is anything but an empty string it means
-    // that a bot has filled the form
-    // doesn't apply to temp users
-    if(!req.body.is_temp && req.body.p102xyzname !== '')
-        return res.send();
-
-    // check if user is already logged in
-    if ( req.body.is_temp && req.cookies[config.cookie_name] ) {
-        const { user, token } = await svc_auth.check_session(
-            req.cookies[config.cookie_name]
+    if(config.registration_closed==true)
+    {
+        return res.status(403).send('Registration is closed.');
+    }
+    else
+    {
+    
+        // check bot trap, if `p102xyzname` is anything but an empty string it means
+        // that a bot has filled the form
+        // doesn't apply to temp users
+        if(!req.body.is_temp && req.body.p102xyzname !== '')
+            return res.send();
+    
+        // check if user is already logged in
+        if ( req.body.is_temp && req.cookies[config.cookie_name] ) {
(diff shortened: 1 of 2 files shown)


decipher2k:
> Thank you for this! I'll do a quick test and merge. We may need to put it in a separate document and link to it from the README...

I'm glad you like it.
The modification for opting out registration has  somehow gotten into the pull request.

src/backend/src/routers/signup.js
```diff
@@ -65,273 +65,281 @@ module.exports = eggspress(['/signup'], {
...
-                    email_confirmed: user.email_confirmed,
-                    requires_email_confirmation: user.requires_email_confirmation,
-                    is_temp: (user.password === null && user.email === null),
-                    taskbar_items: await get_taskbar_items(user),
-                }
+    if(config.registration_closed==true)
+    {
+        return res.status(403).send('Registration is closed.');
+    }
+    else
```
KernelDeimos: We don't need an `else` if there's an early return. Removing this will reduce indentation changes and make the PR easier to review.

KernelDeimos (changes requested):

KernelDeimos:
We have some documentation for this under [Self-Hosting instructions](https://github.com/HeyPuter/puter/blob/main/doc/self-hosters/instructions.md) but we don't have this linked in the README.md file anymore. If you have the time it would be nice to cross-reference what you wrote and what's already there and try to pick the best from both. Ideally all documentation for this should be in `doc/self-hosters` and we will eventually link this in README.md.

decipher2k:
> We have some documentation for this under [Self-Hosting instructions](https://github.com/HeyPuter/puter/blob/main/doc/self-hosters/instructions.md) but we don't have this linked in the README.md file anymore. If you have the time it would be nice to cross-reference what you wrote and what's already there and try to pick the best from both. Ideally all documentation for this should be in `doc/self-hosters` and we will eventually link this in README.md.

Sure, no problem.

src/backend/src/routers/signup.js
24high: updated

KernelDeimos:
> > We have some documentation for this under [Self-Hosting instructions](https://github.com/HeyPuter/puter/blob/main/doc/self-hosters/instructions.md) but we don't have this linked in the README.md file anymore. If you have the time it would be nice to cross-reference what you wrote and what's already there and try to pick the best from both. Ideally all documentation for this should be in `doc/self-hosters` and we will eventually link this in README.md.
> 
> Sure, no problem.

Thanks, diff looks much better now! We'll be able to merge the signup change quicker than the README.md update, but I don't want you to lose credit from us cherry-picking it so if you have time to make a separate PR for it that might be better (sorry about all the git chores btw, I should've brought that up yesterday)

PR #746, opened 2024-08-31
No need word repetition for gambar  in Indonesian

fauzipadlaw:
Screenshot 2024-08-31 at 23 45 56

The literal translation for Pictures IS Gambar-gambar, but after reviewing how it looks on live demo Puter, this translation seems a bit off, since other word like Videos is not Video-video, documents is not Dokumen-dokumen.

So, I think it's better to just translate Pictures to Gambar only. Indonesian actually does not care about singular and plural. If we really need to emphasize it plural, we just repeat the word. Which in this context, it is not that necessary to emphasize it.


jelveh:
Thank you for this. Great example of taking the context into account for i18n. ✌️

PR #782, opened 2024-10-02
CI: Add a simple test workflow

bitsnaps:
Just to make sure everything works fine on all major node versions (>= v18)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,28 @@
+name: test
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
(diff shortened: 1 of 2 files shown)


KernelDeimos:
Thanks for your contribution! I'm going to leave this open for a bit to give us time to get the `npm run test` tests working. The backend tests should also be added here but I'm unsure if they currently follow the right conventions (i.e. exit code) to be included here.

Backend tests are run with: `node src/backend/tools/test.js` - that script invokes a "test kernel" which registers all the services (just like Puter's kernel does), and runs the `_test` method on the ones which implement it.

KernelDeimos:
I updated the text command a while ago and it should be working, so I'm gonna try this out now.

Issue #783, opened 2024-10-02
Play the Puter chime at startup


Zac0511:
It would be cool if the [Puter chime](https://github.com/HeyPuter/puter/blob/main/src/gui/src/audio/puter_chime.mp3) was played at Puter's startup.

Of course, the user could disable it from the `personalization` menu in settings if they don't want it.

Here's what it could look like if that feature was added : 

https://github.com/user-attachments/assets/797a3346-347e-461a-ae33-ff745d9975b6

(make sure you enabled sound on the video)

jelveh:
I love this! How did you manage to play the chime without clicking? That's our only issue right now!

Zac0511:
> I love this! How did you manage to play the chime without clicking? That's our only issue right now!

Well i just recorded a video of me opening Puter and I added the sound afterwards with editing.

KernelDeimos:
We talked about having this play after manual login since a dom event can be triggered for that. Otherwise, we just won't play a sound. This is a good task for a new contributor, so I'm going to wait and see if anyone else wants to take it on.

4nshuman:
Hey @KernelDeimos,
I'd like to take a look into this issue. could you please assign this to me ?
Also do you have any specific chime to implement ?

KernelDeimos:
> Hey @KernelDeimos, I'd like to take a look into this issue. could you please assign this to me ? Also do you have any specific chime to implement ?

Sure thing! The chime is at: [src/gui/src/audio/puter_chime.mp3](https://github.com/HeyPuter/puter/blob/main/src/gui/src/audio/puter_chime.mp3)

Keep in mind, there's slightly annoying issue you'll encounter where audio is not allowed to play until the user clicks something. For this reason, we can only really support a good experience with it when the user starts at the login page and clicks "login".

There's a new "welcome screen", we could also maybe play it there in situations where the chime hasn't already been played at that point. There might be other good situations for it, let us know if you come up with something else.

KernelDeimos:
@4nshuman hi, are you still working on this?

4nshuman:
Hey @KernelDeimos,

Yup I'm still on it. Actually I was tinkering with a couple of solutions to try and play the sound as soon as everything loads up on the page rather than the first user interaction.
I have had some success but still experimenting.

4nshuman:
hey @KernelDeimos,
I've raised a PR, could you please take a look.

Zac0511:
Just noticed it got actually added, i'm closing this issue

PR #831, opened 2024-11-09
Added a chime to play on user's first interaction

4nshuman:
This PR solves the issue : #783 

The idea is to have a global chime playback function that can be called at any moment.
In this particular PR, the global chime playback is called upon on user's first interaction, this is to mimc the boot up of an OS.
Once the playback is through, the event listener will be deactivated so as to not play back the chime on subsequent clicks.

### DIFF
--- a/src/gui/src/initgui.js
+++ b/src/gui/src/initgui.js
@@ -150,6 +150,11 @@ if(jQuery){
     };
 }
 
+window.playStartUpChime = ()=>{
+    let audio = new Audio('/src/audio/puter_chime.mp3');
+    audio.play();
+}
+
 window.initgui = async function(options){
     let url = new URL(window.location);
     url = url.href;
--- a/src/gui/src/initgui.js
+++ b/src/gui/src/initgui.js
@@ -153,7 +153,6 @@ if(jQuery){
 window.playStartUpChime = ()=>{
     let audio = new Audio('/src/audio/puter_chime.mp3');
     audio.play();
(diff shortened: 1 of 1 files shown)


jelveh:
Woooohooo! thank you @4nshuman . Will test ASAP!

KernelDeimos:
@4nshuman Hey, I just merged this, but we might actually need to revert it; here's why:

- User login with `[Enter]` key does not trigger the boot chime
- Clicking login button doesn't trigger the boot chime
- If the page has already been open for some time (just a few seconds, really) the chime should not play as it ~might~ will confuse the user

KernelDeimos:
Apparently there's no way to re-open a PR, which I personally think is a design flaw in GH but what do I know. @4nshuman  please open a new PR for this when you're ready; you should be able to re-use the same branch.

4nshuman:
@KernelDeimos so here's the updated acceptance criteria -
1. Login user flow should also trigger the chime
1a - when the user logs in the chime should be triggered.
1b - the user might login using the keyboard (enter key). This too should trigger the chime.

2. There should be a minimum threshold to a chime playback. If the threshold is crossed the chime should not play.

On point 2 what do you think is a viable threshold ? Does 5 seconds work ?

4nshuman:
Hey @KernelDeimos, @jelveh,
Any updates on the queries ?
Am I missing something ?

4nshuman:
Thanks for the confirmation @jelveh 
I'll take a look into the delay implementation on chime.

PR #908, opened 2024-11-24
Update fa.js

AmirH-Sal:


jelveh:
Thank you very much, sir. Historically, Farsi was lagging behind, but not anymore!! 🙌

PR #921, opened 2024-11-27
Update id.js #865

junian:
Fix Indonesia translation #865


jelveh:
Thank you very much for this!

PR #929, opened 2024-11-28
Added German translation

ahamiltonnn:
#874 
I added the German translations to the undefined words.


jelveh:
Hi @ahamiltonnn , thank you for you contribution. A few of the terms contain placeholders (`%strong%`) that are missing in the PR. Could you look into them before I can merge your PR?

Issue #880, opened 2024-11-20
Enable Save button in Dev Center only when app settings have changed


jelveh:
Currently in the [Dev Center](https://puter.com/app/dev-center), the Save button in app editing mode is always enabled, regardless of whether any changes have been made. This can lead to confusion about whether changes need to be saved and unnecessary API calls.

All the relevant code resides in [`dev-center.js`](https://github.com/HeyPuter/puter/blob/main/src/dev-center/js/dev-center.js)

### Technical Requirements:
1. Track original values of all editable fields when app editing form is loaded
2. Enable Save button only when current values differ from original values
3. Disable Save button when:
   - No changes have been made
   - After successful save (values become the new original state)
   - When changes are reverted to original values

### Fields to Track:
- Title
- Name
- Index URL
- Description
- Icon
- File Associations
- Window settings (height, width, position)
- All checkboxes (background, maximize on start, resizable window, etc.)

and any other fields in the settings form...

### Design Considerations:
- Compare values on each field change
- Handle special cases like icon changes (base64/URL comparisons)
- Consider adding a "Reset" button to revert all changes

ria147:
Hello @jelveh , 

I am new to this repository. Looking to contribute as part of my Software Engineering course project for my bachelor's degree. Is this issue still open and in need of work? 

If so, could you please assign it to me? Thank you

jelveh:
Hi @ria147,

Thank you for your interest in contributing to Puter. Just assigned the issue to you. Let us know if you have any questions or need help.

jelveh:
Hi @ria147, did you have any luck with this? Let me know if you need any help.

ria147:
Hi @jelveh , thank you for checking in! I am trying to run things locally, and when I click on the Dev Center, this is what it looks like as of now for me, so can't replicate the issue locally. However this doesn't happen in the deployed version, where I am able to replicate the issue. Can you help me understand if I'm doing something differently, or am I missing something? 

I forked main, ran npm install and npm start.
![image](https://github.com/user-attachments/assets/b0c1ed93-a4ea-4388-b2a8-f2116741aab7)

jelveh:
Could you send me a screenshot of your browser's dev console? Also, is this an incognito window? if not, can you try in an incognito session?

jelveh:
Copying any errors here would be helpful.

ria147:
Tried it in incognito, still not able to replicate the issue locally. 
Here is errors shown on the dev console
[...]

PR #934, opened 2024-11-29
Enable Save button in Dev Center only when settings change and add a Reset button

ria147:
Issue: #880 

Files changed: `src/dev-center/js/dev-center.js` and `src/dev-center/css/style.css`

Now all original fields (input, checkbox, etc.) on the settings form is tracked, and also added a Reset button. The Save button and Reset button are only enabled if and only if there is a change in the form from the original. 
- Save and Reset buttons are disabled upon first opening settings form (since no changes initially)
- if user changes anything (eg. title from "My-App" to "My-App-2") -> Save button is enabled. Reset button is enabled
- if user clicks on Reset button, it reverts all the form changes to the original values before editing.
- if user changes back to original (eg. title back from "My-App-2" to "My-App") -> Save button is disabled. Reset button is disabled.

**Before any changes**
before-any-changes
**After any changes**
after-any-changes

Manual tests carried out on locally deployed frontend to ensure everything works as expected.

### DIFF
--- a/src/dev-center/css/style.css
+++ b/src/dev-center/css/style.css
@@ -951,6 +951,12 @@ ol li:before {
     margin-top: 20px; 
     margin-bottom: 20px;
 }
+.edit-app-reset-btn{
+    float: right; 
+    margin-top: 20px; 
+    margin-bottom: 20px;
+    margin-right: 10px;
+}
 input:read-only {
     background-color: rgb(242 242 242);
 }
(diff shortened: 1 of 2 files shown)


jelveh:
awesome! Thank you, @ria147. Testing :)

jelveh:
Merged! Thank you @ria147 for your contribution!

PR #939, opened 2024-11-29
Update it.js

danielpetrica:
I updated the missing strings and corrected other string which where not translated to italian


danielpetrica:
this pull request closes this issue https://github.com/HeyPuter/puter/issues/863

Issue #878, opened 2024-11-20
Warn users when deploying `.git` directory in Dev Center


jelveh:
Currently the [Dev Center](https://puter.com/app/dev-center) deploys all files and directories without warning users about potentially sensitive or unnecessary files like `.git` directories. This could lead to security risks and unnecessarily large deployments.

All the relevant code resides in [`dev-center.js`](https://github.com/HeyPuter/puter/blob/main/src/dev-center/js/dev-center.js)

### Technical Requirements:
1. Before deployment, scan the files/directories being deployed for `.git` directory
2. If a `.git` directory is found, show a warning dialog that includes:
   - Clear explanation that deploying `.git` directories may expose sensitive information
   - Warning about increased deployment size
   - Option to continue or cancel the deployment 
   - "Don't show this warning again" checkbox
3. Store the user's preference in [KV store](https://docs.puter.com/KV/set/) if they choose not to see the warning again.

### Design Considerations:
- Warning should not block deployment if user has chosen to disable it
- Consider showing size of `.git` directory in the warning
- Should work with both local file drops and Puter file selections

paulinezhanghh:
Hi @jelveh, I'm new to this project but am very interested in working on this issue. Can I take this?

jelveh:
Hi @paulinezhanghh. Sure thing! I think this is an excellent 'first time issue' to get started with Puter. Assigned.
Let me know if you need any help :)

paulinezhanghh:
> Hi @paulinezhanghh. Sure thing! I think this is an excellent 'first time issue' to get started with Puter. Assigned. Let me know if you need any help :)

Hi! I have a follow-up question about testing. Do you have any preferences or details about how I should test this?

jelveh:
Manual testing would suffice. 

- Deploys containing `.git` subdirectory (at any depth) should warn the user
- Deploys not containing .git should go ahead as expected
- test if "Don't show this warning again" checkbox works.
- test if previous deploy scenarios are still working (no regression)
- more test cases?

jelveh:
Thanks again for taking care of this. Merged and completed. Closing this issue :)

paulinezhanghh:
> Thanks again for taking care of this. Merged and completed. Closing this issue :)

Thank you so much for the quick response but I haven't added the test cases. If you would like to let me add more test cases to enhance the robustness of this feature, I would love to do so!

PR #980, opened 2024-12-03
Added the warning dialog

paulinezhanghh:
If the system detects any sensitive files, it will pop up the dialog showing the warning and provide the user with options to continue or not. However, the checkbox is not working right now and I will try to fix it ASAP and enhance the UI design. Issue can be found here: #878
Screenshot 2024-12-03 at 18 15 40


jelveh:
Thank you for your contribution. Works great!

PR #982, opened 2024-12-04
tweak: added missing hindi translation

rowin-C:
- added missing hindi tranlastion
- added transliteration where translation is meaningless or not present


jelveh:
Thank you @rowin-C for another awesome contribution! Merged.

Issue #944, opened 2024-11-30
Toggling apps to minimize once it clicked


Ayush412-art:
In operating systems like Windows, when we open an application, there are two ways to minimize it: either by clicking the Minimize button (located at the top-right corner) or by clicking on the application's icon (at the bottom of the screen). This enhances the user interface and improves the overall user experience

mariatouil:
Hi! I'd like to work on this issue. Could you please assign it to me?

jelveh:
This makes sense, but only if there is one window open of that specific app (similar to Windows/MacOS behaviour)

@mariatouil appreciate the interest to pick this up, before I assign this to you and updates on the other issue assigned? I don't want to overwhelm you with issue assignments.

arthur-adriansens:
I added it ;) and I added fun animations for taskbar icons!

jelveh:
Donzo by @arthur-adriansens 🫡

PR #1005, opened 2024-12-07
Close apps by clicking on icons (issue #944) + awesome animations!

arthur-adriansens:
I added the app-request from issue #944, and I added a small clicking animation on the taskbar icons, like on MacBooks or Windows

### DIFF
--- a/src/gui/src/UI/UITaskbarItem.js
+++ b/src/gui/src/UI/UITaskbarItem.js
@@ -67,7 +67,10 @@ function UITaskbarItem(options){
     // fade in the taskbar item
     $(el_taskbar_item).show(50);
 
-    $(el_taskbar_item).on("click", function(){
+    $(el_taskbar_item).on("click", function(e){
+        e.preventDefault();
+        e.stopPropagation();
+
         // If this item has an open context menu, don't do anything
         if($(el_taskbar_item).hasClass('has-open-contextmenu'))
             return;
--- a/src/gui/src/UI/UITaskbarItem.js
+++ b/src/gui/src/UI/UITaskbarItem.js
@@ -73,8 +73,16 @@ function UITaskbarItem(options){
             return;
 
         if(options.onClick === undefined || options.onClick(el_taskbar_item) === false){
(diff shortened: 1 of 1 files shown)


arthur-adriansens:
There's only one small bug I can't seem to fix. In a very specific case, when you open Explorer and you haven't interacted with the window (which doesn't happen often), you have to click twice before the window hides. This problem happens **only with Explorer**, in all other apps you can hide the window on the first icon click. 

I found the cause though: for some weird reason, the "window-active" class is already there on all apps, except for when you open Explorer and didn't do anything with the window, then the window-active class is being removed somewhere. _I spend a long time searching, but didn't find it._

jelveh:
Looks pretty awesome! I'm testing and will merge soon

arthur-adriansens:
Nice, thank you so much! I'm searching for new ideas in the issues tab :)

jelveh:
> I found the cause though: for some weird reason, the "window-active" class is already there on all apps, except for when you open Explorer and didn't do anything with the window, then the window-active class is being removed somewhere. I spend a long time searching, but didn't find it.

Thanks for bringing this to my attention. I'm trying to fix and push the patch for this and then merge everything.

jelveh:
Fixed. Interesting bug: When a taskbar item is clicked, the click is also propagated to the desktop and clicking on a desktop would naturally take away the focus from all windows. This only happened with Explorer because the click would immediately bubble whereas with other apps, the app had to be opened first which takes, at the very least, a few milliseconds and by that time the app has taken the focus back.

arthur-adriansens:
Nice! Thanks for the help, and sorry for creating an extra issue. I, too, noticed that the desktop was being clicked, and I tried to move the taskbar out of the desktop div, but I was scared something would be dependent on it being inside the desktop div 😅. Didn't realize that it was actually the cause!

jelveh:
I'm just happy the fix was actually simple. Was fearing the worst 👀
This has also fixed a few downstream issues as well, such as the window stack order not being corrupted which is important. Thanks again ✌️

PR #1013, opened 2024-12-08
Added missing Italian translations (#968)

CriDM:
#968 from https://www.upwork.com/ 
Cristian Della Monica


jelveh:
Hi @danielpetrica, could you look into CLA before I can review and merge?

CriDM:
Now, how works the payment?

Issue #1030, opened 2024-12-09
Bug: message was cut off in the middle of conversation


devanenWl:
### Issue Description

When I ask something which requires to generate long message, it will be cut off although it hasn't finished yet. I think it because of the hard-coded 1000 max_tokens
```js
const completion = await this.anthropic.messages.stream({
    model: model ?? this.get_default_model(),
    max_tokens: 1000,
    temperature: 0,
    system: PUTER_PROMPT + JSON.stringify(system_prompts),
    messages: adapted_messages,
});
```

### Steps to reproduce
```html
<html>
<body>
    <script src="https://js.puter.com/v2/"></script>
    <script>
    (async () => {
        const resp = await puter.ai.chat([{"role":"user","content":"Can you help me to create a chess game in React ?"}], {model: 'claude-3-5-sonnet', stream: true });
        for await ( const part of resp ) puter.print(part?.text?.replaceAll('\n', '<br>'));
    })();
    </script>
</body>
</html>
```

### Deployment

- [X] Production (puter.com)
- [ ] Development (`npm run start`)
- [ ] Docker (via `docker run`)
- [ ] Docker (via `docker-compose`)

PR #1031, opened 2024-12-09
feat: set max token to 8096

devanenWl:
#1030


KernelDeimos:
Yep, we can do this now that we track token usage. Thanks for the PR!

KernelDeimos:
How did you find the process of making this change? Is there any documentation we could add that would have made it easier?

Issue #983, opened 2024-12-04
Bug: Stuck on Creating new app loading screen


rowin-C:
### Issue Description

While creating a new app in dev center , it gets stuck on the loading modal.
If we then reload the page then loading modal is gone and we are in the "My apps" page

### Steps to reproduce

go to dev center and create a new app

### Expected behaviour

It should directly redirect to "My apps" page

### Addition Information or Screenshots (if applicable)

https://github.com/user-attachments/assets/fd006015-4097-4244-8826-6c02ab7d3ccd

https://github.com/user-attachments/assets/d3902086-4d5a-4de1-994a-2705073d6177

### Deployment

- [ ] Production (puter.com)
- [X] Development (`npm run start`)
- [ ] Docker (via `docker run`)
- [ ] Docker (via `docker-compose`)

rowin-C:
If anyone else could confirm it , then it would confirm the issue

jelveh:
Any error dumps in terminal/browser console?

rowin-C:
this error popped up.

![image](https://github.com/user-attachments/assets/a5bb9baf-6e29-4d2c-a4af-869b563eab5f)

![image](https://github.com/user-attachments/assets/c3f847b8-cac1-40da-90e8-ae69a1080fb0)

mariatouil:
Hey,I just fixed this issue. Could you assign it to me, please?

mariatouil:
Hi,

I’ve submitted a Pull Request to address this issue: https://github.com/HeyPuter/puter/pull/1051.
Could you please review it and let me know if everything looks good?

Thank you in advance for your feedback!

jelveh:
This has been resolved

PR #1053, opened 2024-12-11
Revert "fix: resolve issue #983 - Stuck on Creating new app loading s…

jelveh:
…creen"


mariatouil:
@jelveh Hi, I noticed that the fix for issue #983 - Stuck on "Creating new app" loading screen was reverted in #1053. Could you please clarify why it was reverted? Was there an issue with the implementation or any other reason?

Issue #1033, opened 2024-12-10
Add app category display in Dev Center


jelveh:
Currently, the Dev Center doesn't display app categories in its interface. We should add category display both in the main app list and individual app settings pages to help developers better organize and identify their apps.

#### UI Changes Required

**Main App List:**
- Add a category label/badge next to each app name in the list
- Consider using subtle visual differentiation (like muted colors) to maintain readability
- Handle cases where category is undefined/empty gracefully (e.g., display "Uncategorized" or simply omit)
- Ensure proper alignment and spacing with existing app name and metadata

**Individual App Settings Page:**
- Add category field to the app's title or heading
- Real-time update of category display when edited

#### Technical Implementation

**Data Structure:**
- App categories are stored in the `metadata` attribute of app objects
- Example app object structure:
```javascript
{
  name: "my-app",
  metadata: {
    category: "productivity"
  },
  // ... other app properties
}
```

**Code Changes Required:**
1. Dev Center App List Component:
   - Update app list item rendering to include category
   - Add category styling/components
   - Handle category data fetching/display

2. App Update Logic:
   - Update local state/UI after successful category changes

#### UX Considerations
- Clear visual hierarchy: category should be visible but not dominate the app name
- Consistent styling across both list view and settings page
- Graceful handling of empty/undefined categories

#### Implementation Notes
- Files to modify are in `src/dev-center/`:

#### Testing Requirements
- Verify category display in app list
- Test category updates in settings
- Ensure proper handling of apps without categories
- Test UI responsiveness and layout
- Verify real-time updates after category changes

Raghavendra-Aprameya:
Hi jelveh,

This is a great idea, and I’d love to work on it. I’ve reviewed the requirements and have some thoughts on how to approach the implementation, including adding category labels in the app list and handling updates on the settings page.

If no one else is currently assigned, may I take this up? Let me know if there are any specific guidelines or expectations you’d like me to follow.

Looking forward to your response!

Best regards,  
Aprameya Raghavendra.

jelveh:
Thank you @Raghavendra-Aprameya, just assigned it to you. 

> Let me know if there are any specific guidelines or expectations you’d like me to follow.

So far only what's in the issue description. But if anything is vague, we could discuss it here.

Let me know if you need help with anything 🫡

mariatouil:
@jelveh Hi,
I’ve submitted a Pull Request to address this issue: https://github.com/HeyPuter/puter/pull/1059.
[...]

PR #1059, opened 2024-12-12
feat:Add app category display to Dev Center interface (#1033)

mariatouil:
Add app category display to Dev Center interface (#1033)

- Updated the main app list to include a category label/badge next to each app name
- Handled undefined or empty categories by displaying "Uncategorized" or omitting the label
- Added category field to individual app settings page, with real-time update after category changes
- Ensured consistent UI and styling for category display across both the main list and settings page
- Improved alignment and spacing for category labels to match existing app metadata layout
- Updated app data structure to support category metadata in the app objects

Testing:
- Verified category display in app list
- Tested category updates in settings
- Ensured graceful handling of apps without categories
- Checked responsiveness and UI layout on different screen sizes

image

image

### DIFF
--- a/src/dev-center/js/dev-center.js
+++ b/src/dev-center/js/dev-center.js
@@ -483,13 +483,19 @@ function applink(app) {
  * $('#edit-app').html(appEditHTML);
  */
 
+// Modify generate_edit_app_section function for category display
 function generate_edit_app_section(app) {
     if(app.result)
         app = app.result;
 
     let maximize_on_start = app.maximize_on_start ? 'checked' : '';
 
-    let h = ``;
+    const selectedCategory = app.metadata?.category || "";
+    const categoryOptions = APP_CATEGORIES.map(category => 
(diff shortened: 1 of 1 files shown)


jelveh:
hi @mariatouil thanks again for looking into this issue. This is not a complete implementation. For example, category is not shown in an app's settings page. Also, categories in app list are pushing the quick links for each app out of view (i.e. 'settings', 'add shortcut to desktop', etc)

mariatouil:
@jelveh hey ,
Thank you for your feedback! Regarding the category in the app's settings page, I want to clarify that I have already added the category to the application settings, as shown in the attached screenshot. Could you confirm if this is what you were referring to, or if there are other settings where the category should also appear?

As for the quick links being pushed out of view (e.g., 'settings', 'add shortcut to desktop', etc.), I didn’t fully understand the issue. Could you please elaborate a bit more or provide an example so I can address it properly?

image

Thanks again!

Issue #1071, opened 2024-12-19
Bug:  Success Message Not Displayed After Deployment


mariatouil:
### Issue Description

**Description**
The success message (#deploy-success-msg) that is supposed to appear after a successful deployment does not show up, even when the index.html file is properly deployed, and the process reaches 100%.

### Steps to reproduce

Drag and drop a valid index.html file into the deployment area.
Wait for the deployment process to reach 100%.
Observe that the success message does not appear.

### Expected behaviour

The success message (New version deployed successfully 🎉) should be displayed immediately after a successful deployment.

### Addition Information or Screenshots (if applicable)

Observed Behavior
Although the deployment seems to complete successfully, the success message does not appear.

https://github.com/user-attachments/assets/0e09e806-89fe-4995-a900-354f9beff4ff

### Deployment

- [ ] Production (puter.com)
- [X] Development (`npm run start`)
- [ ] Docker (via `docker run`)
- [ ] Docker (via `docker-compose`)

mariatouil:
@jelveh Hi, could you please assign this issue to me? I’ve fixed this issue as you can see in the record

https://github.com/user-attachments/assets/4be16326-b6a8-40bd-8a77-c0c3569385a3

PR #1072, opened 2024-12-19
Fix : the issue with the success message not displaying after deployment

mariatouil:
**Description:**
This pull request addresses the bug where the success message (#deploy-success-msg) was not displayed after a successful deployment, even when the process reached 100% ,addressing issue #1071

**Changes Made:**
Verified that the message appears consistently after deployment completion.

https://github.com/user-attachments/assets/4b58a3cb-fcf0-4562-b839-98991a889e54

### DIFF
--- a/src/dev-center/js/dev-center.js
+++ b/src/dev-center/js/dev-center.js
@@ -288,6 +288,8 @@ async function create_app(title, source_path = null, items = null) {
     
         })
         .then(async (app) => {
+            $('.new-app-modal').get(0).close();
+            window.location.reload();
             let app_dir;
             // ----------------------------------------------------
             // Create app directory in AppData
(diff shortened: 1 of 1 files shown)


jelveh:
I can't reproduce this issue. Even if I did, the window.reload way of updating the app state is not correct.

mariatouil:
@jelveh These two lines:

**$('.new-app-modal').get(0).close();  
window.location.reload();**  
actually address a different issue that has already been fixed: resolve issue #983 - Stuck on "Creating new app" loading screen.

I had already submitted a pull request for this, and it was merged. However, when cloning the project, I couldn't find these modifications in the codebase, so I added them directly to resolve both problems simultaneously.

Issue #754, opened 2024-09-05
Complete Malayalam translation based on English file


jelveh:
The Malayalam translation file (`ml.js`) needs to be updated to include all translations present in the English file (`en.js`). Currently, the Malayalam translation may be incomplete or outdated compared to the English version.

### Steps to resolve
1. Compare the contents of `src/gui/src/i18n/translations/en.js` and `src/gui/src/i18n/translations/ml.js`
2. For each translation key in `en.js`, ensure a corresponding translation exists in `ml.js`
3. If a translation is missing in `ml.js`, add it with an appropriate Malayalam translation
4. If any translations in `ml.js` are outdated or incorrect, update them to match the current English version

### Additional notes
- Please ensure that the translations are culturally appropriate and maintain the intended meaning of the English text
- If certain terms or phrases don't have a direct Malayalam equivalent, consider adding translator notes or using transliteration where appropriate

### Files to be updated
- `src/gui/src/i18n/translations/ml.js`

### Related files
- `src/gui/src/i18n/translations/en.js` (reference file)

Thank you for helping to improve the Malayalam localization of our project!

rinogodson:
Assign me.

jelveh:
Done 🫡

joswin18:
Hi, I noticed that the Malayalam translation file (`ml.js`) is missing in the directory `src/gui/src/i18n/translations/`. I would like to create this file and add translations for all the keys present in the English file (`en.js`).  

I'll ensure the translations are accurate, culturally appropriate, and aligned with the intended meanings in English. Could you please assign this issue to me?

Thank you!

> The Malayalam translation file (`ml.js`) needs to be updated to include all translations present in the English file (`en.js`). Currently, the Malayalam translation may be incomplete or outdated compared to the English version.
> 
> ### Steps to resolve
> 1. Compare the contents of `src/gui/src/i18n/translations/en.js` and `src/gui/src/i18n/translations/ml.js`
> 2. For each translation key in `en.js`, ensure a corresponding translation exists in `ml.js`
> 3. If a translation is missing in `ml.js`, add it with an appropriate Malayalam translation
> 4. If any translations in `ml.js` are outdated or incorrect, update them to match the current English version
> 
> ### Additional notes
> * Please ensure that the translations are culturally appropriate and maintain the intended meaning of the English text
> * If certain terms or phrases don't have a direct Malayalam equivalent, consider adding translator notes or using transliteration where appropriate
> 
> ### Files to be updated
[...]

PR #1075, opened 2024-12-20
Complete Malayalam translation based on English file #754

joswin18:
- Created the Malayalam translation file (ml.js) by adding all translation keys present in the English file (en.js).
- Translated all keys from English to Malayalam, ensuring that all text is culturally appropriate and contextually accurate.
- Added comments or placeholders where necessary for future adjustments.
- Ensured that the translations are ready for implementation across the application.


jelveh:
Thank you for your contribution. Merged ✌️

PR #1090, opened 2025-01-15
TLS Support through rustls-wasm

ProgrammerIn-wonderland:


ProgrammerIn-wonderland:
Can't believe my eyes but it really is here finally
API use example
```js
const socket = new puter.net.tls.TLSSocket("google.com", 443);
socket.on("data", (data) => {
    console.log(new TextDecoder().decode(data))
})
socket.on("open", (data) => {
    socket.write(new TextEncoder().encode("GET / HTTP/1.1\r\nHost: google.com\r\n\r\n"));
})
socket.on("error", (e) => {
    console.log(e)
})
```

PR #1093, opened 2025-01-17
Complete the Norwegian Bokmål (Norsk Bokmål) translation of Puter

AryanTavish:
Done with the translation and also optimized the billing section 
Issue 857
#857


jelveh:
Thank you for contributing this translation!

One small issue to fix: The billing section is currently structured as an object (`billing = { ... }`), but it needs to be formatted as individual string keys like the English version, where each key follows the pattern `'billing.key_name'`. This is important because using actual nested objects would cause issues with the translation system.

For example, instead of:
```javascript
billing = {
    change_payment_method: "Endre",
    // ...
}
```

It should be:
```javascript
'billing.change_payment_method': "Endre",
// ...
```

Could you update the billing section to match this format? Once that's fixed, the translation will be ready to merge. Thanks again for your contribution!

AryanTavish:
sure thing! , updated the commit :)

Issue #988, opened 2024-12-05
Add "Upload Files" Button in the Deployment Section


HatemLamia:
To enhance user experience and simplify the file upload process, I thought of adding an "Upload Files" button in the deployment section. This feature would allow users to upload their files more intuitively and quickly.

I’ve already started working on it, but I’d appreciate your feedback and suggestions before proceeding. Do you think this feature aligns with our current needs, or are there any aspects that could be improved?

Thanks in advance for your input!

jelveh:
Would this be a button that opens the filedialog/filepicker?

HatemLamia:
Yes, it's a button that opens a selector to simplify downloads.

HatemLamia:
I've already added the button.
<img width="746" alt="image" src="https://github.com/user-attachments/assets/396d0768-b677-4644-b66a-511349d6ae4d">

jelveh:
I just saw this! Github notification makes me miss some comments...
This looks pretty good. Looking forward to the PR.

reynaldichernando:
Revisiting this issue, because I think it's a good idea to have the upload files in Dev Center

Last update about this was in #1098. And also a comment by @KernelDeimos about Puter file picker to open instead of the default browser's.

>  One thing I noticed though is that I expected the Puter file picker to open but instead I got my browser's default file picker. That makes me wonder if there's a good way to provide both options.

I think we can include both, where we have the options for
- Upload file
- Upload from Puter

and it can look something like this 

<img width="240" height="128" alt="Image" src="https://github.com/user-attachments/assets/9832a665-5566-4b4a-b956-6967a1784a90" />

where the upload file from browser is the default one, and the dropdown option for people to upload from puter

the default upload being the first one, assuming people aren't familiar with the puter ecosystem yet

what do you guys think?

reynaldichernando:
<img width="722" height="421" alt="Image" src="https://github.com/user-attachments/assets/fa95048b-1800-474a-9c08-cfb6f6c47366" />

another consideration is this, i checked how upload works for the app logo, it first open the puter file selector
but there is a Upload button at the bottom left

however upon choosing something to upload, this doesn't set it automatically as the selected file, it just upload it to the opened directory

PR #1098, opened 2025-01-27
FEAT:add upload btn #988

HatemLamia:
Context
This change addresses issue https://github.com/HeyPuter/puter/issues/988, which requested a visible button to allow quick access to the file upload functionality.

Changes Made
Added an "Upload Files" button in the deployment section.
Implemented the logic to handle file uploads.
Ensured the button's styling aligns with the existing UI design.

puter

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7800,9 +7800,10 @@
       }
     },
     "node_modules/axios": {
-      "version": "1.7.7",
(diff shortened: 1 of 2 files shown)


KernelDeimos:
I just tested this and it's working fine, not sure why this wasn't merged earlier. One thing I noticed though is that I expected the Puter file picker to open but instead I got my browser's default file picker. That makes me wonder if there's a good way to provide both options.

PR #1100, opened 2025-01-31
Update fr.js

Zac0511:
I fixed 2 translations, and i also deleted a copy of the billing part, as it were in the files 2 times but with different translations, which could create problems, so i kept only one. The second one (the one i deleted) had strange comment and the translations were different from the actual translations, which could imply this copy was made by an AI, but now its fixed.


jelveh:
Thank you @Zac0511! I'm merging this 🫡

PR #1153, opened 2025-03-07
Create devcontainer.json for developers using codespaces gets a GUI to open it in browser because portforwarding simply not working

thisisharshavardhan:


KernelDeimos:
Hi, can you add more explanation to this PR? I don't know what port 6080 is for or how it relates to Puter. Also, I don't use devcontainers so I have no context for this.

thisisharshavardhan:
Hi @KernelDeimos,

Thanks for checking this PR!

The reason for adding a GUI in the devcontainer.json is that when opening Codespaces in a browser, I need to access it via puter.localhost:4100. However, this gives an "Invalid Host" error. By using a GUI inside Codespaces, we can open the website directly within the container without needing to manually deal with host resolution issues.

Let me know if you need more details!

KernelDeimos:
More details would be helpful but I'm not sure what details to ask for since I haven't used codespaces. That way I'm thinking about this right now, it seems like it's a tool that you're using and therefore is a separate concern from our repository - i.e. if this belongs here everybody's config for their own IDEs and tools would belong here too which doesn't make a lot of sense. It's possible that we ***really should*** have supporting config for codespaces in Puter's repo, but if that's the case you'll need to convince me since I'm not familiar with it.

thisisharshavardhan:
Thanks for your feedback! I understand your concern about whether this belongs in the repository. The reason for adding this is to improve the developer experience when using Codespaces with this project.

Since Codespaces does not handle host resolution properly (resulting in the "Invalid Host" error when trying to access puter.localhost:4100), adding a GUI inside the devcontainer ensures that developers can access the interface without manual configuration.

If Codespaces is not a primary focus of this repo, I understand if this isn't a priority. However, if you think support for Codespaces could be useful for contributors, I'd be happy to refine this PR further to align with the repo's goals.

Let me know what you think!

KernelDeimos:
I decided to try Codespaces out. So far my experience was not great, but I can see the value in it if I am able to get it working. Right now I'm stuck on an issue where Codespaces automatically runs `npm install` in node version 16 which builds an incompatible binary for the better-sqlite3 module which seems to cause irreparable issues when I run Puter even in the correct node version (22) after running a subsequent `npm rebuild`.

If you can help me get through these issues and I can get codespaces working so that I can also use it myself, then I will merge this.

thisisharshavardhan:
@KernelDeimos 
i updated the configuration in .devcontainer, it will not execute "npm install" automatically and you will get a port 6080 open, there you will get GUI of codespace VM and you can open puter.localhost:4100 in the browser of codespace 

( the reason for GUI in codespace is when you open codespace in browser and run puter then you wont be able to open puter.localhost:4100, you will get host resolution issues so that you can open it in browser of codespace via GUI of codespace)

you don't need GUI of codespace when you connect to codespace via vs code or any local IDE .( there is no problem when you connect to codespace to local IDE , you can directly open puter.localhost:4100 in your browser ) port forwarding issue only occuring when connected to codespace using browser

let me know if you face any problem

Issue #1149, opened 2025-03-04
Enhance AI Command in Puter's shell


KernelDeimos:
## Phoenix Shell `ai` Command

Puter's shell, [Phoenix](https://github.com/HeyPuter/puter/blob/870e4aa93807f1fcd37af29d9369737ab9f4b4b1/src/phoenix), has a command called [`ai`](https://github.com/HeyPuter/puter/blob/6a184d52b47d80f23babaa94f9ccc32ed6ea14be/src/phoenix/src/puter-shell/coreutils/ai.js). This command takes a string argument and prompts an LLM. The command is able to see the history of your terminal so you can ask questions about output you saw from a command, if you made an error - that sort of thing. It would be really cool if the `ai` command was able to help you write commands or perform other behaviors in Puter. Doing this with proper safeguards to prevent mistakes or damage to user data will be an important and interesting challenge.

ntwari-bruce:
@KernelDeimos could you assign me to this issue

ParinAcharyaGit:
@KernelDeimos could you also assign me to this issue? Thanks

ntwari-bruce:
@KernelDeimos 
I’ve been testing and exploring the ai command in the terminal to better understand how to implement the improvements suggested in this issue. From my testing, I see that the AI already suggests how users can execute certain commands—for example, if a user types: 
          ai "How do I create a new directory here?" - It provides guidance on using mkdir.   

You also mentioned that it would be great if the AI command could help users perform other behaviors in Puter. Based on that, I was thinking of extending its functionality to allow users to execute commands directly from the AI command.

For instance:
          ai "Create a new folder in this directory"

The AI would automatically create the folder, rather than just suggesting the command.

To ensure safety, I’d implement safeguards for sensitive commands—especially those that could delete files or modify important system data. The AI would prompt for confirmation before executing anything that might be harmful.

Would love to hear your thoughts on this approach before I start implementing! Looking forward to contributing.

ntwari-bruce:
@KernelDeimos , I’m having trouble running commands in the terminal on localhost. The ai command is not returning any response, and other basic commands are also not working. When I run the help command, it lists these commands as available, but they don’t execute.

Could you provide any guidance on troubleshooting this? Regarding co-assigning me with @ParinAcharyaGit , I have no problem collaborating with him, but for now i need to fix few issues before we can start together.

ntwari-bruce:
@KernelDeimos you can co-assign @ParinAcharyaGit

[...]

PR #1156, opened 2025-03-07
Enhanced ai command to perfom other commands

ntwari-bruce:
This pull request enhances the ai command to recognize and execute system commands (e.g., mkdir, ls, etc.) directly from a prompt. Previously, the AI command could only respond with text, but now it can detect embedded commands and execute them seamlessly.

Changes Made
	•	Introduced logic to parse and extract system commands from the AI response using the pattern %%%command%%%.
	•	Added functionality to execute the extracted command within the shell environment.
	•	Moved the await ctx.externs.out.write(message + '\n'); statement into the else block to ensure regular messages are printed only when no command is detected. Now, ai command has enhanced capabilities where users can use it to interact with the terminal. This uses "ctx.shell.runPipeline" to run commands from inside a command.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "experiments/js-parse-and-output"
       ],
       "dependencies": {
+        "@heyputer/putility": "^1.0.2",
         "dedent": "^1.5.3",
         "javascript-time-ago": "^2.5.11",
         "json-colorizer": "^3.0.1",
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     ]
   },
   "dependencies": {
+    "@heyputer/putility": "^1.0.2",
     "dedent": "^1.5.3",
     "javascript-time-ago": "^2.5.11",
     "json-colorizer": "^3.0.1",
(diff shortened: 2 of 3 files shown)


KernelDeimos:
Great work! Before I can merge this to `main` it's very important that the user is asked for confirmation before running an AI-generated command. Our release cycles are very fast so this will go into production soon after being merged.

This is mentioned [in the issue](https://github.com/HeyPuter/puter/issues/1149):
> Doing this with proper safeguards to prevent mistakes or damage to user data will be an important and interesting challenge.

Based on the source from the `cat` command, this should work to get a line of input from the user:
```javascript
({ value: line } = await ctx.externs.in_.read())
if ( ! (line ?? '').toLowerCase().startsWith('y') ) {
  // Display error and return
}
// Run command
```

I haven't tested the code snippet above, but that's the general idea for capturing a Y/n prompt from the user.

ntwari-bruce:
I have updated the changes; Now, users will be able to confirm the command first before running it.

KernelDeimos:
Perfect, I no review comments so I'm going to merge this now. I'll open a new issue for using tool use which would potentially allow the ai command to take multiple steps for complex tasks.

KernelDeimos:
Added [the subsequent issue](https://github.com/HeyPuter/puter/issues/1163). If you'd like you can take this too since you're already familiar with this command, or better yet you could help someone else through it.

ntwari-bruce:
@KernelDeimos yes i would like to take it

Issue #828, opened 2024-11-05
Add an API to list all available AI models


jelveh:

chetan21122004:
hi would you provide more details , i would like to contribute

cprogrammer1209:
Hi @jelveh,
Can you explain it more, I would like to take this issue and contribute.

PR #1160, opened 2025-03-08
add API to get a list of all available models

arunabhcode:
Addressing https://github.com/HeyPuter/puter/issues/828

It adds two calls listModels() and listProviders().

### DIFF
--- a/src/puter-js/src/modules/AI.js
+++ b/src/puter-js/src/modules/AI.js
@@ -37,6 +37,60 @@ class AI{
         this.APIOrigin = APIOrigin;
     }
 
+     /**
+     * Returns a list of available AI models.
+     * @param {string} provider - The provider to filter the models by.
+     * @returns {Object} Object containing lists of available models by provider
+     */
+     async listModels(provider) {
+        const modelsByProvider = {};
+
+        const modelsList = await puter.drivers.call('puter-chat-completion','ai-chat','models');
+
+        if (modelsList && modelsList.result && Array.isArray(modelsList.result)) {
+            modelsList.result.forEach((item) => {
+                if (item.provider && item.id) {
+                    if (provider) {
+                        if (item.provider === provider) { // Only add models for the specified provider
+                            if (!modelsByProvider[item.provider]) {
+                                modelsByProvider[item.provider] = [];
+                            }
+                            modelsByProvider[item.provider].push(item.id);
+                        }
+                    } else { // Add all models
+                        if (!modelsByProvider[item.provider]) {
+                            modelsByProvider[item.provider] = [];
+                        }
+                        modelsByProvider[item.provider].push(item.id);
+                    }
+                }
+            });
+        }
+ 
+        return modelsByProvider;
+    }
+
+    /**
+     * Returns a list of all available AI providers
+     * @returns {Array} Array containing providers
+     */
+    async listModelProviders() {
+        let providers = [];
+        const models = await puter.drivers.call('puter-chat-completion','ai-chat','models');
+
+        if (models && models.result && Array.isArray(models.result)) {
+            providers = new Set(); // Use a Set to store unique providers
+
+            models.result.forEach((item) => {
+            if (item.provider) {
+                providers.add(item.provider);
+            }
+            });
+            providers = Array.from(providers); // Convert Set to an array
+        }
+        return providers;
+    }
+    
     img2txt = async (...args) => {
         let MAX_INPUT_SIZE = 10 * 1024 * 1024;
         let options = {};
--- a/src/puter-js/src/modules/AI.js
+++ b/src/puter-js/src/modules/AI.js
@@ -46,7 +46,6 @@ class AI{
         const modelsByProvider = {};
 
         const models = await puter.drivers.call('puter-chat-completion','ai-chat','models');
-        //console.log(models);
 
         if (!models || !models.result || !Array.isArray(models.result)) {
             return modelsByProvider;
@@ -58,7 +57,6 @@ class AI{
             modelsByProvider[item.provider].push(item.id);
         });
         
-        //console.log(modelsByProvider);
         return modelsByProvider;
     }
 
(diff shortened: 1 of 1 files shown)


arunabhcode:
```
Sample output: 
Available models: {
  "chat": {
    "openai": [
      "gpt-4o",
      "gpt-4o-mini"
    ],
    "anthropic": [
      "claude-3-haiku-20240307",
      "claude-3-5-sonnet-20240620",
      "claude-3-5-sonnet-20241022",
      "claude-3-5-sonnet-latest",
      "claude-3-7-sonnet-latest"
    ],
    "together": [
      "meta-llama/Meta-Llama-3.1-8B-Instruct-Turbo",
      "meta-llama/Meta-Llama-3.1-70B-Instruct-Turbo",
      "meta-llama/Meta-Llama-3.1-405B-Instruct-Turbo",
      "google/gemma-2-27b-it"
    ],
    "mistral": [
      "mistral-large-latest",
      "codestral-latest"
    ],
    "groq": [
      "distil-whisper-large-v3-en",
      "gemma2-9b-it",
      "gemma-7b-it",
      "llama-3.1-70b-versatile",
      "llama-3.1-8b-instant",
      "llama3-70b-8192",
      "llama3-8b-8192",
      "llama3-groq-70b-8192-tool-use-preview",
      "llama3-groq-8b-8192-tool-use-preview",
      "llama-guard-3-8b",
      "mixtral-8x7b-32768",
      "whisper-large-v3"
    ],
    "xai": [
      "grok-beta"
    ],
    "deepseek": [
      "deepseek-chat",
      "deepseek-reasoner"
    ],
    "gemini": [
      "gemini-1.5-flash",
      "gemini-2.0-flash"
    ]
  }
}
Available providers: {
  "chat": [
    "openai",
    "anthropic",
    "together",
    "mistral",
    "groq",
    "xai",
    "deepseek",
    "gemini"
  ]
}
```

KernelDeimos:
Hi, thanks for the contribution. This unfortunately doesn't address the issue properly - that's maybe on us because the issue you linked to doesn't have any description. I'll add further detail on monday. This should be requesting model information from the backend services in the [AI module](https://github.com/HeyPuter/puter/tree/main/src/backend/src/modules/puterai). I'll try to remember to add more information about this by the end of Monday.

arunabhcode:
@KernelDeimos Sounds good, waiting for further clarification will try to work on it in the meantime and guess around what's needed.

KernelDeimos:
Hi again, this request to backend gets all supported models

```javascript
await puter.drivers.call('puter-chat-completion', 'ai-chat', 'models');
```

This endpoint doesn't split it into "models" and "providers", but it does provide enough information for you to derive that.

arunabhcode:
@KernelDeimos so you want to retrieve the dict returned by calling this and then process that dict to create the dict above instead of hardcoding?

KernelDeimos:
Yes that's correct

arunabhcode:
@avijh squashed our commits into one since this is a small PR.

src/puter-js/src/modules/AI.js
```diff
@@ -37,6 +37,60 @@ class AI{
...
+     */
+    async listModelProviders() {
+        let providers = [];
+        const models = await puter.drivers.call('puter-chat-completion','ai-chat','models');
+
+        if (models && models.result && Array.isArray(models.result)) {
+            providers = new Set(); // Use a Set to store unique providers
+
+            models.result.forEach((item) => {
+            if (item.provider) {
```
KernelDeimos:
```suggestion
                if (item.provider) {
```

src/puter-js/src/modules/AI.js
```diff
@@ -37,6 +37,60 @@ class AI{
...
+                providers.add(item.provider);
+            }
```
KernelDeimos: I think this is supposed to be intended one more level

```suggestion
                if (item.provider) {
                    providers.add(item.provider);
                }
```

src/puter-js/src/modules/AI.js
```diff
@@ -37,6 +37,60 @@ class AI{
...
+        const modelsByProvider = {};
+
+        const modelsList = await puter.drivers.call('puter-chat-completion','ai-chat','models');
+
+        if (modelsList && modelsList.result && Array.isArray(modelsList.result)) {
+            modelsList.result.forEach((item) => {
+                if (item.provider && item.id) {
+                    if (provider) {
+                        if (item.provider === provider) { // Only add models for the specified provider
+                            if (!modelsByProvider[item.provider]) {
```
KernelDeimos: Whenever you see `if if if if`, you can be sure there's a simpler approach. This amount of nesting increases a metric in static code analysis known as "cognitive complexity" which is a measure of how difficult the code is for somebody to understand and maintain later.

We can use [guard clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) (also I found [this video](https://www.youtube.com/watch?v=qgStum-3Y7Q) which is a bit long but you can get the jist of it by just scanning through real quick and looking at how he's changing the code) to simplify (untested):
```javascript
...forEach(item => {
    if ( ! item.id ) return;
    if ( provider && ( item.provider !== provider ) ) return;
    if ( ! modelsByProvider[item.provider] ) {
      modelsByProvider[item.provider] = [];
    }
    modelsByProvider[item.provider].push(item.id);
}
```

Technically we could even combine the first two `if` statements, but I think it's easier to read when separate concerns still belong on separate lines.

KernelDeimos:
Manual test passed. I did manual testing early since I know testing this without API keys is difficult. I'll do a quick test again after my review comments are addressed, and then this will be ready to merge.

https://github.com/user-attachments/assets/7930b116-a5c5-4341-9fc1-fe092570ee6b

src/puter-js/src/modules/AI.js
avijh: Agreed. I modified the code per your suggestions and also removed redundant {} for readability.

KernelDeimos:
Nice work! Thanks for addressing my comments

PR #1164, opened 2025-03-10
Update events docs

ajaykrishnaswamy:
This PR includes:
     1. Updated events documentation in doc/contributors/extensions/events.json.js with the latest events from the codebase
     2. Updated doc_helper.js script for generating event documentation


doc_helper.js
KernelDeimos, line 1: Ideally this should be in the `tools` directory, that's where we keep scripts like this.

doc_helper.js
```diff
@@ -0,0 +1,304 @@
...
+    properties.forEach(prop => {
+        propertyDetails[prop] = {
+            type: guessType(prop),
+            mutability: hasAllowEffect ? 'effect' : 'no-effect',
+            summary: guessSummary(prop, eventName)
+        };
+    });
+}
+
+function generateDescription(eventName) {
```
KernelDeimos: I really like this approach. We need a way to override these in case the generated description is wrong. (also consider this for other generated text)

package.json
```diff
@@ -47,7 +47,8 @@
     ]
   },
   "dependencies": {
-    "@heyputer/putility": "^1.0.2",
+    "bcrypt": "^5.1.1",
+    "better-sqlite3": "^8.0.0",
```
KernelDeimos: We have these dependencies in the `package.json` in src/backend, which is where they are used. Is there a reason for elevating the scope of these dependencies?

doc_helper.js
ajaykrishnaswamy: Got it, I will move the file.

package.json
ajaykrishnaswamy: Hello Eric, I didn't realize that there was already a package.json file in the src/backend. I assumed I had deleted the file somehow when I cloned the repo so I regenerated it in the main directory. I will remove the duplicate package.json file in the next PR.

KernelDeimos: You can update this PR by adding this suggestion

```suggestion
```

PR #1165, opened 2025-03-11
Initial Setup Wizard

wingding12:
**Tasks:**
- Configure subdomain behavior
- Configure domain name, or nip.io mode
- Set a password for the admin user

**Steps to use the wizard (can be found in README):**
1. Run `npm start` after cloning puter files
2. Configure settings and create an admin password (default is 9668fafe)
3. Click complete setup
4. KILL the puter build, then rerun it. The default user is admin and the password should be updated to the setup wizard.

**To return to the wizard for different configuration settings:**
1. In the /puter directory, run the following commands:
    - rm ./volatile/runtime/config/setup-completed
    - rm ./volatile/runtime/config/wizard-config.json
    and confirm removal. You may also navigate into the appropriate directory and directly delete the files.
    
2. Run `npm start`. The setup wizard should reappear with default configurations.


KernelDeimos:
Hello, thanks for your contribution! I really like the UI. I noticed a couple of issues with UX and functionality:
- `nip.io` setting doesn't seem to take effect for me; I don't see the changes in my `config.json`
- since the user is providing a new password anyway, I don't think we should replace the randomly generated default password with a hardcoded one

There are also numerous whitespace/style changes. If you can remove them that will make it easier for me to do a deeper review.

psg1986 (approved):

KernelDeimos:
Please ignore the review by `psg1986`. I don't think that's a real account.

wingding12:
Hi, could you please test the setup wizard and let me know what issues you run into? Locally, it seems to be working fine for myself.

KernelDeimos:
I did a manual test just now but I think I did it too early. Let me know when this PR is ready for me to test again. In case it's helpful though, here's a recording of my test anyway.

https://github.com/user-attachments/assets/c0afeb16-f364-4f48-a080-75e6521280b5

wingding12:
I’ve gotten to a point where changes are being made directly to the correct config.json file, right now the only thing I have left so far to change is reverting the nip.io domain value if the user decided to go back to the setup wizard.

I’ve also reduced some files so that the user only needs to delete ./volatile/runtime/config/setup-completed

Could you please test out whether or not my implementation is correct for IP-based configuration? At the moment I’ve just left the nip.io domain to use the local port, just figuring out an automated way to change config.json to go back to localhost, but currently just changing the file directly works as intended

KernelDeimos:
Hi, I got your message on Slack earlier. I've done another manual test. There seem to be some additional issues.
- `admin` user is no longer created
- the password set for the admin user is stored in a plaintext file. I'm not sure what this is for, but we can't store passwords in plaintext. User passwords do not need to be stored anywhere other than in the database where they are stored already
- when using nip.io, Puter doesn't work anymore. If I checkout the `main` branch it seems to be working.

Here's a recording of my manual text (1.5x speed for convenience)

https://github.com/user-attachments/assets/bc5db820-7822-4e98-9cd6-6ece2cffef6c

PR #1167, opened 2025-03-11
Added captcha verification system for login and signup

Jomaguy:
For my Captcha verification I added the following components

Captcha Module:
The Captcha Module serves as the integration point for the captcha verification system within the Puter platform. It registers the captcha service and middleware with the application, making them available throughout the system. This module enables protection against automated abuse for sensitive operations like login and signup.

Captcha Service:
The Captcha Service handles the core functionality of generating and verifying captchas. It creates visual challenges using SVG-based captchas, manages token generation and storage, and validates user responses. The service supports configurable difficulty levels and token expiration times, and uses an in-memory token storage system with automatic cleanup of expired tokens.

Captcha Middleware:
The Express middleware component provides a simple way to protect routes with captcha verification. It can be configured to either always require captcha verification or conditionally require it based on user behavior (such as failed login attempts). The middleware seamlessly integrates with existing routes, checking for valid captcha tokens and answers before allowing requests to proceed.


src/backend/src/modules/captcha/CaptchaModule.js
```diff
@@ -0,0 +1,76 @@
...
+ * that can be used to protect routes.
+ */
+class CaptchaModule extends AdvancedBase {
+    async install(context) {
+        const services = context.get('services');
+        
+        const { CaptchaService } = require('./services/CaptchaService');
+        
+        // Get configuration from environment or use defaults
+        const captchaEnabled = process.env.CAPTCHA_ENABLED !== 'false'; // Enabled by default
```
KernelDeimos: This should probably come from configuration instead.
```javascript
const config = require("../../config");
```

src/backend/src/modules/captcha/CaptchaModule.js
```diff
@@ -0,0 +1,76 @@
...
+                    }
+                });
+            }
+        } catch (error) {
+            console.error('Error accessing captcha service after registration:', error);
+        }
+        
+        // Register captcha middleware
+        const captchaMiddleware = require('./middleware/captcha-middleware');
+        context.set('captcha-middleware', captchaMiddleware);
```
KernelDeimos: Is `captcha-middleware` accessed via context anywhere?

src/backend/src/modules/captcha/middleware/captcha-middleware.js
```diff
@@ -0,0 +1,170 @@
...
+        const services = Context.get('services');
+        
+        // Fail closed if services aren't available (in strict mode)
+        if (!services) {
+            console.warn('Captcha middleware: services not available in Context');
+            if (strictMode) {
+                // Use next(error) instead of throwing directly
+                return next(APIError.create('internal_error', null, {
+                    message: 'Captcha service unavailable',
+                    status: 500
```
KernelDeimos: 503 might be more appropriate
```suggestion
                    status: 503
```

src/backend/src/modules/captcha/middleware/captcha-middleware.js
```diff
@@ -0,0 +1,170 @@
...
+        
+        let captchaService;
+        try {
+            captchaService = services.get('captcha');
+        } catch (error) {
+            console.warn('Captcha middleware: captcha service not available', error);
+            if (strictMode) {
+                return next(APIError.create('internal_error', null, {
+                    message: 'Captcha service unavailable',
+                    status: 500
```
KernelDeimos:
```suggestion
                    status: 503
```

.lsp/.cache/db.transit.json
KernelDeimos, line 1: should this file be here?

.clj-kondo/.cache/v1/lock
KernelDeimos, line 1: should this file be here?

KernelDeimos (commented):
Thanks for your contribution! This looks pretty good so far. I noticed a possible area of improvement: it would be nice if we could conditionally display the captcha for login/signup using logic in a backend extension. I noticed you already implemented the behavior to check a flag on the requester object, the only missing piece  then is to emit an event for extensions and update login/signup to be conditional.

Here's an event from CleanEmailService that lets extensions validate email addresses:
```javascript
const svc_event = this.services.get('event');
const event = { allow: true, email };
await svc_event.emit('email.validate', event);

if ( ! event.allow ) return false;
```

src/backend/src/modules/captcha/CaptchaModule.js
Jomaguy: You're right, it's not. I removed the content integration

KernelDeimos:
I just tried the latest changes. It seems to work great on the backend side of things. On the frontend, because you made the change I requested for conditional captcha requirement via extensions, we no longer have a good way to determine when the captcha element doesn't need to be displayed.

I think the best way to resolve this is to split the captcha middleware into two middlewares:
1. checkCaptcha - "first half": finds out if authentication actions will require a captcha. PuterHomepageService can use this to set a gui param, or maybe we set it in the [the new `/whoarewe` endpoint](https://github.com/HeyPuter/puter/commit/ad7b7b951345bb23c9010c8e3509673412c5d6d3).
2. requireCaptcha - "second half": depends on result form checkCaptcha, does actual captcha check and error responses.

Jomaguy:
I think I closed this by mistake

KernelDeimos:
That's okay, I assume #1174 contains these changes plus the new ones correct? In the future you should be able to re-open closed pull-requests

PR #1168, opened 2025-03-12
Update README.md

Gautam7352:
Without clear instructions it wasted a few hours just for setting up. 
With this update Atleast no one will need to guess with domain name in localhost now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,9 @@ git clone https://github.com/HeyPuter/puter
 cd puter
 npm install
 npm start
+This will launch Puter at http://puter.localhost:4100 (or the next available port).
 ```
 
-This will launch Puter at http://puter.localhost:4100 (or the next available port).
 
 If this does not work, see [First Run Issues](./doc/self-hosters/first-run-issues.md) for
 troubleshooting steps.
@@ -59,6 +59,7 @@ troubleshooting steps.
 
 ```bash
 mkdir puter && cd puter && mkdir -p puter/config puter/data && sudo chown -R 1000:1000 puter && docker run --rm -p 4100:4100 -v `pwd`/puter/config:/etc/puter -v `pwd`/puter/data:/var/puter  ghcr.io/heyputer/puter
+This will should launch Puter at http://puter.localhost:4100 (or the next available port).
 ```
 
 <br/>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,9 +46,9 @@ git clone https://github.com/HeyPuter/puter
 cd puter
 npm install
 npm start
+This will launch Puter at http://puter.localhost:4100 (or the next available port).
```
KernelDeimos: why is this moved into the code block?

Gautam7352: Not just that, you might notice that I have added it to all the installation methods I have added complete link where it's hosted.

Using subdomains is not a general practice that's followed, which is causing multiple fresh installers to face "invalid host header" issue.
For me it wasted 4 to 5 hours and I had to go through internet and issues section to figure it out, 
why anyone needs to go through that if it can be mitigated by modifying readme?


PS: For me docker method worked so far, but npm install is not installing all dependencies smoothly. Maybe there are some pre-requisites or maybe some other issue, will figure it out and make required PR as needed.

Gautam7352: Not just that, you might notice that in all running approaches I have added complete link where it's hosted. 

Increases relative Visibility

Using subdomains is not a general practice that's followed, which is causing multiple fresh installers to face "invalid host header" issue.

For me it wasted 4 to 5 hours and I had to go through internet and issues section to figure it out, 
why anyone needs to go through that if it can be mitigated by modifying readme?


PS: For me docker method worked so far but made the change presuming for all installation methods Host header is same.
"npm install" is not installing all dependencies smoothly. Maybe there are some pre-requisites or maybe some other issue, will figure it out and make required PR as needed.

KernelDeimos: I appreciate the readme changes and I agree that adding this information in the docker instructions is also helpful. I too have contributed documentation changes to other projects after I spent hours troubleshooting, usually only to be met with dismissive or even hostile responses. What I don't understand though is the rationale for moving this text into a code block, it should be either a comment or it should stay outside of the code block. Maybe we can use a bold heading or a blockquote to bring attention to it - I'd prefer it stays outside of the code block because even with a comment marker zsh doesn't like comments in the interactive prompt (which I think was an odd decision on zsh's part, it surprises me every time it says `#` isn't a command)

```suggestion
# This will launch Puter at http://puter.localhost:4100 (or the next available port).
```

KernelDeimos:
Thanks! This looks good to me; I'm merging now.

Issue #535, opened 2024-07-06
Complete Farsi translation based on English file


jelveh:
The Farsi translation file (`fa.js`) needs to be updated to include all translations present in the English file (`en.js`). Currently, the Farsi translation may be incomplete or outdated compared to the English version.

### Steps to resolve
1. Compare the contents of `src/i18n/translations/en.js` and `src/i18n/translations/fa.js`
2. For each translation key in `en.js`, ensure a corresponding translation exists in `fa.js`
3. If a translation is missing in `fa.js`, add it with an appropriate Farsi translation
4. If any translations in `fa.js` are outdated or incorrect, update them to match the current English version

### Additional notes
- Please ensure that the translations are culturally appropriate and maintain the intended meaning of the English text
- If certain terms or phrases don't have a direct Farsi equivalent, consider adding translator notes or using transliteration where appropriate

### Files to be updated
- `src/i18n/translations/fa.js`

### Related files
- `src/i18n/translations/en.js` (reference file)

Thank you for helping to improve the Farsi localization of our project!

KernelDeimos:
Assigned! Thanks for taking on this task

jelveh:
Hello @atharva-patil-23, this issue was mistakenly left open. Sorry about that! 

Let me know if you want to work on other open issues.

PR #1169, opened 2025-03-12
tweak: reorder and fix Farsi translation of i18n keys

shushannap:
(Complete Farsi translation based on English file  #535)

- Analyze en.js and fa.js file to find discrepencies
- Reorder fa.js to correctly match en.js in alphabetical order
- Add comments to break up 2FA Setup, 2FA Login, Sharing, and Billing
- Delete comments of English translations in fa.js
- Kept translation notes from previous translator 
- Ensure Farsi translations are close as possible to original source matter. All keys were already filled out. Upon review, no translation errors were found.

<br> </br>

Contributor Questions:
- Would you prefer I add back in the English translations into fa.js?
- If yes, would you like me to add them to every single key instead of the few that had them prior?

### DIFF
--- a/src/gui/src/i18n/translations/fa.js
+++ b/src/gui/src/i18n/translations/fa.js
@@ -17,6 +17,7 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+
 const fa = {
   name: "فارسی",
   english_name: "Farsi",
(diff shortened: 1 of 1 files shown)


KernelDeimos:
Thanks for your contribution! @jelveh will review this soon

PR #1170, opened 2025-03-12
fix: made protocol dynamic in the publishURL displayed on 'Publish As Website' dialogbox

### LINKED ISSUE #1157: Bug: when running locally, "Publish as Website" dialog has HTTPS hard-coded
### Issue Description

Bug: when running locally, "Publish as Website" dialog has HTTPS hard-coded

### Steps to reproduce

_No response_

### Expected behaviour

_No response_

### Addition Information or Screenshots (if applicable)

_No response_

### Deployment

- [ ] Production (puter.com)
- [ ] Development (`npm run start`)
- [ ] Docker (via `docker run`)
- [ ] Docker (via `docker-compose`)

### Puter version

_No response_

sashpawar11:
This fix resolves #1157 

**Updates**
- Created a helper function `window.extractProtocol()` in `helper.js` to get the accurate protocol from the window URL, as this URL has the required data per configuration via  `static_hotsing_domain` and `protocol`
- The helper function is then called in-place of the hard-coded protocol **https** in the `protocol` part of the publish URL that is displayed on the 'Publish As Website' dialogbox.

The above approach hence resolves the issue, where-in when testing puter on localhost, the `protocol` is always displayed as `https://` in the publish URL, since it is hardcoded in the codebase.

After the new change:
image

### DIFF
--- a/src/gui/src/UI/UIWindowPublishWebsite.js
+++ b/src/gui/src/UI/UIWindowPublishWebsite.js
@@ -37,7 +37,7 @@ async function UIWindowPublishWebsite(target_dir_uid, target_dir_name, target_di
             // subdomain
             h += `<div style="overflow: hidden;">`;
                 h += `<label style="margin-bottom: 10px;">${i18n('pick_name_for_website')}</label>`;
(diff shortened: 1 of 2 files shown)


KernelDeimos:
Thanks!

Issue #1171, opened 2025-03-13
Move current datetime stamp to the toolbar as a widget for a polished-OS look


sashpawar11:
Currently the date and time are displayed as a basic datetime stamp at the bottom right corner of the putter desktop.

We can enhance its visibility by moving it to the top-toolbar, preferably on the top-left corner of the toolbar in its own widget, for ease of access and a more polished look. Let me know any further ideas/thoughts on this!

**Current State:**

![Image](https://github.com/user-attachments/assets/c01f67c3-08f9-4301-8c9d-98d59a2e9ed6)

**DateTime Widget:**

<img width="518" alt="Image" src="https://github.com/user-attachments/assets/85c68be5-9c87-4f5d-8afb-3c352b23a298" />

sashpawar11:
Hey @KernelDeimos - please review. If feasible, I would love to work on it!
Thanks

KernelDeimos:
I like this idea. We should loop @jelveh into this discussion too. We need to think about this carefully because whatever we do here will set a precedent for the location of similar widgets. To me this makes sense though - that's the same place where the clock is on my phone.

KernelDeimos:
btw if you already wrote the code for this you should put up a PR for it. That might lend some inspiration to further discussion.

sashpawar11:
Hey @KernelDeimos - thanks for the insight. No - I haven't implemented it yet, just poured my ideas into this one hehe! Will get a PR for the same. 

And yes, I love the idea of 'Widgets' and the API's for the background apps outlined by you, it would be one of the nice-to-haves for sure!

Thankyou

olii-dev:
Hey! Can I be assigned this? I would love to try it.

KernelDeimos:
@olii-dev this is being worked on by the author of this issue already, but if you want a similar UI task I'm sure I can find one

KernelDeimos:
@olii-dev what do you think of adding something like this after we merge #1172 ?

https://x.com/rauchg/status/1899526475843530985

You should work closely with @sashpawar11 if you decide to take it on, since your implementation should re-use concepts from 1172

olii-dev:
@KernelDeimos I will do my best with @sashpawar11 to do this!

@sashpawar11 how do we connect?

sashpawar11:
Hey @olii-dev  - we can connect on discord if you're on it. My discord username # - sash_2001

olii-dev:
[...]

PR #1172, opened 2025-03-13
dev: moved current datetime to toolbar as a widget

sashpawar11:
This PR creates the initial implementation for #1171.

**ChangeLog**
- Current clock element has been moved to the toolbar under `UIDesktop` with other toolbar buttons
- `display_ct()` under `UIDesktop`  has been tweaked to display a better/sleek datetimestamp
- `window.change_clock_visible` intialization on DOM load has been moved to `UIDesktop` due to DOM element accessibility bug
- interval for `display_ct()` has been changed to `60 * 1000 ms` as seconds are no longer displayed in the clock, hence there is no need to update the element every 1000ms (1s).

**New Toolbar Clock UI:**
image

**Puter Desktop View:**
image


sashpawar11:
@KernelDeimos @jelveh - here's the initial PR for the proposed enhancement under #1157

KernelDeimos:
Thanks, this implementation is very straightforward. No comments, I'm merging it now.

KernelDeimos:
This PR introduced unexpected movement of other icons on the upper bar when the clock is visible. It was a simple fix so I fixed that on main. Here's the commit if you'd like to review it: https://github.com/HeyPuter/puter/commit/8edcf7575d242e30b6c88eb1e9f91c3017835f6c

PR #1175, opened 2025-03-14
UIAlert with dynamic icon types and default button configurations

### LINKED ISSUE #1107: Add multiple icons for puter.ui.alert()
Currently, the `puter.ui.alert()` dialog always show a "warning" icon : 

Image

It would be useful if a new parameter could be added for having 3 additionnal icons, they could be : 
* An "info" icon : useful when showing informations
* An "?" icon : useful when the app asks something
* An "X" error icon : useful when a critical error happens in app
* And the already existing "warning" icon for warnings

Windows and macOS already use those 4 options : 

Image

purohitamann:
### Summary
This PR fixes the issue with alert icons not displaying correctly.

### Changes
- Updated icon for 
- default configurations for alerts 
- Added default fallbacks for missing icons

### Related Issue
Fixes #1107

Demo
Screenshot 2025-03-14 at 1 38 33 PM
Screenshot 2025-03-14 at 1 37 15 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,6 +44,198 @@
         "webpack-cli": "^5.1.1"
       }
     },
+    "asdf/backend": {
+      "version": "2.1.0",
+      "extraneous": true,
+      "license": "AGPL-3.0-only",
+      "dependencies": {
+        "@heyputer/kv.js": "^0.1.3",
+        "@heyputer/multest": "^0.0.2",
+        "@heyputer/putility": "^1.0.0",
+        "@opentelemetry/api": "^1.4.1",
+        "@opentelemetry/auto-instrumentations-node": "^0.43.0",
+        "@opentelemetry/exporter-trace-otlp-grpc": "^0.40.0",
+        "@opentelemetry/sdk-metrics": "^1.14.0",
+        "@opentelemetry/sdk-node": "^0.49.1",
+        "@pagerduty/pdjs": "^2.2.4",
+        "@smithy/node-http-handler": "^2.2.2",
+        "args": "^5.0.3",
+        "aws-sdk": "^2.1383.0",
+        "axios": "^1.4.0",
+        "bcrypt": "^5.1.0",
+        "better-sqlite3": "^9.4.3",
+        "busboy": "^1.6.0",
+        "chai-as-promised": "^7.1.1",
+        "clean-css": "^5.3.2",
+        "composite-error": "^1.0.2",
+        "compression": "^1.7.4",
+        "cookie-parser": "^1.4.6",
+        "dedent": "^1.5.3",
+        "express": "^4.18.2",
+        "file-type": "^18.5.0",
+        "form-data": "^4.0.0",
+        "handlebars": "^4.7.8",
+        "helmet": "^7.0.0",
+        "hi-base32": "^0.5.1",
+        "html-entities": "^2.3.3",
+        "is-glob": "^4.0.3",
+        "isbot": "^3.7.1",
+        "jimp": "^0.22.8",
+        "js-sha256": "^0.9.0",
+        "json5": "^2.2.3",
+        "jsonwebtoken": "^9.0.0",
+        "knex": "^3.1.0",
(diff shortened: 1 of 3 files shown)


purohitamann:
@jelveh please take a look:D

src/gui/src/UI/UIAlert.js
```diff
@@ -19,31 +19,72 @@
...
+            info: { icon: "reminder.svg", title: "Info", color: "#1976D2" },
+            success: { icon: "c-check.svg", title: "Success!", color: "#388E3C" },
+            confirm: { icon: "question.svg", title: "Are you sure?", color: "#555555" }
+        };
 
-        // set body icon
-        options.body_icon = options.body_icon ?? window.icons['warning-sign.svg'];
-        if(options.type === 'success')
+        // Set default values
+        const alertType = alertTypes[options.iconType] || alertTypes.warning;
```
KernelDeimos: is `iconType` the best name for this option? We're storing it as `alertType` which suggests this is a general theme for the dialog (and indeed it looks like we set a value for `color` based on this too). I think the option should be simply `type`.

src/gui/src/UI/UIAlert.js
```diff
@@ -19,31 +19,72 @@
...
+        options.message = options.title || alertType.title;
+        options.body_icon = options.body_icon ?? window.icons[alertType.icon];
+        options.color = alertType.color;
```
KernelDeimos: Maybe we should allow an override here.

```suggestion
        options.color = options.color ?? alertType.color;
```

src/gui/src/UI/UIAlert.js
```diff
@@ -19,31 +19,72 @@
...
+                    ];
+                    break;
+                case "error":
+                    options.buttons = [
+                        { label: "Retry", value: "retry", type: "danger" },
+                        { label: "Cancel", value: "cancel", type: "secondary" }
+                    ];
+                    break;
+                default:
+                    options.buttons = [{ label: "OK", value: true, type: "primary" }];
```
KernelDeimos:
```suggestion
                    options.buttons = [{ label: i18n('ok'), value: true, type: "primary" }];
```

package-lock.json
KernelDeimos, line 1: Since this PR does not change packages, `package-lock.json` should not be updated here. If it's out of sync, then it does need to be changed, but it should be changed separately.

KernelDeimos (commented):
This looks good, I just have a few minor comments. This will be a nice update, thanks!

src/gui/src/UI/UIAlert.js
purohitamann: Fair enough! It's gonna be `type`

package-lock.json
purohitamann: so I will just do a reset for this file then

purohitamann (commented):
@KernelDeimos , what do you think?

KernelDeimos:
Looks good, thanks!

src/gui/src/UI/UIAlert.js
```diff
@@ -19,31 +19,72 @@
...
+            success: { icon: "c-check.svg", title: "Success!", color: "#388E3C" },
+            confirm: { icon: "question.svg", title: "Are you sure?", color: "#555555" }
+        };
 
-        // set body icon
-        options.body_icon = options.body_icon ?? window.icons['warning-sign.svg'];
-        if(options.type === 'success')
+        // Set default values
+        const alertType = alertTypes[options.type] || alertTypes.warning;
+        options.message = options.title || alertType.title;
```
KernelDeimos, line 54: This line broke alerts but wasn't caught during testing because I only tested the alert types. Alert messages don't work at all because of this - were messages tested?

Suresh-cs-q:
`if(options.type === 'success') {
        // Set default values
        const alertType = alertTypes[options.type] || alertTypes.warning;
        options.message = options.title || alertType.title;
}`


The original code missed braces ({}), causing only the const alertType line to run conditionally, while options.message ran unconditionally (and likely threw errors for non-success types).

PR #1177, opened 2025-03-14
GUI Permission Dialog

### LINKED ISSUE #1150: GUI Permission Dialog
## Permission Dialog

This task is to build a dialog in Puter's GUI that allows an app to request permissions. We had one before but it wasn't working properly.

There are two possible scopes for this work:
1. only implement for inside Puter's desktop
2. implement for apps in Puter as well as websites using puter.js

Option 2 is more difficult. Apps in Puter's desktop are easier because we can communicate with Puter's desktop bidirectionally via postMessage and have it handle the UI for permission requests. Websites, on the other hand, need to ask the browser to open a new tab or popup window for requesting the permissions. Because of [cross-origin isolation](https://web.dev/articles/cross-origin-isolation-guide), this becomes very difficult and in some cases impossible.

**I highly recommend sticking with Option 1 to start.**, but keep Option 2 in mind as potential future work that the architecture of the solution ideally shouldn't inhibit.

### Communication with GUI

- The [UI Module](https://github.com/HeyPuter/puter/tree/1e86039cf51b5043d17bcb0d904c5f45f1e9a52c/src/puter-js/src/modules/UI.js) puter.js GUI implements all the UI calls seen on [the puter.js docs](https://docs.puter.com).
- The **menubar** is a particularly good end-to-end example
- If you search `setMenubar`, you'll see it's handled by [IPC.js](https://github.com/HeyPuter/puter/tree/1e86039cf51b5043d17bcb0d904c5f45f1e9a52c/src/gui/src/IPC.js). Keep this in mind; it's relevant for the next item.
- Take a look at [ExecService](https://github.com/HeyPuter/puter/tree/1e86039cf51b5043d17bcb0d904c5f45f1e9a52c/src/gui/src/services/ExecService.js), in particular where it calls `register_ipc_handler` on [IPCService](https://github.com/HeyPuter/puter/tree/1e86039cf51b5043d17bcb0d904c5f45f1e9a52c/src/gui/src/services/IPCService.js). You may have noticed an anti-pattern in IPC.js (violates separation of concerns); this more modular approach will help us incrementally separate the logic of how IPC works from the events that are handled through IPC.

### Existing Code

- If you [search for "requestPermission"](https://github.com/search?q=repo%3AHeyPuter%2Fputer%20requestPermission&type=code) you'll notice a few places where we refer to or use UIWindowRequestPermission. Puter.js even calls `puter.ui.requestPermission`, which doesn't actually exist anymore. This old request permission dialog may serve as a useful reference.

tbrar06:
- Added SDK endpoint for apps to request permission to (1) access a file (2) post to thread (3) invoke interfaces through services (4) perform an action through a driver
- Open for extension for adding new permission types and web implementation (see issue)

Fixes #1150


KernelDeimos:
Perfect, this seems to be working without any issues. It would be nice to be able to specify multiple permissions and request them all in one list. I'm going consider that a the scope of a subsequent PR and merge this one as-is.

PR #1181, opened 2025-03-14
tweak : moved clock to right between profile and search

### LINKED ISSUE #1176: Move clock to make room for "unified menubar"
There are [apps that use a unified menubar](https://puter.com/app/jsoneditor) and this currently causes the clock to be pushed to the right

Image

If we move the clock here...

Image

between the magnifying glass and the profile picture, it's guaranteed not to move around.

We now know there are several scenarios for which the clock needs to be tested:
- each visibility setting: auto, hide, show
- on a user that has never configured clock visibility
- while using an app that uses the "unified menubar"

sashpawar11:
The tweaks under this commit resolve #1176 

**Changelog**
- Moved clock to the right side between the profile and search icons, ensuring its position is less likely to change due to other interactions on the toolbar
- Apps utilizing the mac style menubar will remain undisturbed.
- Cleaned up old positioning logic that was added for the initial clock enhancement

**Tested against each personalization settings**

1 : Hide - always hidden
image

2 : Show - always visible
image

3 : Auto - Default, visible only in full-screen mode
image

### DIFF
--- a/src/gui/src/UI/UIDesktop.js
+++ b/src/gui/src/UI/UIDesktop.js
@@ -1103,12 +1103,9 @@ async function UIDesktop(options){
     let ht = '';
     ht += `<div class="toolbar" style="height:${window.toolbar_height}px; min-height:${window.toolbar_height}px; max-height:${window.toolbar_height}px;">`;
         // logo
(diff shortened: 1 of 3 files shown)


sashpawar11:
@KernelDeimos - this is ready for review.
Thanks

KernelDeimos:
I just tested it and this is working great - thanks!

PR #1183, opened 2025-03-15
Added Context Menu for Search Results (Right-Click Feature)

Harmanpawar15:
## 📝 Summary
This PR adds a right-click context menu to search results using `UIContextMenu`. Users can now:
- Right-click on a search result to see a menu.
- Choose "Open File" to open the file (same as left-click).
- Choose "Open Containing Folder" to navigate to the file’s location in File Explorer.

## ✅ Changes
- Integrated `UIContextMenu` for search results.
- Preserved existing left-click behavior.
- Used `UIWindow` to open the containing folder.

## 🔍 How to Test
1. Search for a file.
2. Right-click on a result → A context menu should appear.
3. Click "Open File" → The file should open.
4. Click "Open Containing Folder" → File Explorer should open at the file’s location.

## 📌 Notes
- No existing functionality was altered.
- This aligns with the contributing guidelines and maintains UI consistency.

### DIFF
--- a/src/gui/src/UI/UIContextMenu.js
+++ b/src/gui/src/UI/UIContextMenu.js
@@ -439,6 +439,7 @@
  * @requires jQuery-menu-aim
  */
 
+
 function UIContextMenu(options){
     $('.window-active .window-app-iframe').css('pointer-events', 'none');
 
--- a/src/gui/src/UI/UIContextMenu.js
+++ b/src/gui/src/UI/UIContextMenu.js
@@ -460,9 +460,12 @@ function UIContextMenu(options){
             
         for(let i=0; i < options.items.length; i++){
             // item
+
             if(!options.items[i].is_divider && options.items[i] !== '-'){
+               
                 // single item
                 if(options.items[i].items === undefined){
+
                     h += `<li data-action="${i}" 
                             class="context-menu-item ${options.items[i].disabled ? ' context-menu-item-disabled' : ''}"
                             >`;
--- a/src/gui/src/UI/UIContextMenu.js
+++ b/src/gui/src/UI/UIContextMenu.js
@@ -360,6 +360,9 @@
     };
 })(jQuery);
 
+
+
+
 /**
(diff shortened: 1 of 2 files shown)


KernelDeimos:
Manually tested - looks good. Concerning the UX, I had two thoughts that are in conflict:
- It's convenient that the search window doesn't close when you open something, because you might want to open another result as well
- The fact that it doesn't close looks kind of like a bug because of the nature of this type of dialog (dialogs with an overlay typically disappear when you take an action)

This a solution I think will keep the convenience of the window not closing and make it seem more natural: if you click an item normally the search window goes away like it usually does, if you right-click an item and open it that way the dialog turns into a regular window. (i.e. no overlay, so it won't go away if you click outside of it; has a titlebar shown so there's an X button). Let me know if this makes sense or if you have any questions. I'm also going to tag @jelveh in case he wants to weigh in.

I recorded a manual test (1.6x speed)

https://github.com/user-attachments/assets/71e521ef-5913-490c-8d79-be07c14d515b

KernelDeimos:
Also I think the use of emojis in comments is something we should do more often. I makes separate sections of the code quicker to identify (really important because I think a developer's brain has to perform that task unconsciously a lot more frequently than we think about) and it's super on-point with Puter's brand/personality. (+ everything should support UTF-8 nowadays)

KernelDeimos:
Hi, are you still working on this? If not, we can merge it into a feature branch and see if someone else picks it up. Otherwise we just need left-click to close the dialogue - that's the most important change because it's a UX regression, and I can merge this after that. We can sort out the nuance of right-click behavior later.

Harmanpawar15:
Hey @KernelDeimos, thanks for the detailed feedback! 😊

Apologies for the delay—had some technical issues with my laptop, but I just got it back today. I’ll work on the left-click behavior fix ASAP and push an update soon.

I agree with your suggestion regarding right-click behavior, and I’ll keep that in mind while making the changes. Will follow up once the fix is ready! 🚀

Thanks again for the patience!

PR #1188, opened 2025-03-16
Draft Pull request Webdav issue

dawit2123:
I have solved the issue of the error that you've got. When I try the propfind command, there's an error that pops up from the file that's already done which is FSNodeContext that's: [INFO::fsnode-context] (571.757s) fetching entry: C:
WebDAV operation error: Error: FAILED TO GET THE CORRECT CONTEXT


KernelDeimos:
After taking a look I can tell that the issue is coming from improper initialization. Instead of starting the webdav server using a function in the `run-selfhosted.js` script, make it backend service in a new module. The directory will be `src/backend/src/modules/webdav`. There's [an example module called `template`](https://github.com/HeyPuter/puter/blob/6a184d52b47d80f23babaa94f9ccc32ed6ea14be/src/backend/src/modules/template) in the repo. Perform initialization in a method of the service called `init()`.

KernelDeimos:
@dawit2123 are you still working on this?

dawit2123:
@KernelDeimos 
Yes, I am working on it, and I don't forget it.
I wasn't active on it because of the exam and other things happening until the coming Friday. 
But I will update you on my latest code this week for testing.

PR #1189, opened 2025-03-17
Notification management

RadoBoiii:
Fixed the notification management system with a load more button to load additional notifications

- Added notification history sidebar to display past notifications
- Implemented pagination with "Load more" button
- Added infinite scroll support for seamless loading
- Fixed UI state inconsistencies
- Added proper error handling
- Improved mobile responsiveness

### DIFF
--- a/src/gui/src/UI/Settings/UITabUsage.js
+++ b/src/gui/src/UI/Settings/UITabUsage.js
@@ -62,9 +62,25 @@ export default {
                     if ( ! entry.usage_percentage ) {
                         entry.usage_percentage = (entry.used / entry.available * 100).toFixed(0);
                     }
+
+                    // Skip the 'ai-chat (complete)' entry since we've made it infinite for now
+                    if(entry.name === 'ai-chat (complete)')
+                        return;
+
+                    if(entry.name.startsWith('es:subdomain'))
+                        return;
+
+                    if(entry.name.startsWith('es:app'))
+                        return;
+
+                    let name = entry.name;
+
+                    if(name === 'convert-api (convert)')
+                        name = `File Conversions`;
+
                     h += `
                         <div class="driver-usage" style="margin-bottom: 10px;">
-                            <h3 style="margin-bottom: 5px; font-size: 14px;">${html_encode(entry.name)}:</h3>
+                            <h3 style="margin-bottom: 5px; font-size: 14px;">${html_encode(name)}:</h3>
                             <span style="font-size: 13px; margin-bottom: 3px;">${Number(entry.used)} used of ${Number(entry.available)}</span>
                             <div class="usage-progbar-wrapper" style="width: 100%;">
                                 <div class="usage-progbar" style="width: ${Number(entry.usage_percentage)}%;"><span class="usage-progbar-percent">${Number(entry.usage_percentage)}%</span></div>


RadoBoiii:
Resolving conflicts in the pull requests

experiments/x86emu/www/package-lock.json
KernelDeimos, line 1: Why was this modified? This seems out-of-scope for this PR. If this package-lock.json is out of sync we should update it but in a different PR.

src/backend/src/modules/puterai/AIChatService.js
```diff
@@ -883,4 +883,4 @@ class AIChatService extends BaseService {
     }
 }
 
-module.exports = { AIChatService };
+module.exports = { AIChatService };
```
KernelDeimos: text files should end with a line feed (LF) character so append operations behave as expected

src/backend/src/services/NotificationService.js
```diff
@@ -1,5 +1,4 @@
-// METADATA // {"ai-commented":{"service":"mistral","model":"mistral-large-latest"}}
```
KernelDeimos: Why was this comment removed? We have a tool that's able to generate automated comments (it's experimental; it's not very reliable but we might do more work on it later) and this metadata comment tells the tool if the file has already been commented on. We might use this same format for storing other metadata for source files in the future.

src/gui/src/UI/UIDesktop.js
```diff
@@ -42,12 +42,33 @@ import UIWindowWelcome from "./UIWindowWelcome.js"
...
+        console.error('Failed to fetch profile picture:', error);
+    }
+    return null;
+}
 
 async function UIDesktop(options){
     let h = '';
 
     // Set up the desktop channel for communication between different tabs in the same browser
-    window.channel = new BroadcastChannel('puter-desktop-channel');
```
KernelDeimos: Was it verified that this former-global wasn't being referenced anywhere else?

src/gui/src/UI/UIDesktop.js
```diff
@@ -1289,20 +1345,19 @@ async function UIDesktop(options){
...
+        var dt=x.getDate().toString();
         dt=dt.length==1 ? 0+dt : dt;
         
-        var day = x.toLocaleString('default',{weekday : 'short'});
-       
-   
-        var x1= day + ", " + month + " " + dt; 
-        x1 =  hours + ":" +  minutes   + ampm + " " + x1;
+        var x1=month + "/" + dt + "/" + x.getFullYear(); 
+        x1 = x1 + " - " +  hours + ":" +  minutes + ":" +  seconds + " " + ampm;
```
KernelDeimos: Why are we making changes to the clock here?

src/gui/src/UI/UIDesktop.js
```diff
@@ -1385,6 +1440,22 @@ $(document).on('click', '.qr-btn', async function (e) {
...
+    // Toggle active state of the button
+    $(this).toggleClass('active');
+    
+    // Create or show notifications sidebar
+    UIWindowNotifications();
+})
+
+$(document).on('click', '.notification', function() {
+    // Update badge after notification is clicked
+    setTimeout(window.update_notification_badge_count, 100);
```
KernelDeimos: Why are we polling so frequently here? If I merged this as-is, I wouldn't be surprised if it had a noticeable impact on `puter.com`. We get notifications over websockets, so we can usually update the count there; this should then be a very infrequent poll (several minutes) just in case the badge goes out of sync.

src/gui/src/UI/UIWindowNotifications.js
```diff
@@ -0,0 +1,442 @@
...
+        });
+
+        try {
+            if (!window.user?.id) {
+                notifLogger.error('No user ID found');
+                throw new Error('User ID not found');
+            }
+
+            const db = await open({
+                filename: window.dbPath || './puter-database.sqlite',
```
KernelDeimos: ngl this scared the crap out of me because I thought maybe this is working. After a lot of manual testing I can confirm that this does not work. Not only does this not work, but this won't and should not ever work. If you ever do this and it works it means there's a severe security issue in the project your contributing to.

KernelDeimos (changes requested):
There are a lot of `package-lock.json` changes that don't seem relevant to this PR. They should be removed. If package-lock.json files are out of sync this should be addressed in a separate PR.

There are attempts to communicate with the database from client-side in this PR. That will never work.

KernelDeimos:
Thanks for fixing, here's my second manual test. I noticed a few issues. I also noticed there was a `/count` endpoint being called and I couldn't find its implementation, it's possible many of these issues have a common root cause.

1. Clicking the bell once doesn't work, which I found counterintuitive. We don't have double-click behavior for any of the icons on that title bar and I don't think users would expect it.
2. The bell is always grey even if there are notifications. I remember seeing code to maintain a badge for the notification count, but I didn't see one in the UI, so I assume this isn't working at the moment.
3. New notifications don't "pop up" anymore. I had to open the notification sidebar to see that they were there.
4. I found an odd condition where marking a notification as read didn't work. I'm not sure what caused this.
5. I observed the notification side bar going blank after opening it and acknowledging a notification. I'm not sure what caused this.

https://github.com/user-attachments/assets/b775ee81-1f14-4982-8087-87ef154668e3

KernelDeimos:
I did another manual test. This looks a lot better than previously. I noticed the notification toast is still gone but I see discussion that that's still being worked on.

https://github.com/user-attachments/assets/733d5386-6532-481e-bfec-b55d03df0ddc

PR #1192, opened 2025-03-17
Unable to fetch voices for AWS Polly

matteopilotto:
Keep getting the following error:
`TypeError: puter.ai.listVoices is not a function`

### DIFF
--- a/polly-app/app.js
+++ b/polly-app/app.js
@@ -1,0 +1,174 @@
+document.addEventListener("DOMContentLoaded", () => {
+    // Elements
+    const textInput = document.getElementById("text-input");
+    const voiceSelect = document.getElementById("voice-select");
+    const languageSelect = document.getElementById("language-select");
+    const speakBtn = document.getElementById("speak-btn");
+    const testModeBtn = document.getElementById("test-mode-btn");
+    const stopBtn = document.getElementById("stop-btn");
+    const audioContainer = document.getElementById("audio-container");
+    const statusMessage = document.getElementById("status-message");
+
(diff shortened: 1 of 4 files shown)


KernelDeimos:
I wasn't able to reprod this in a manual test. You mentioned you already tried hard-reload without cache. Was dev-tools open you used Ctrl+Shift+R? I've found that the hotkey doesn't actually work unless devtools is open (there's no visibility on this, so that's actually a UX bug in chromium IMO). It's also possible there's something else different about your environment that's causing puter.js builds to fail.

https://github.com/user-attachments/assets/5e342d11-77b5-4627-8266-380a6a8427c1

matteopilotto:
@KernelDeimos thanks for the recording.

I tested the function following your approach, step-by-step, and it actually work.
I assume the problem lies in the way I'm using it in the polly-app.
In the app I'm also calling `puter.ai.txt2speech(text)`, but in that case the function works as expected without any error.

If you have a chance, could you also take a look at the code inside the polly-app directory?
Thanks

KernelDeimos:
@matteopilotto I see the problem, your script url is `https://js.puter.com/v2/` - well, you won't get `puter.ai.newFeature` until it's deployed to `js.puter.com` (which in this case, it is not). In development you can use `/puter.js/v2` as the URL (with no protocol or domain) to get the local version of puter.js.

matteopilotto:
Oh! I see...
Now things are working properly now.
Thanks a lot for the support Eric.
Much (much) appreciated.

PR #1194, opened 2025-03-19
Ai command tool function

ntwari-bruce:
Enhancement: Implement Function Calling for AI Command Execution

Previously, the AI command execution relied on pattern matching (%%% markers) to identify and execute shell commands. This approach had limitations in terms of reliability, parameter validation, and command handling.

Changes made:
1. Replaced command pattern matching with OpenAI's function calling feature
2. Integrated with shell's builtin commands system to:
   - Automatically generate tool definitions from existing command metadata
   - Properly handle command parameters and options
   - Support structured command execution

3. Added streaming support for real-time AI responses
4. Improved user interaction with:
   - Clear command execution confirmations
   - Better error handling
   - Proper command parameter validation

5. Enhanced chat history handling to maintain the context of tool usage

Example usage:
```bash
ai "create a directory named test"
# Now properly uses mkdir command through function calling
# Instead of pattern matching "%%%mkdir test%%%"
```

Benefits:
- More reliable command execution
- Better parameter validation
- Structured command handling
- Improved error messages
- Maintainable code structure that leverages existing command definitions

This enhancement sets the foundation for adding more command capabilities and improving AI-shell interaction.


KernelDeimos:
I did a manual test for this. I'm not able to see tool calls from the LLM in the network tab which I think is due to an issue on my own device rather than being an issue in this PR; I'm considering this a manual test pass.

https://github.com/user-attachments/assets/e04ca529-8a34-45ab-a3d1-2f6ab94fd9d9

src/phoenix/src/puter-shell/coreutils/ai.js
```diff
@@ -17,6 +17,7 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 import { Exit } from './coreutil_lib/exit.js';
+import builtins from '../coreutils/__exports__.js';
```
KernelDeimos: This is a circular dependency. We could solve this by adding a `list()` method to BuiltinCommandProvider. Then, we export command providers and the `ai` command will call `list()` on any "enumerable command provider" (any "SomethingCommandProvider" that implements `list()`)

ntwari-bruce:
@KernelDeimos I have pushed the new changes fixing the circular dependency and added list function to BuiltinCommand

KernelDeimos:
Thanks for the changes. I noticed a new issue that I missed before in manual testing. Your change for the circular dependency moves it rather than fixing it so I provided some more explanation.

### A new issue I noticed

I noticed sometimes it fails to recognize that the LLM provided a command. I was suspicious of this `%%%` output and I think I know what's going on.

image

This is actually coming from [a system prompt](https://github.com/HeyPuter/puter/blob/b0efff7a4f656cd180ccfadd98609ef37e1155dd/src/phoenix/src/puter-shell/plugins/ChatHistoryPlugin.js#L30) in the part of Phoenix shell that generates the command history for the AI command:

```javascript
        {
            role: 'system',
            content:
                // note: this really doesn't work at all; GPT is effectively incapable of following this instruction.
                'You can provide commands to the user by prefixing a line in your response with %%%. The user will then be able to run the command by accepting confirmation.',
        },
```

I think we can fix this occasional failure simply be removing that system prompt.

### Circular Dependency

> @KernelDeimos I have pushed the new changes fixing the circular dependency and added list function to BuiltinCommand

You didn't actually fix the circular dependency, you moved it; this is what happened:

- **before:** `coreutils/ai.js` **➙** `coreutils/__exports__.js` **➙** `coreutils/ai.js`
- **after:** `coreutils/ai.js` **➙** `providers/BuiltinCommandProvider.js` **➙** `coreutils/__exports__.js` **➙** `coreutils/ai.js`

I was suggesting that we expose the instances of command providers created in `src/phoenix/src/puter-shell/main.js` to `ctx.externs`. Actually, it looks like `commandProvider` is already exposed, which is an instance of **CompositeCommandProvider**. CompositeCommandProvider is the "try this command provider first, then this one, then this one" implementation. If you add a `list()` method to CompositeCommandProvider that calls list on any command providers under it that also implement list (which will just be BuitinCommandProvider for now) and return all the results, then you can access `ctx.externs.commandProvider.list` from inside `ai.js` command.

### Manual Test Recording (1.5x speed)

https://github.com/user-attachments/assets/c54f9340-541c-47e7-b08d-89480ea0bc24

KernelDeimos:
Manual test 3 (I say "2" in the video; that's an error) passes. Merging this now.

https://github.com/user-attachments/assets/fedb8252-6aff-4904-bae6-e1d6110e718f

Issue #1196, opened 2025-03-19
Make it easier to add file extensions to an app in bulk


KernelDeimos:
This issue affect the Dev Center app (`src/dev-center`) when creating/editing an app.

Right now it's time consuming to add many extensions to an app because you "have to press Enter every time". Pasting a CSV with file extensions should work - that would be much better.

KernelDeimos:
I've submitted PR #1205 to address this issue. The PR adds support for pasting multiple file extensions at once, which makes it much easier to add many extensions in bulk. Now users can simply paste a comma, space, or tab-separated list of extensions instead of having to press Enter after each one.

PR #1205, opened 2025-03-24
Make it easier to add file extensions to an app in bulk

KernelDeimos:
### Please Note
I'm testing out an agentic AI code writing tool called [mycoder](https://github.com/drivecore/mycoder). Most of the comments (except for review comments) below are not written by me personally. Most of the code (except a small fix) is generated. Everything was manually tested by me and I reviewed all the code.

### Comment from `mycoder`
This PR addresses issue #1196 by adding support for pasting multiple file extensions at once in the Dev Center app.\n\n## Changes\n\n- Added support for pasting CSV-formatted file extensions (comma, semicolon, tab, or space separated)\n- Added a custom paste event handler to process and validate pasted extensions\n- Updated the placeholder text to inform users they can paste multiple extensions at once\n- Updated the UI text to make the feature more discoverable\n\nNow when users want to add many file extensions to their app, they can simply paste them all at once instead of having to press Enter after each one, which significantly improves the user experience when configuring apps.\n\nFixes #1196

### DIFF
--- a/src/dev-center/js/dev-center.js
+++ b/src/dev-center/js/dev-center.js
@@ -569,8 +569,8 @@ function generate_edit_app_section(app) {
 
                 <label for="edit-app-filetype-associations">File Associations</label>
                <p style="margin-top: 10px; font-size:13px;">A list of file type specifiers. For example if you include <code>.txt</code> your apps could be opened when a user clicks on a TXT file.</p>
-               <p style="margin-top: 5px; font-size:13px;">You can paste multiple extensions at once (comma, space, or tab separated).</p>
-               <textarea id="edit-app-filetype-associations"  placeholder="Paste multiple extensions: .txt, .doc, .pdf, application/json">${JSON.stringify(app.filetype_associations.map(item => ({ "value": item })), null, app.filetype_associations.length)}</textarea>
+               <p style="margin-top: 5px; font-size:13px;">You can paste multiple extensions at once (comma, space, or tab separated) or press comma to add each extension.</p>
+               <textarea id="edit-app-filetype-associations"  placeholder="Paste multiple extensions like: .txt, .doc, .pdf, application/json">${JSON.stringify(app.filetype_associations.map(item => ({ "value": item })), null, app.filetype_associations.length)}</textarea>
 
                 <h3 style="font-size: 23px; border-bottom: 1px solid #EEE; margin-top: 50px; margin-bottom: 0px;">Window</h3>
                 <div>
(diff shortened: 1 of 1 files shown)


KernelDeimos:
I've updated the implementation to fix the issue. The new approach uses a proper paste event handler that:

1. Detects when text with delimiters (commas, spaces, tabs, etc.) is pasted
2. Parses the pasted text into individual extensions
3. Validates each extension against the pattern
4. Adds valid extensions as tags using the tagify.addTags() method

This should now work correctly for bulk pasting file extensions. I've also improved the help text to make the feature more discoverable.

KernelDeimos:
I've fixed the issue with the temporary red duplicate tags appearing:

1. Added explicit duplicate prevention in the Tagify initialization with 
2. Added checks to prevent adding extensions that already exist as tags in both paste and keydown handlers
3. Improved the input clearing behavior to prevent any lingering text

These changes should provide a much smoother user experience when pasting multiple extensions or typing with commas. The temporary red tags should no longer appear, and the system will silently skip any duplicates instead of showing them temporarily.

KernelDeimos:
I've fixed the issue with the temporary red duplicate tags appearing:

1. Added explicit duplicate prevention in the Tagify initialization with 'duplicates: false'
2. Added checks to prevent adding extensions that already exist as tags in both paste and keydown handlers
3. Improved the input clearing behavior to prevent any lingering text

These changes should provide a much smoother user experience when pasting multiple extensions or typing with commas. The temporary red tags should no longer appear, and the system will silently skip any duplicates instead of showing them temporarily.

KernelDeimos:
I've fixed the edge case you discovered with backspace and paste operations:

1. Added special handling to detect when there's existing text in the input field before pasting
2. Properly handle the case where a user types something, presses backspace to delete a tag, and then pastes text
3. Improved the paste handler to handle both delimited and non-delimited pastes more intelligently
4. Added more validation to ensure proper parsing in all edge cases

This should now handle the scenario you described where pressing backspace after creating a tag and then pasting would result in concatenated tags like '.jpg.txt'.

KernelDeimos:
I've completely rewritten the paste and keydown handlers to fix the issue with concatenation:

1. Now always preventing the default paste behavior to take full control of the process
2. Added explicit input clearing before and after every paste operation
3. Added a focus handler to ensure the input is always cleared when focused
4. Simplified the paste logic to avoid any potential for concatenation
5. Added more robust error handling throughout

These changes should fix the issue where typing '.jpg', pressing comma, then pasting '.txt,.pdf' would result in a '.jpg.txt' tag. The input is now cleared at multiple points to ensure no text remains that could be concatenated with pasted content.

KernelDeimos:
I've completely reimplemented the bulk file extension paste functionality with a more robust approach:

1. Created a completely separate paste handler that captures paste events at the capture phase, before Tagify's built-in handler
2. Added a focus handler to ensure proper cleanup of the input field
3. Improved the comma key handler for adding tags one by one
4. Updated the help text to clearly explain the feature
5. Added better error handling throughout

This implementation avoids the issues with concatenated tags and should work reliably in all scenarios. The paste handler now properly intercepts paste events and processes them correctly, regardless of the current state of the input field.

src/dev-center/js/dev-center.js
```diff
@@ -569,7 +569,8 @@ function generate_edit_app_section(app) {
 
                 <label for="edit-app-filetype-associations">File Associations</label>
                <p style="margin-top: 10px; font-size:13px;">A list of file type specifiers. For example if you include <code>.txt</code> your apps could be opened when a user clicks on a TXT file.</p>
-               <textarea id="edit-app-filetype-associations"  placeholder=".txt  .jpg    application/json">${JSON.stringify(app.filetype_associations.map(item => ({ "value": item })), null, app.filetype_associations.length)}</textarea>
+               <p style="margin-top: 5px; font-size:13px;">You can paste multiple extensions at once (comma, space, or tab separated) or press comma to add each extension.</p>
+               <textarea id="edit-app-filetype-associations"  placeholder="Paste multiple extensions like: .txt, .doc, .pdf, application/json">${JSON.stringify(app.filetype_associations.map(item => ({ "value": item })), null, app.filetype_associations.length)}</textarea>
```
KernelDeimos: This JSON.stringify here is insecure because JSON strings can contain angled brackets. In other places we handle this using the following approach:

```javascript
JSON.stringify(gui_params).replace(/</g, '\\u003c')
```

KernelDeimos (commented):
Thanks AI, just one suggestion for security reasons

KernelDeimos:
I've fixed the security issue with the JSON.stringify call. The code now properly escapes < characters using  to prevent potential XSS vulnerabilities.

KernelDeimos:
I've fixed the security issue with the JSON.stringify call. The code now properly escapes < characters using the \\u003c escape sequence to prevent potential XSS vulnerabilities.

Issue #1207, opened 2025-03-24
Fix check_usage_ in AIChatService to only consider past month usage


KernelDeimos:
## Issue Description\n\nThe `check_usage_` function in `src/backend/src/modules/puterai/AIChatService.js` currently queries all usage data when checking user limits. According to requirements, it should only consider usage from the past month.\n\n## Current Implementation\nThe current implementation queries the total sum of costs without any time filtering:\n```javascript\nconst [row] = await this.db.read(\n    'SELECT SUM(`cost`) AS sum FROM `ai_usage` ' +\n    'WHERE `user_id` = ?',\n    [actor.type.user.id]\n);\n```\n\n## Expected Behavior\nThe query should include a date filter to only consider usage from the past month.\n\n## Table Structure\nThe `ai_usage` table has a `created_at` column that can be used for filtering by date.\n\n## Proposed Fix\nUpdate the SQL query to include a date filter for the past month.

PR #1208, opened 2025-03-24
fix: update check_usage_ to only consider past month usage

KernelDeimos:
## Description\nThis PR fixes the `check_usage_` function in `AIChatService.js` to only consider AI usage from the past month when checking against limits, rather than all historical usage.\n\n## Changes\n- Modified the SQL query to include a date filter for the past month\n- Added code to calculate the date one month ago\n\n## Related Issues\nFixes #1207\n\n## Testing\nManually verified that the SQL query now includes the date filter correctly.\n\n[ai]


KernelDeimos:
[ai] Closing this PR as it was created from the wrong branch. Please see PR #1209 instead.

PR #1211, opened 2025-03-24
feat: add costly model to FakeChatService

### LINKED ISSUE #1210: Alternative to FakeChatService that incurs a "cost"
It would be helpful to have an implementation of `puter-chat-completion` like FakeChatService which reports to have incurred cost just like the other implementations (without, of course, incurring any real-world cost). This could be one of the available "models" on FakeChatService.

The Puter AI module is not documented yet, but [this open issue for documenting it contains useful links](#1133).

Comparing FakeChatService to [ClaudeService](src/backend/src/modules/puterai/ClaudeService.js) will be a helpful reference. Look at [AIChatService](src/backend/src/modules/puterai/AIChatService.js) if you want to determine the ramifications of implementing the `models` method (i.e. do we need entries for the existing `fake` and `abuse` pseudo-models?).

The new model could be called `costly`.

KernelDeimos:
[ai]

## Description
This PR adds a new 'costly' model to FakeChatService that simulates incurring costs similar to real AI services. This model can be used for testing cost tracking without incurring actual external API costs.

## Changes
- Added a `models()` method to return detailed information about the available models, including their costs
- Updated the `list()` method to include the new `costly` model in the list of available models
- Enhanced the `complete()` method to calculate token usage and report costs for the `costly` model

## Fixes
Closes #1210

## Testing
The implementation has been tested by building the project successfully. The `costly` model will:
- Calculate input tokens based on message content
- Generate a random output token count between 50-200
- Report usage via the 'ai.prompt.report-usage' event
- Return the token usage information in the response

## Note
This implementation follows the patterns established in the existing code and allows for cost tracking testing without incurring real API costs.

### DIFF
--- a/src/backend/src/modules/puterai/AIChatService.js
+++ b/src/backend/src/modules/puterai/AIChatService.js
@@ -78,7 +78,8 @@ class AIChatService extends BaseService {
 
         const svc_event = this.services.get('event');
         svc_event.on('ai.prompt.report-usage', async (_, details) => {
-            if ( details.service_used === 'fake-chat' ) return;
+            // Only skip usage reporting for fake-chat if it's not using the costly model
+            if ( details.service_used === 'fake-chat' && details.model_used !== 'costly' ) return;
 
             const values = {
                 user_id: details.actor?.type?.user?.id,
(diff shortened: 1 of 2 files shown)


KernelDeimos:
[ai] Added _init method to FakeChatService to register itself as a provider with AIChatService, ensuring that the service and its models are properly recognized by the system. This is consistent with how other chat services like ClaudeService register themselves.

KernelDeimos:
[ai] Modified the event handler in AIChatService to only skip usage reporting for fake-chat when it's not using the costly model. This allows the costly model to properly report its usage and costs while still excluding the regular fake model from cost tracking.

Issue #1203, opened 2025-03-22
Bug: Fatal error in starting the app


kristavlastou:
### Issue Description

Hello, although Puter used to run completly fine with npm install and npm run, I merged the most resent version locally and now I cannot run the project. Specifficly :                                                                                           CONFIGURATION MUTATED AT RUNTIME mod_directories to [ '{source}/../mods/mods_enabled', '{source}/../extensions' ]
WRITING TO mod_packages\testex\package.json
C:\Users\krist\.vscode\Puter\src\backend\src\Kernel.js:406
                    throw new Error(exit code: ${code});
                    ^

Error: exit code: -4058
    at ChildProcess.<anonymous> (C:\Users\krist\.vscode\Puter\src\backend\src\Kernel.js:406:27)
    at ChildProcess.emit (node:events:518:28)
    at ChildProcess.emit (node:domain:489:12)
    at maybeClose (node:internal/child_process:1101:16)
    at ChildProcess._handle.onexit (node:internal/child_process:304:5)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
Can anyone help it's important. Thank you very much

### Deployment

- [ ] Production (puter.com)
- [ ] Development (`npm run start`)
- [ ] Docker (via `docker run`)
- [ ] Docker (via `docker-compose`)

kristavlastou:
Hello again, I tried to figure out what happend and this is my resolution : 
Problem
npm install failed in Kernel.js with the error:
Error: spawn EINVAL
The issue occurred because spawn('npm', ['install']) doesn’t work properly on Windows, as npm is a batch script (npm.cmd)
Solution

Modified Kernel.js to use the correct command based on the OS:

const npmCmd = process.platform === "win32" ? "npm.cmd" : "npm";
const proc = spawn(npmCmd, ["install"], { cwd: path, shell: true, stdio: "inherit" });

Result

With this change, npm install now executes correctly across all operating systems. Tell me if my solution fits you or if other people had the same problem as well so that we can include it in the main code. Thank you!

KernelDeimos:
Hi, I'm not sure what's happening here since we rarely test under windows, but seeing `npm.cmd` seems very strange to me, even in a Windows environment. Unless we hear form another Windows user I have to suspect this is an issue with your environment specifically.

kristavlastou:
[...]

PR #1214, opened 2025-03-24
Fix npm install error on Windows (spawn EINVAL) (#1203)

kristavlastou:
Resolved an issue where `spawn('npm', ['install'])` failed on Windows due to npm being a batch script.  
- Updated `Kernel.js` to determine the correct npm command based on the OS.  
- Now uses `npm.cmd` on Windows and `npm` otherwise.  

This ensures `npm install` runs correctly across all operating systems.


DarkCodeStudios:
You have duplicated the proc declaration in your function, which causes a SyntaxError. 

```
    async run_npm_install (path) {
        const npmCmd = process.platform === "win32" ? "npm.cmd" : "npm";
        const proc = spawn(npmCmd, ["install"], { cwd: path, shell: true, stdio: "inherit" });
    
        return new Promise((rslv, rjct) => {
            proc.on('close', code => {
                if (code !== 0) {
                    return rjct(new Error(`exit code: ${code}`));
                }
                rslv();
            });
            proc.on('error', err => {
                rjct(err);
            });
        });
    }    
```

kristavlastou:
Thank you for your feedback but I cannot spot the duplicated proc declaration as I declare it only in line 403 as I modified only  2 lines  in the function. Also Vscode also doesn't point out any Syntax error.  Can you explain it further?

KernelDeimos:
@DarkCodeStudios is it possible you misinterpreted the diff? I don't see a duplication declaration either.

image

DarkCodeStudios:
@KernelDeimos Strangely once I entered everything you have from the terminal it gave me a syntax error for having repeated the function 2 times in the entire kernel code `proc `

Issue #682, opened 2024-08-14
Links shortcuts as icons on desktop


BaskovKonstantin:
I think that the ability to save links as icons on the desktop and in directories was an extremely useful feature, since thanks to this it will be possible to work in the browser as on a computer, which would be extremely useful for people who work most of their working time inside the browser

jelveh:
This should probably be implemented via copy/pasting or dragging links onto the desktop. So when a link is dropped or pasted on Puter it turns into the item representing that link.

KernelDeimos:
Good pick, I forgot about this one. You're assigned now!

PR #1219, opened 2025-03-25
feat: add desktop link shortcuts (refs #682)

rodrick-mpofu:
Summary

- Adds a “New Link” option to the desktop context menu, prompting for a URL.

- paste a URL now creates a .url shortcut.

- Double-clicking a link shortcut item opens the URL in a new tab.

How to Test

- Right-click empty desktop → “New Link” → enter any valid URL.

- Copy a URL from your browser’s address bar onto the desktop.

- Paste a URL onto the desktop

- Double-click the new link icon. It should open the URL in a new browser tab.

References

- Issue #682 – this PR implements the requested link shortcut feature.

Screenshot 2025-03-25 024512


KernelDeimos:
Hello, thanks for your contribution. This might take us a little while to review because:
- amount of new code is relatively high for the feature being added
- "cognitive complexity" metric is high (nesting, branching, and complex expressions contribute to this)

This is more likely to get merged into v2.6 if you are able to reduce either of these metrics. I will be doing a manual test for this first instead of a code review, since that seems to be the low-hanging fruit here.

KernelDeimos:
I'm noticing an issue during manual testing. When creating a web link we actually get two icons, one of which isn't "real" (goes away after a page reload, doesn't go away after a desktop refresh though) and has a graphic error in the icon. How were these changes tested?

https://github.com/user-attachments/assets/f36186d2-f332-4594-b950-71ee2e6c196c

jelveh:
@rodrick-mpofu thank you for your contribution, this is a very useful feature!
I made a quick modification to disable favicons for now. The reason for that is that given our scale, we can't rely on Google's favicon service which is not well-documented in terms of reliability. Using the generic link icon for now would be sufficient and later on we could revisit favicons. 

I'm going to merge this soon.

PR #1228, opened 2025-03-27
feat: add public endpoint for models list

### LINKED ISSUE #1227: Public endpoint for models list
We need a public unauthenticated endpoint to list available AI chat models.

- Add ChatAPIService for public endpoints that don't go through the driver interface (they might user drivers behind the scenes)
- [SUService](https://github.com/HeyPuter/puter/blob/f5a7b06d40aafd90b73801f26b176da1cd099f50/src/backend/src/services/SUService.js) can be used to invoke a driver as the system user (which is fine for listing models); an example use is by [SystemDataService](src/backend/src/services/SystemDataService.js). Of course, it's not necessary to go through the driver service in a backend endpoint; if it's easier just call the trait directly on all the AI chat implementors.
- AI chat implementors are all registered with [AIChatService](src/backend/src/modules/puterai/AIChatService.js)
- The permission [group endpooints](https://github.com/HeyPuter/puter/blob/daa3449f29b6496d9d7db6d7c2252812fa6e4966/src/backend/src/services/PermissionAPIService.js) are a good reference for how to currently add endpoints

KernelDeimos:
This PR adds a public unauthenticated endpoint to list available AI chat models.

## Changes
- Created ChatAPIService for public endpoints
- Added /chat/models and /chat/models/details endpoints
- Registered service in CoreModule
- Added tests for the new service

Closes #1227

[ai]

### DIFF
--- a/src/backend/src/services/ChatAPIService.js
+++ b/src/backend/src/services/ChatAPIService.js
@@ -49,7 +49,7 @@ class ChatAPIService extends BaseService {
         })();
 
         // Register the router with the Express app
-        app.use('/chat', router);
+        app.use('/puterai/chat', router);
 
         // Install endpoints
         this.install_chat_endpoints_({ router });
--- a/src/backend/src/services/ChatAPIService.js
+++ b/src/backend/src/services/ChatAPIService.js
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2024-present Puter Technologies Inc.
+ * 
+ * This file is part of Puter.
+ * 
+ * Puter is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


src/backend/src/services/ChatAPIService.js
```diff
@@ -0,0 +1,115 @@
...
+    async ['__on_install.routes'] (_, { app }) {
+        // Create a router for chat API endpoints
+        const router = (() => {
+            const require = this.require;
+            const express = require('express');
+            return express.Router();
+        })();
+
+        // Register the router with the Express app
+        app.use('/chat', router);
```
KernelDeimos:
```suggestion
        app.use('/puterai/chat', router);
```

PR #1237, opened 2025-03-29
Fix#857:  Add missing Norwegian Bokmål translations

Anarvsingh:
Added the missing Norwegian Bokmål translations src/gui/src/i18n/translations/nn.js.


Anarvsingh:
Apologies I didn't saw that the issue was already assigned to someone else

PR #1244, opened 2025-04-01
refactor: generalize handling of costly services

KernelDeimos:
### What's Changed

Usage/cost tracking per user was over-fitted to LLM/AI services. This commit adds CostService and updates AIChatService to use that instead of updating the `ai_usage` table. The `ai_usage` table is now considered deprecated.

A service called PermissiveCreditService is enabled by default in SelfHostedModule. This is so users who host their own Puter instances don't have to worry about usage limits, since they're providing their own keys anyway.

### What it Means for the Community

- When hosting your own Puter, there will be no usage limits by default.
- After exhausting usage on production Puter (`puter.com`), it will become available again after a shorter period of time.


KernelDeimos:
This was merged manually

PR #1250, opened 2025-04-23
fix/bug#1233-iframe-overlap-issue

Niveditarani:
This PR fixes the bug#1233- iframe overlap issue.

CSS property overflow: hidden at the parent container was interfering with the layout calculations and was forcefully clipping child elements. Making the resize div to float beside the iframe instead of overlapping.

Before:
image

After fix:
image

### DIFF
--- a/src/gui/src/css/style.css
+++ b/src/gui/src/css/style.css
@@ -904,13 +904,13 @@ span.header-sort-icon img {
     padding: 0;
     border: 1px solid #9a9a9a;
     box-shadow: 0px 0px 15px #00000066;
-    /* overflow: hidden; */
     border-radius: 4px;
     border: none;
     transition: opacity .2s;
     user-select: none !important;
     -moz-user-select: none !important;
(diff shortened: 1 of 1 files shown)


Niveditarani:
overflow: hidden property in the parent container was enforcing strict containment- making child elements to respect the parent's geometric boundaries (including border- radius).
		
Now without this property, children will naturally overflow their parent's boundaries including overflowing past rounded corners. So the border-radius wasn't actually removed - it was just being masked by overflowing content, breaking expected visual boundaries (e.g., titlebar background escaping the border-radius).
		
This commit replaces overflow: hidden with contain: paint, which maintains a clipping boundary similar to overflow: hidden, but without affecting layout flow—allowing the resize div to properly overlap the iframe instead of floating beside it.

KernelDeimos:
Thanks! I was able to test locally with JSONEditor added in my Dev Center and that's working better, everything else looks good too.

PR #1257, opened 2025-05-02
dev: improve Dockerfile efficiency and resilience

chintanboghara:
This pull request (PR) refines the Dockerfile to improve build efficiency, enhance security, and increase resilience to changes in the repository while maintaining full functionality.

### Build Optimization
- **Improved Layer Caching**: Separated the copying of `package.json` from the source code to leverage Docker's layer caching mechanism.
- **Build Performance**: Added environment variables (`NODE_ENV=production`, `NPM_CONFIG_LOGLEVEL=error`, `NPM_CONFIG_PROGRESS=false`) to reduce npm verbosity and speed up builds.
- **Better Build Output**: Included echo statements in the dependency installation retry mechanism for clearer build logs.

### Enhanced Security
- **User Permissions**: Created a proper directory structure and set ownership before copying files.
- **File Ownership**: Used the `--chown=node:node` flag during COPY operations to ensure correct file permissions.
- **Earlier User Transition**: Applied security best practices by switching to a non-root user environment earlier in the build process.

### Production Image Improvements
- **Selective File Copying**: Only necessary files (dist, node_modules, package.json, src, config, LICENSE) are copied to the production image instead of the entire repository.
- **Direct Command Execution**: Changed the startup command from `npm start` to the direct execution of `node src/index.js`.
- **Dependency Management**: Removed a redundant `npm install` in the production stage that attempted to resolve an `lru-cache` issue.
- **Additional Metadata**: Added a maintainer label for better identification of the image.

### Reliability Enhancements
- **Improved Health Check**: Replaced `wget` with `curl` and added better parameters (start period, more retries, longer timeout).
- **Enhanced Dependency Installation**: Improved retry logic with clearer error messages during the npm installation process.
- **Simplified Build Commands**: Streamlined the GUI build command by removing unnecessary `cd` commands.

## Testing
The refined Dockerfile has been tested to ensure:
- All application functionality works as expected
- The build process completes successfully
- The container runs without any permission issues
- The health check operates correctly

PR #1258, opened 2025-05-02
dev: enhance .dockerignore with comprehensive exclusions

chintanboghara:
- Organize exclusions by logical categories with comments
- Add version control and editor configuration files
- Exclude logs, environment files, and temporary files
- Improve security by preventing sensitive files from inclusion
- Reduce Docker build context size for faster builds

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,55 @@
-.dockerignore
+# Version control
+.git
+.github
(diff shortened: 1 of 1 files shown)


KernelDeimos:
I would prefer if this was combined with #1257 so that your changes can be tested together.

PR #1261, opened 2025-05-07
Update first-run-issues.md

atagtm:
Fix the missing 'sharp' dependency on FreeBSD, since it doesn't have prebuilt binaries for this OS.


KernelDeimos:
Nice, thanks! Out of curiosity, why were you running Puter under Unix? Was it a particular experiment/project or do you use FreeBSD as your main?

atagtm:
> Nice, thanks! Out of curiosity, why were you running Puter under Unix? Was it a particular experiment/project or do you use FreeBSD as your main?

Because this is my main OS 🥹

PR #1292, opened 2025-05-30
fix: enable and fix date command tests #796

### LINKED ISSUE #796: Test for `date` command in phoenix fails
I've disabled this test for now (823ab284483120f582afbb6d7c912a419567f0d2), as it's low-priority and I wasn't able to figure out why it's not working after some time investigating the issue.

This may be low-priority but it's a good first issue for anyone who's interested.

krikera:


KernelDeimos:
Hi, thanks for your contribution. I have two comments so far:
- You mentioned tests were re-enabled but I don't see the corresponding change that does this - can you point it out for me?
- [Formatting/style changes should be in separate PRs.](https://github.com/HeyPuter/puter/blob/main/CONTRIBUTING.md#separate-prs-for-formtting-and-code)

krikera:
Hi, @KernelDeimos 
You're right — I mentioned that the tests were re-enabled, but after checking again, I realized that change isn't actually included in this PR. That was a mistake on my part — apologies for the confusion.
I've removed the reference from the PR description, and I’ll open a separate PR specifically for that fix to keep things clean.

PR #1319, opened 2025-06-28
fix: arrow key navigation in Explorer (#1089)

### LINKED ISSUE #1089: Bug: Arrow Keys Misbehave on Explorer
### Issue Description

Works fine in the default view mode,
but not in the following cases:
image
image

For example, <kbd>↓</kbd> only changes the scroll position but doesn't change the selection
while <kbd>→</kbd> does select the next item **below** in the details view (the 2nd screenshot).

Hope I've made myself clear.

### Steps to reproduce

1. open Explorer
2. change the view mode (or whatever it's called) to the 2nd or the 3rd one
3. try selecting items using keyboard

### Expected behaviour

Each key "works as intended"

### Addition Information or Screenshots (if applicable)

_No response_

### Deployment

- [X] Production (puter.com)
- [X] Development (`npm run start`)
- [ ] Docker (via `docker run`)
- [ ] Docker (via `docker-compose`)

### Puter version

2.5.0

Krisss46:
Fixes #1089

The original issue described poor arrow key navigation in the Explorer view.  
The root cause was in `keyboard.js` where the logic for selecting the new/next item in containers was inconsistent.

This fix updates the selection code to work consistently across different container types, including dynamically structured ones.

**Additional feature:**  
Added wrapping navigation — when pressing left at the left edge, the selection wraps around to the right edge of the container.

**Note:**  
When selecting multiple items with SHIFT, the mainly selected item may overlap with previously selected items, so the selection doesn’t jump over pre-selected items.

### DIFF
--- a/src/gui/src/keyboard.js
+++ b/src/gui/src/keyboard.js
@@ -212,95 +212,119 @@ $(document).bind('keydown', async function(e){
         // ----------------------------------------------
         // Navigate items in the active item container
         // ----------------------------------------------
-        else if(!$(focused_el).is('input') && !$(focused_el).is('textarea') && (e.which === 37 || e.which === 38 || e.which === 39 || e.which === 40)){
-            let item_width = 110, item_height = 110, selected_item;
-            // select first item in container if none is selected
-            if($(window.active_item_container).find('.item-selected').length === 0){
-                selected_item = $(window.active_item_container).find('.item').get(0);
-                window.active_element = selected_item;
-                $(window.active_item_container).find('.item-selected').removeClass('item-selected');
-                $(selected_item).addClass('item-selected');
-                return false;
-            }
-            // if Shift key is pressed and ONE item is already selected, pick that item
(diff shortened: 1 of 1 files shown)


jelveh:
Hey @Krisss46, thank you for taking this up, it's a challenging one, and your PR almost completely fixes it right off the bat!

So far, I've found only one bug. In the list view of the Explorer, the down arrow key does not always work as expected, as demonstrated in the video below. The video should be self-explanatory but let me know if I can help further debug this.


https://github.com/user-attachments/assets/74f5d6d1-9c24-41a8-a47d-c7367cd7416a

Krisss46:
previous solution relied on center point of items but the longer the file name is the more offset were center points to the right.
if the offset exceeded `const axisThreshold = 30` in pixels, then those items would be considered invalid.

ive switched to origin points instead since those are all aligned

jelveh:
Great! I couldn't find any other issues. I'm going to merge it now.
thanks for this awesome PR!

PR #1338, opened 2025-07-08
Add a basic Firebase Studio config file

Zac0511:
I added a basic config file for any user that wants to help developping Puter and uses Google Firebase Studio
Thoses files does not change anything to Puter itself.

I did try setting up the preview system that would allow Puter to run inside Firebase Studio, but i failed, so i disabled that in the config file.

I also added an icon to show the Puter icon in Firebase Studio : 
image

I also noticed a new Studio Workspace can be quickly created from a link, such as this one for Puter : https://studio.firebase.google.com/import?url=https://github.com/HeyPuter/puter


jelveh:
Awesome! Thanks @Zac0511

PR #1353, opened 2025-07-15
fix: remove the "delegate" branch

XiaochenCui:


KernelDeimos:
@XiaochenCui you can commit directly to `refactor/puter-ai-router`, let me know if it doesn't let you

PR #1363, opened 2025-07-17
Update Portuguese and Spanish translation README

IgorJFS:
Updates based on the main README

1. Puter twitter link switched to https://x.com/HeyPuter
2. Removed outdated info of Puter (check lines 5-7)
3. add Self-Hostage heading
4. Update the recomended node version to 23+ instead of 22+


jelveh:
According to 3 different AI models these two are great translations! Thank you for your contribution. Merging now.

PR #1365, opened 2025-07-19
feat(gui): Add animation effects to the taskbar

GitHubliuzhijie:
- Add new animation functionality to the taskbar; icons will enlarge when the mouse hovers over them
- Introduce new CSS properties and styles to support the animation effects
- Implement mouse move and leave event handling in UITaskbar
- Add UITaskBarCreateCurve and UITask components


GitHubliuzhijie:
Is there any problem with the restore submission

PR #1369, opened 2025-07-21
fs: adapt to different mkdir api, block write to root dir, add tests

### LINKED ISSUE #1367: fs: failed to mkdir when `parent` arg is missing
# Issue

apitest throws an error:

```
> node ./tools/api-tester/apitest.js --config=./tools/api-tester/config.yml

Post body:  {"path":"/admin/api_test","overwrite":true}
Response body:  {
  message: 'Cannot create directories in the root directory.',
  code: 'forbidden'
}
```

# Cause

This is due to PR #1351 which prohibit empty `parent` arg.

# Solution

Currently, there are 2 kinds of backend `mkdir` calls in the system:

- give full path at once: `{"path":"/admin/api_test", args...}`
- give path in format of parent + path: `{"parent": "/admin", path":"api_test", args...}`

We’re planning to make `mkdir` work for both cases and merge them in the future. Backwards compatible isn’t a major concern for the backend API.

# Others

Here is a comparison of these 2 format:

- give full path at once: `{"path":"/admin/api_test", args...}`, this api is used by apitest, and is more similar to POSIX spec (https://linux.die.net/man/3/mkdir)
- give path in format of parent + path: `{"parent": "/admin", path":"api_test", args...}`, this api is used by puter-js `puter.fs.mkdir`

XiaochenCui:
- fix #1367
- adapt to different `mkdir` APIs
- add tests for `create_missing_parents` arg
- add tests for "mkdir in root directory"

test:
```
> node ./tools/api-tester/apitest.js --config=./tools/api-tester/config.yml --unit

=== Test Results ===
┌─────────────────────────────────┬────────┬────────┬───────┐
│ (index)                         │ passed │ failed │ total │
├─────────────────────────────────┼────────┼────────┼───────┤
│ Cartesian Test for /write       │ 24     │ 0      │ 24    │
│ Cartesian Test for /move        │ 44     │ 4      │ 48    │
│ Cartesian Test for /copy        │ 144    │ 0      │ 144   │
│ write and read                  │ 6      │ 0      │ 6     │
│ move                            │ 4      │ 0      │ 4     │
│ stat                            │ 133    │ 1      │ 134   │
│ readdir                         │ 115    │ 0      │ 115   │
│ mkdir                           │ 22     │ 0      │ 22    │
│ batch                           │ 4      │ 0      │ 4     │
│ delete                          │ 7      │ 0      │ 7     │
│ single write for trace and span │ 0      │ 0      │ 0     │
└─────────────────────────────────┴────────┴────────┴───────┘
```

All `mkdir` tests passed. The failed tests are unrelated to this PR.

### DIFF
--- a/src/backend/src/filesystem/hl_operations/hl_mkdir.js
+++ b/src/backend/src/filesystem/hl_operations/hl_mkdir.js
@@ -280,22 +280,7 @@ class HLMkdir extends HLFilesystemOperation {
         }
 
         let parent_node = values.parent || await fs.node(new RootNodeSelector());
-        if ( parent_node.isRoot ) {
-            // root directory is read-only
-            throw APIError.create('forbidden', null, {
-                message: 'Cannot create directories in the root directory.'
-            });
-        }
         console.log('USING PARENT', parent_node.selector.describe());
-
-        // TODO: this can be removed upon completion of: https://github.com/HeyPuter/puter/issues/1352
-        if ( parent_node.isRoot ) {
-            // root directory is read-only
-            throw APIError.create('forbidden', null, {
-                message: 'Cannot create directories in the root directory.'
-            });
-        }
-
 
         let target_basename = _path.basename(values.path);
 
@@ -305,6 +290,14 @@ class HLMkdir extends HLFilesystemOperation {
             ? await this._create_top_parent({ top_parent: parent_node })
             : await this._get_existing_top_parent({ top_parent: parent_node })
             ;
+
+        // TODO: this can be removed upon completion of: https://github.com/HeyPuter/puter/issues/1352
+        if ( top_parent.isRoot ) {
+            // root directory is read-only
+            throw APIError.create('forbidden', null, {
+                message: 'Cannot create directories in the root directory.'
+            });
+        }
 
         // `parent_node` becomes the parent of the last directory name
         // specified under `path`.
--- a/src/backend/src/filesystem/hl_operations/hl_mkdir.js
+++ b/src/backend/src/filesystem/hl_operations/hl_mkdir.js
@@ -271,8 +271,32 @@ class HLMkdir extends HLFilesystemOperation {
             });
         }
 
+        // Unify the following formats:
+        // - full path: {"path":"/foo/bar", args...}, used by apitest (./tools/api-tester/apitest.js)
+        // - parent + path: {"parent": "/foo", "path":"bar", args...}, used by puter-js (puter.fs.mkdir("/foo/bar"))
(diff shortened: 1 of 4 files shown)


XiaochenCui:
@KernelDeimos 
this PR is ready for review & merge

src/backend/src/filesystem/hl_operations/hl_mkdir.js
```diff
@@ -271,8 +271,32 @@ class HLMkdir extends HLFilesystemOperation {
...
+        // Unify the following formats:
+        // - full path: {"path":"/foo/bar", args...}, used by apitest (./tools/api-tester/apitest.js)
+        // - parent + path: {"parent": "/foo", "path":"bar", args...}, used by puter-js (puter.fs.mkdir("/foo/bar"))
+        if ( !values.parent && values.path ) {
+            values.parent = await fs.node(new NodePathSelector(_path.dirname(values.path)));
+            values.path = _path.basename(values.path);
+        }
+
         let parent_node = values.parent || await fs.node(new RootNodeSelector());
+        if ( parent_node.isRoot ) {
```
KernelDeimos: What is the reason for changing this from `top_parent` to `parent_node`? What happens differently?

The reason this was `top_parent` before is because `top_parent` was the node immediately above the new directory being created, which is the only case when we want to throw this error (i.e. you're trying to make a new directory immediately under `/`, which isn't allowed). Has this changed?

src/backend/src/filesystem/hl_operations/hl_mkdir.js
```diff
@@ -271,8 +271,32 @@ class HLMkdir extends HLFilesystemOperation {
...
+            // root directory is read-only
+            throw APIError.create('forbidden', null, {
+                message: 'Cannot create directories in the root directory.'
+            });
+        }
         console.log('USING PARENT', parent_node.selector.describe());
+
+        // TODO: this can be removed upon completion of: https://github.com/HeyPuter/puter/issues/1352
+        if ( parent_node.isRoot ) {
```
KernelDeimos: This looks identical to the block above

src/backend/src/filesystem/hl_operations/hl_mkdir.js
XiaochenCui: yeah, `top_parent` is the right choice here

src/backend/src/filesystem/hl_operations/hl_mkdir.js
XiaochenCui: I've removed the duplicate

PR #1378, opened 2025-07-28
Fix issue in invalid social media url

emmanuel-ferdman:
## PR Summary
This PR fixes the validation logic for social media url to prevent runtime issues of assigning to `const` variable `social_media_image`. Instead, this PR uses a flag `is_social_media_image_valid`.


KernelDeimos:
Hi, thanks for your contribution. Can you explain what is being fixed here? After a quick review it looks like the code is doing exactly the same thing as it was doing before but in a different way.

KernelDeimos:
Hi, thanks for the fix! I'm merging this now

Issue #542, opened 2024-07-06
Complete Norwegian Bokmål translation based on English file


jelveh:
The Norwegian Bokmål translation file (`nb.js`) needs to be updated to include all translations present in the English file (`en.js`). Currently, the Norwegian Bokmål translation may be incomplete or outdated compared to the English version.

### Steps to resolve
1. Compare the contents of `src/i18n/translations/en.js` and `src/i18n/translations/nb.js`
2. For each translation key in `en.js`, ensure a corresponding translation exists in `nb.js`
3. If a translation is missing in `nb.js`, add it with an appropriate Norwegian Bokmål translation
4. If any translations in `nb.js` are outdated or incorrect, update them to match the current English version

### Additional notes
- Please ensure that the translations are culturally appropriate and maintain the intended meaning of the English text
- If certain terms or phrases don't have a direct Norwegian Bokmål equivalent, consider adding translator notes or using transliteration where appropriate

### Files to be updated
- `src/i18n/translations/nb.js`

### Related files
- `src/i18n/translations/en.js` (reference file)

Thank you for helping to improve the Norwegian Bokmål localization of our project!

ruthwikchikoti:
@jelveh  can you please assign this ?

Shahidpathan22:
I would like to work on this

PR #1381, opened 2025-07-29
fix: issue #542 , add missing Norwegian Bokmål translations based on English file

GlennDmello17:
- Compared `nb.js` with `en.js` and added all missing translation keys
- Used DeepL and google translate to auto-translate Norwegian Bokmål phrases
- if any changes , please let me know .


jelveh:
Thank you for your contribution. 2 AIs are telling me this is a solid translation, until we have native speakers reviewing these, I'd go with it. Merging.

PR #1388, opened 2025-07-30
Allow - in worker name

ProgrammerIn-wonderland:
Sent from my iPhone


KernelDeimos:
Merges from my iPhone

PR #1392, opened 2025-07-30
fs/event: consolidate fs.written to fs.write

### LINKED ISSUE #1391: Consolidate `fs.written` and `fs.write`
`fs.written` was added to provide an event that's fired after the entire contents of a file has been written. The previously-existing `fs.write` does not wait for upload. It was realized that none of the listeners to `fs.write` interact with the file stream, so we can make `fs.write` behave the way `fs.written` currently does and remove the `fs.written` event.

This PR requires updating WorkerService: https://github.com/HeyPuter/puter/blob/85a0f71bfd66c884c8f603cc1f274d5a465a2933/src/backend/src/services/worker/WorkerService.js#L93

Workers need to be fully tested after these changes.

XiaochenCui:
fix #1391

Workers need to be fully tested before merge.

### DIFF
--- a/src/backend/src/modules/puterfs/lib/PuterFSProvider.js
+++ b/src/backend/src/modules/puterfs/lib/PuterFSProvider.js
@@ -722,10 +722,6 @@ class PuterFSProvider extends putility.AdvancedBase {
         (async () => {
             await entryOp.awaitDone();
             resourceService.free(uid);
-            svc_event.emit('fs.written.file', {
-                node,
-                context: this.context,
-            });
         })();
 
         state_upload.post_insert({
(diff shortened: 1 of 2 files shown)


XiaochenCui:
the "worker reload on file update" feature has been tested manually

src/backend/src/modules/puterfs/lib/PuterFSProvider.js
KernelDeimos, line 1: `'fs.write.file'` should be emitted after `await entryOp.awaitDone()` and also after `await svc_fileCache.invalidate(node)`. This can be done using the TeePromise class with two instances in an asynchronous IIFE.

KernelDeimos:
'fs.write.file' should be emitted after await entryOp.awaitDone() and also after await svc_fileCache.invalidate(node). This can be done using the TeePromise class with two instances in an asynchronous IIFE.

XiaochenCui:
> 'fs.write.file' should be emitted after await entryOp.awaitDone() and also after await svc_fileCache.invalidate(node).

This has been solved, I'll test the feature manually with Rafflesia

Issue #1399, opened 2025-07-31
Adding more documentation


onel:
Really interesting project.
I see parts of the code base are already well covered by reference docs but some are missing.

Would you be open to accepting some contributions regarding that?

rohanrawat09:
Hi! I'm Rohan, a JavaScript and Node js developer excited to contribute to open source. I'd love to take it up. Can you please assign it to me? 😊

jelveh:
@onel thank you for offering to help with this! That would be very welcome and appreciated.

I can assign this to both you and @rohanrawat09. You can decide how to split the work, for example, you docoument the backend and Rohan does the frontend. Any amount of documentation is greatly appreciated, even if small.

onel:
Sounds good @jelveh 

@rohanrawat09 do you prefer a specific part of the app?
I suggest we start each with some folders and see how it goes.
Do you want to start with: `src/backend/src/api`?
I can start with something like `src/putility`

rohanrawat09:
@onel I started with "src/gui" . I have already submitted a brief description of utils.js  (it haven't approved yet), which is the main logic of gui build. 
I prefer starting with frontend. What's your suggestion ?

onel:
Yeah, that sounds good to me. You do the FE, I'll look at the BE.
Glad to hear you already contributed

KernelDeimos:
@onel I maintain putility and backend. putility desperately needs documentation, especially the part implementing "features for classes" which has profound implications but is not widely understood.

Other backend concepts that need documentation include Context, Actor, the permission system internals, BroadcastService, dev console widgets... and if anyone wants more ideas I can come up with more. Feel free to reach out if you need any help.

onel:
that's good to know @KernelDeimos

happy to start with that section as it's as not as big as the whole BE and the move one to others.

jelveh:
Hey @onel  and @rohanrawat09, let me and @KernelDeimos  know if you need help with anything here.

onel:
hey @jelveh I finally got some time to focus on this and made the first PR. 
LMK if the style / content is appropriate and we can adjust.

I'll focus next on `putility`

jelveh:
Hi @onel thank you so much for your hard work on this! Please let me know if there is any part you'd like to work on as documentation is very critical right now.

Apologies for the delay in merging your PR last time, we now have put some processes in place to respond much more quickly to PRs now.

onel:
all good @jelveh. I'll love to contribute more.

[...]

PR #1433, opened 2025-08-12
Reference docs - putility

onel:
#1399

### DIFF
--- a/src/putility/src/concepts/Service.js
+++ b/src/putility/src/concepts/Service.js
@@ -80,8 +80,8 @@ class Service extends AdvancedBase {
         return ins;
     }
 
-    /** @type {Object} Implementation of the TService trait */
     static IMPLEMENTS = {
+        /** @type {Object} Implementation of the TService trait */
         [TService]: {
             /**
              * Initializes the service by running init hooks and calling _init if present


src/putility/src/bases/BasicBase.js
```diff
@@ -27,6 +36,12 @@ class BasicBase {
         return chain.reverse();
     }
 
+    /**
+     * Merges static array properties from all classes in the inheritance chain.
+     * Avoids duplicating the same array reference multiple times.
```
KernelDeimos: The statement here might be misleading. It probably **should** behave as described but right now the behaviour is more accurately described as follows:

```suggestion
     * Avoids duplicating the same array reference from contiguous members
     * of the inheritance chain (useful when using the decorator pattern with
     * multiple classes sharing a common base)
```

src/putility/src/concepts/Service.js
```diff
@@ -20,42 +20,75 @@
...
     static create ({ parameters, context }) {
         const ins = new this();
         ins._.context = context;
         ins.as(TService).construct(parameters);
         return ins;
     }
 
+    /** @type {Object} Implementation of the TService trait */
     static IMPLEMENTS = {
         [TService]: {
```
KernelDeimos: Probably here, right?

```suggestion
    static IMPLEMENTS = {
        /** @type {Object} Implementation of the TService trait */
        [TService]: {
```

src/putility/src/libs/context.js
```diff
@@ -33,6 +48,12 @@ class Context {
         }
         return this.sub(values);
     }
+    /**
+     * Creates a sub-context that inherits from the current context with additional or overridden values.
+     * Nested Context instances are recursively sub-contexted with corresponding new values.
```
KernelDeimos, line 53: This might benefit from further explanation and an example. (This comment is non-blocking for merging)

src/putility/src/concepts/Service.js
onel: hmm, you might be right, yes. it is specific to that

src/putility/src/libs/context.js
onel: for more complex components, examples do help but it's also needed to understand how it's used, which I couldn't at this point. should I try and add an example here?

KernelDeimos: it's like this:

```javascript
const context = Context.get(); // get current context

// Create a subcontext
const sub = context.sub({
  somekey: 'somevalue', // set or override some key
  nested: new Context({ anotherkey: 'anothervalue' }),
})
```

Nested contexts are rarely used if at all though, so it's not very important to document that behavior. I'm merging this PR as-is.

PR #1438, opened 2025-08-14
Refactor GeminiService for modularity, error handling, and readability

imrozkhan205:
Summary:
Refactored GeminiService to improve readability, error handling, and modularity.

Changes:
- Broke out logic in `complete()` into smaller helper methods
- Added JSDoc comments
- Introduced constants for default values
- Avoided variable shadowing
- Added basic error handling

Notes:
- No breaking changes
- Tests pass / app runs as expected

### DIFF
--- a/src/backend/src/modules/puterai/GeminiService.js
+++ b/src/backend/src/modules/puterai/GeminiService.js
@@ -5,6 +5,10 @@ const { TypedValue } = require("../../services/drivers/meta/Runtime");
 const putility = require("@heyputer/putility");
 const FunctionCalling = require("./lib/FunctionCalling");
 
+// Constants
+const DEFAULT_MODEL = 'gemini-2.0-flash';
+const DEFAULT_TEMPERATURE = 0.7;
+
 class GeminiService extends BaseService {
     async _init () {
         const svc_aiChat = this.services.get('ai-chat');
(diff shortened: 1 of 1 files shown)


KernelDeimos:
Hi, thanks for the contribution.

Is this refactor a preliminary step for another contribution? If it is I'd rather have it together with those changes in one PR so I can make better sense of it. This refactor on its own is a little confusing to me - I don't see the advantage of these changes. It seems overly-specific (why GeminiService specifically? Is refactoring implementations of the `puter-chat-completion` interface even worthwhile? Often it makes more sense to completely re-write an implementation - i.e. new version of an SDK - than to maintain it, so why worry about its maintainability?)

These are some specific things about this PR that look incorrect or out of place to me with the context that I have:
1. jsdoc comment on complete is sort of in the wrong place. Implementors of `complete` need to know the interface, and callers of `complete` need to know the interface, but nobody needs to know the signature for specific implementations like GeminiService on an individual basis - all implementations adhere to the same signature.
2. separating code into multiple functions is great when you know there's going to be re-use or it enhances readability. As far as I can tell, neither of those are true for this ~50 LOC implementation. The comment "History is separate, so the last message gets special treatment." is a lot more helpful then seeing a call to `extractLastMessageParts` (makes a future reader ask: what is "last parts" and why are we extracting it?) Guidelines will often say "10 to 20 lines for a function" but practically there are places where that rule of the thumb is not applicable, and this is definitely one of them. To understand the code you now need to jump between multiple places, and the vertical span (number of lines) of code has doubled to introduce this situation.
3. RE: your description; where does variable shadowing occur in the existing code? I don't see any places where variable shadowing was avoided.

Adding constants for default values to the top of the implementation makes sense. I will accept that change in a separate PR.

B226376513 (approved):

PR #1447, opened 2025-08-17
Added Norwegian Bokmål translations for missing keys

Shahidpathan22:
There were some keys missing in the (src/gui/src/i18n/translations/nb.js) file when compared to the (src/gui/src/i18n/translations/en.js), hence added and translated the missing terms.


jelveh:
Thank you  😇

PR #1467, opened 2025-08-26
test: fix apitest, add new unit tests and benchmarks, normalize output

XiaochenCui:
this PR updates the apitest:

- make all tests works
- add detailed docs for usage and design
- add new unit tests and benchmark
- add special output to work with the CI script

### DIFF
--- a/tools/api-tester/apitest.js
+++ b/tools/api-tester/apitest.js
@@ -9,13 +9,14 @@ const { parseArgs } = require('node:util');
 
 const args = process.argv.slice(2);
 
-let config, report, suiteName;
+let config, report, suiteName, onlycase, bench, unit, stopOnFailure, id;
 
 try {
     const parsed = parseArgs({
         options: {
             config: {
                 type: 'string',
+                default: './tools/api-tester/config.yml',
             },
             report: {
                 type: 'string',
@@ -24,6 +25,7 @@ try {
             bench: { type: 'boolean' },
             unit: { type: 'boolean' },
             suite: { type: 'string' },
+            'stop-on-failure': { type: 'boolean' },
         },
         allowPositionals: true,
     });
@@ -35,6 +37,7 @@ try {
         bench,
         unit,
         suite: suiteName,
+        'stop-on-failure': stopOnFailure,
     }, positionals: [id] } = parsed);
 
     onlycase = onlycase !== undefined ? Number.parseInt(onlycase) : undefined;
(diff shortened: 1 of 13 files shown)


tools/api-tester/apitest.js
```diff
@@ -49,6 +52,7 @@ try {
         '  --config=<path>  (required)  Path to configuration file\n' +
         '  --report=<path>  (optional)  Output file for full test results\n' +
         '  --suite=<name>   (optional)  Run only tests with matching suite name\n' +
+        '  --stop-on-failure (optional)  Stop execution on first test failure\n' +
```
KernelDeimos, line 55: missing `--onlycase` here - what does it do?

tools/api-tester/tests/mkdir.js
```diff
@@ -145,9 +146,9 @@ module.exports = {
                 try {
                     await t.mkdir_v2('a/b', 'c');
                 } catch (e) {
-                    // TODO (xiaochen): `t.mkdir('a/b/c')` throws 409, unify the
+                    // TODO (xiaochen): `t.mkdir('a/b/c')` throws 409/422, unify the
                     // behavior of these two cases.
-                    expect(e.response.status).equal(422);
+                    expect(e.response.status).oneOf([409, 422]);
```
KernelDeimos, line 151: Is this error code dependent on the filesystem implementation? (whether it's memoryfs or puterfs)

tools/api-tester/apitest.js
XiaochenCui: `--onlycase` doesn't work, I propose to remove this arg.

A bigger problem is that the current implementation doesn't support run a specific case - since case can be nested and we don't know a case's children before running it. e.g:

```js
        // we cannot skip 'test_a' when filter=test_b, since test_b might be a child of 'test_a'
        await t.case('test_a', async () => {
            
            await t.case('test_b', async () => {
                ...
            });
```

a good news is `--suite` now supported:

- Filter tests by suite name:

    ```bash
    node ./tools/api-tester/apitest.js --config=./tools/api-tester/config.yml --unit --suite=mkdir
    ```

- Filter benchmarks by name:

    ```bash
    node ./tools/api-tester/apitest.js --config=./tools/api-tester/config.yml --bench --suite=stat_intensive_1
    ```

tools/api-tester/tests/mkdir.js
XiaochenCui: The error code depends on the state left by previous tests. I left a TODO here to (plan to) make it more deterministic and meaningful.

PR #1477, opened 2025-09-03
Fix buggy arrow key navigation in start menu

bomberfish:
This commit fixes incorrect behaviour exhibited when navigating the start menu using the arrow keys. Specifically, navigating between the Recent and Recommended sections when the Recent section was not completely full would act weirdly.


jelveh:
Found a small issue. When pressing up on the first item in the recommended section focus goes to search rather than the item above the current one.

https://github.com/user-attachments/assets/9336fee0-9d86-4490-967b-1979d2071a4e

jelveh:
There is another issue now. When I move up to search, the last item i was on is still highlighted. And also when I move down again search is still focused:

https://github.com/user-attachments/assets/018cce2e-cb55-4289-9a3a-cb5df110ef04

jelveh:
Another problem. Here I'm pressing the up arrow only, you can see the issue:


https://github.com/user-attachments/assets/ea7a96e9-cddd-4ae9-a66e-c6161b9b4fc0

jelveh:
Works as expected, I couldn't find any more issues. Merging 🎉

PR #1481, opened 2025-09-04
fix(webdav): properly escape URLs with encodeURI for RFC compliance

jaiswalsachin49:
### Problem
URLs in WebDAV PROPFIND responses were not properly escaped with `encodeURI`, causing compatibility issues with WebDAV clients like Cyberduck. This breaks RFC compliance for WebDAV implementations.

### Root Cause
While most dynamic URLs throughout the WebDAV implementation were correctly escaped, several hardcoded URLs in static response functions were missing proper URI encoding:

- `createStaticDavRootResponse()` - hardcoded `/dav/` and `/dav/admin/` paths
- `createRootWebDAVResponse()` - hardcoded `/` and `/dav/` paths  
- Some dynamic paths in PROPPATCH and LOCK responses

### Solution
Applied consistent `encodeURI` encoding to all URL paths in WebDAV responses:

```javascript
// Before (non-compliant)
<D:href>/dav/</D:href>
<D:href>/dav/admin/</D:href>

// After (RFC compliant)  
<D:href>/dav${escapeXml(encodeURI('/'))}</D:href>
<D:href>/dav${escapeXml(encodeURI('/admin/'))}</D:href>

### DIFF
--- a/src/backend/src/services/WebDavFS.js
+++ b/src/backend/src/services/WebDavFS.js
@@ -58,7 +58,7 @@ function convertToWebDAVPropfindXML(fsEntry) {
   const xml = `<?xml version="1.0" encoding="utf-8"?>
 <D:multistatus xmlns:D="DAV:">
   <D:response>
-    <D:href>/dav${escapeXml(href)}</D:href>
+    <D:href>/dav${escapeXml(encodeURI(href))}</D:href>
     <D:propstat>
       <D:prop>
         <D:displayname>${escapeXml(fsEntry.name)}</D:displayname>
@@ -113,7 +113,7 @@ function convertMultipleToWebDAVPropfindXML(selfStat, fsEntries) {
     }
 
     return `  <D:response>
-    <D:href>/dav${escapeXml(href)}</D:href>
+    <D:href>/dav${escapeXml(encodeURI(href))}</D:href>
     <D:propstat>
       <D:prop>
         <D:displayname>${escapeXml(fsEntry.name)}</D:displayname>
@@ -198,7 +198,7 @@ function createStaticDavRootResponse() {
   return `<?xml version="1.0" encoding="utf-8"?>
 <D:multistatus xmlns:D="DAV:">
   <D:response>
-    <D:href>/dav/</D:href>
+    <D:href>/dav${escapeXml(encodeURI('/'))}</D:href>
     <D:propstat>
       <D:prop>
         <D:displayname>dav</D:displayname>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/backend/src/services/WebDavFS.js [hunk 1]
```diff
@@ -198,7 +198,7 @@ function createStaticDavRootResponse() {
   return `<?xml version="1.0" encoding="utf-8"?>
 <D:multistatus xmlns:D="DAV:">
   <D:response>
-    <D:href>/dav/</D:href>
+    <D:href>/dav${escapeXml(encodeURI('/'))}</D:href>
```
ProgrammerIn-wonderland: I don't think we need to uriencode this hardcoded string

src/backend/src/services/WebDavFS.js
```diff
@@ -223,7 +223,7 @@ function createStaticDavRootResponse() {
     </D:propstat>
   </D:response>
   <D:response>
-    <D:href>/dav/admin/</D:href>
+    <D:href>/dav${escapeXml(encodeURI('/admin/'))}</D:href>
```
ProgrammerIn-wonderland: Also static

src/backend/src/services/WebDavFS.js
```diff
@@ -279,7 +279,7 @@ function createRootWebDAVResponse() {
     </D:propstat>
   </D:response>
   <D:response>
-    <D:href>/dav/</D:href>
+    <D:href>/dav${escapeXml(encodeURI('/'))}</D:href>
```
ProgrammerIn-wonderland: Also static

ProgrammerIn-wonderland:
PR fixes the issue but it also unnecessarily escapes already safe and fully static strings

src/backend/src/services/WebDavFS.js
jaiswalsachin49: Thanks for the feedback! You're absolutely right - for these static hardcoded strings, encodeURI() is redundant since:

•  encodeURI('/') returns '/' 
•  encodeURI('/admin/') returns '/admin/'

I've updated the code to remove the unnecessary encodeURI() calls for these static paths while keeping them for dynamic href and filePath variables where they're actually needed for proper URL encoding.

The changes maintain WebDAV compliance (dynamic paths still get encoded) while eliminating the redundant function calls for static strings.

ProgrammerIn-wonderland:
While you did remove the encodeURI on the statics you kept the escapeXML which you also added

ProgrammerIn-wonderland:
Fixed with 35c66f7d42441bf5d2449dc34c505ce0023f3a4c

PR #1487, opened 2025-09-06
Add context menu to apps in the start menu

bomberfish:


jelveh:
If I right-click on two items consecutively this happens:

Screenshot 2025-09-06 at 10 25 41 AM

jelveh:
ahhh! I just realized this was a draft PR. I'll wait for the final version

jelveh:
One issue I noticed; when the user right click on a launch menu item, the "active state" is removed from the item i.e. it's not highlighted any longer.

Mouse over item (highlighted):
Screenshot 2025-09-08 at 6 14 11 PM

Right-click (highlight/active state removed):
Screenshot 2025-09-08 at 6 14 51 PM

This is confusing to the user as once the menu is opened, the connection between menu and item is basically removed.

jelveh:
When the context menu is opened the active state is now correctly preserved, but when the context menu is closed the active state isn't removed.

jelveh:
if you open a context menu anywhere else on the desktop and then open the new launch menu context menu the shadow behind it is much more pronounced

jelveh:
There is some extra space that shows up to the right and bottom of the launcher, randomly. I can't reproduce it consistently right now but it does show up.
Screenshot 2025-09-12 at 6 04 07 PM

ProgrammerIn-wonderland:
image

Just to add it to a todo, clicking on white space shows the browser native context menu, I think this should be supressed

PR #1543, opened 2025-09-12
Updating the Arabic (العربية) translation 

shadow-taw:
### Updating the Arabic (العربية) translation 
Previously committed Arabic translation appeared reversed or improperly aligned due to **RTL** (Right-To-Left) formatting issues. This update improves the layout and visual structure for proper display in **RTL** languages.

2025-09-11_16-12(better)

- In Arabic, text flows from right to left, so punctuation like periods and commas should appear at the left end of the sentence — the opposite of English. Placing them on the right looks incorrect and breaks proper **RTL** formatting.

- The images are from the live Demo so you can use them as a reference.(_The images on the right are the valid ones_)

- ### PS : this line is not necessary for Arabic


jelveh:
Great! AI is telling me this is a huge improvement. Also, RTL is on the roadmap! So things would improve a lot soon

PR #1559, opened 2025-09-14
fix: add Chinese transaltions

Webb321:
### Closes https://github.com/HeyPuter/puter/issues/1502
This PR updates the Chinese translation in this [file](https://github.com/HeyPuter/puter/blob/main/src/gui/src/i18n/translations/zh.js). I added translations for all previously missing keys (replacing undefined values) with accurate equivalents.


jelveh:
Thank you for your contribution! Merging :)

Issue #1563, opened 2025-09-15
remove suggested apps from `/readdir`


KernelDeimos:
Suggested apps currently accounts for about 25% of the time cost of readdir

PR #1596, opened 2025-09-20
dev: remove suggested apps from readdir

mac146:
This PR removes the call to `fetchSuggestedApps` in `/readdir` to improve performance.

Issue: #1563

- `/readdir` now only fetches subdomains.
- Suggested apps are no longer loaded, reducing time cost (~25% of readdir).
- Functionality of directory listing remains unchanged.

### DIFF
--- a/src/backend/src/filesystem/hl_operations/hl_readdir.js
+++ b/src/backend/src/filesystem/hl_operations/hl_readdir.js
@@ -95,7 +95,7 @@ class HLReadDir extends HLFilesystemOperation {
 
             if ( ! no_assocs ) {
                 await Promise.all([
-                    child.fetchSuggestedApps(user),
+                   // child.fetchSuggestedApps(user),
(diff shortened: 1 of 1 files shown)


usv240:
Hey,
I'll try to work on this, can you assign it to me?

ProgrammerIn-wonderland:
I think this is a breaking change since apps may depend on this behavior

HOWEVER, I think it is likely a good idea to make this a flag and have a `puter.fs.readdir(path, {lessData: true});` 
cc: @jelveh 

Edit: upon further review this code is already gated behind no_assocs, but this flag doesn't seem to be exposed in puter.js

jelveh:
Hi there,
The issue opened was unfortunately not fully explained. This is not as simple as removing suggested_apps and it will break existing processes. There is a bit of a refactor involved. I'm going to close the PR and assign the issue to myself because there is a large amount of context missing.

Issue #1424, opened 2025-08-07
[quirk] UIAlert stringifies value


KernelDeimos:
The following code would not work as intended:

```javascript
const answer = await UIAlert({
    message: `Press "yes" to perform dangerous action!`,
    buttons: [
        {
            label: i18n('yes'),
            value: true,
        },
        {
            label: i18n('no'),
            value: false,
        },
    ]
});
if ( answer ) do_dangerous_action();
```

Regardless of which button the user presses, `do_dangerous_action()` is called. This is because `value` is stringified unnecessarily. Instead of the value `false`, we get the value `"false"`, which is truthy.

jelveh:
I couldn't reproduce this. Pasted this exact code in the desktop, changed it worked as expected.

dtakahashi333:
@jelveh I would like to work on this issue. Can you please assign it to me?

jelveh:
Yesd @dtakahashi333! go for it :)

dtakahashi333:
@KernelDeimos I have a potential patch for this issue. Could you please advise me on how to test my code?

jelveh:
Could you open a pr? I'll take a look :)

dtakahashi333:
Sure I will :)

dtakahashi333:
@jelveh I have a question related to this issue. I noticed that the `html_encode` function currently returns `an empty string` for the boolean value `false`, whereas it returns `'true'` for the boolean `true`. I'm wondering if this is the expected behavior.

Do you know where the `html_encode` function is defined?

jelveh:
We use [html-entities](https://github.com/mdevils/html-entities), so I'd say the behaviour you see in `html_encode` is a quirk of the html-entities module. Is this causing issues?

dtakahashi333:
Thank you for the clarification!

That also explains why you weren’t able to reproduce the issue mentioned by @KernelDeimos, as `answer` becomes an empty string, which effectively bypasses the condition that triggers `do_dangerous_action()`.

Since `jQuery` [.attr()](https://api.jquery.com/attr/) function always returns attribute values as strings, it's safer to recover the original value types after retrieving them from HTML elements when clicking buttons. This would help avoid the potential issue @KernelDeimos mentioned, by ensuring the values behave as expected in conditional logic.

PR #1597, opened 2025-09-21
fix: UIAlert stringifies value (#1424)

dtakahashi333:
This patch fixes a potential issue in `UIAlert` where values were being **stringified** unexpectedly, by introducing the `parseDataValue()` function, which attempts to recover the original JavaScript value type.

### DIFF
--- a/src/gui/src/UI/UIAlert.js
+++ b/src/gui/src/UI/UIAlert.js
@@ -160,11 +160,49 @@ function UIAlert(options) {
         $(el_window).find('.alert-resp-button').on('click', async function (event) {
             event.preventDefault();
             event.stopPropagation();
-            resolve($(this).attr('data-value'));
+            // Recover the original data-value type before resolving.
+            resolve(parseDataValue($(this).attr('data-value')));
             $(el_window).close();
             return false;
         })
     })
+}
+
+/**
+ * Parses a string retrieved from an HTML data attribute or similar source,
+ * attempting to recover its original JavaScript value type.
+ *
+ * This function handles common conversions such as:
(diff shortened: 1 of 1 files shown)


jelveh:
Thanks @dtakahashi333 for this! At first glance, this makes sense; but I need to do some deep testing as a lot of devs depend on `UIAlert`. If you're testing right now as well, do let me know if something comes up, I'd like to help/discuss it.

dtakahashi333:
@jelveh I understand your concern, and I totally agree with you. Considering the impact of this update, I believe it's fine to leave it as is unless a user reports it as an issue.

Based on my inspection, as long as [html_encode()](https://github.com/mdevils/html-entities) behaves as expected, which returns an `empty string` for `false`, I believe the issue @KernelDeimos mentioned will not occur.

Do you have any thoughts on this?

jelveh:
This makes sense. The old code (causing the quirk) is not very elegant though and I wish we could just replace it. Let me do more testing and I'll get back to you

dtakahashi333:
@jelveh Do you happen to know if the **Puter** project is planning to use any GUI test automation tools, like [Playwright](https://playwright.dev/), [Cypress](https://www.cypress.io/), or similar?

jelveh:
This is something we'd love to have! Would you be interested in taking a stab at it?

dtakahashi333:
@jelveh Sure, I’d be happy to!

jelveh:
Awesome! Could you either open an issue or email me at nj@puter.com so that I could help you get started on it? Let me know what you will need and how we could plan this. I'd be happy to help in any way I can.

PR #1602, opened 2025-09-22
feat: add hello-world example page

ZayanLambe13:


ZayanLambe13:
This PR updates the host OS detection logic to correctly identify desktop-like iPads (iPads with mouse/keyboard support) in addition to macOS, Windows, and other platforms. This ensures better compatibility and accurate UI behavior for users accessing Puter from such devices.

Added isDesktopIPad() function to check for iPads with fine pointer devices.

Updated window.detectHostOS to return "desktop" for desktop-like iPads.

Preserves existing detection for macOS, Windows, and other platforms.

This improves platform detection for GUI and other environment-specific functionality.

PR #1611, opened 2025-09-24
client-cache: update cache-invalidation policy to ts-based comparison

XiaochenCui:
This PR replace the cache-invalidation policy from "invalidate on any update" to "invalidate when client-cache has an older timestamp".

It's ready for review/merge and there are some work to be done in the following PRs:
- persistent ts on server
- persistent cache on client

### DIFF
--- a/src/backend/src/routers/filesystem_api/cache.js
+++ b/src/backend/src/routers/filesystem_api/cache.js
@@ -1,0 +1,39 @@
+/*
+ * Copyright (C) 2024-present Puter Technologies Inc.
+ *
+ * This file is part of Puter.
+ *
+ * Puter is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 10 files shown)


XiaochenCui:
@jelveh this PR is ready for review/merge

XiaochenCui:
this PR doesn't work with https://github.com/HeyPuter/kv.js/pull/19
I'll reopen it once the "persistent on server/client" is finished

More specifically, "kv.js persistent" feature enable a new session to reuse the cache left by previous sessions. Once the cache is persistent, the timestamp used for invalidation should be persistent as well.

XiaochenCui:
@KernelDeimos @jelveh this PR is ready for review, now the `last_updated_ts` is persistent on both server and client. (client persistent is done by https://github.com/HeyPuter/kv.js/pull/19)

There are some details to pay attention in this PR:
- I created a new file `src/backend/src/routers/filesystem_api/cache.js` which exposes `/cache/last-change-timestamp` http api. Does this meet our system standard? (It looks like a temporary file/api)
- Both pull and push are used in timestamp update, should we use just one approach?
- Should we use websocket instead of http for "get the last-change-timestamp" api?

PR #1614, opened 2025-09-25
feat: add missing Persian and Finnish translations

mrtbgau:
#1529


mrtbgau:
Can t do a PR per commit and my last commit was added to this PR

PR #1620, opened 2025-09-25
client-cache: remove pulling

XiaochenCui:
This is a hot-fix that remove the DDOS caused by http pulling. It's tested and ready for review/merge.

There are some issues remained that related to client-cache:

- [x] We set the cache no matter the pushing websocket is connected or not, this will cause stale cache when the ws connection is lost. (fixed: it's fine, "websocket lost and failed to reconnect" means the server is unreachable, cache validity isn’t a concern in this case.)
- [x] Seems puter.fs.socket failed to listen on 'cache.updated' event after reconnected. (How to reproduce: kill it via `puter.fs.socket.disconnect();`) (fixed: use `puter.fs.initializeSocket();` instead)
- [x] Remote ts is undefined. (client log: `remote timestamp (undefined) is newer than local timestamp (undefined), flushing cache`) (fixed in this PR)
- [x] `getCacheTimestamp` is unused. (fixed: this is fine, we need this api on session start to validate the local cache)

### DIFF
--- a/src/puter-js/src/modules/FileSystem/index.js
+++ b/src/puter-js/src/modules/FileSystem/index.js
@@ -112,8 +112,8 @@ export class PuterJSFileSystemModule extends AdvancedBase {
     bindSocketEvents() {
         this.socket.on('cache.updated', (item) => {
             const local_ts = puter._cache.get(LAST_UPDATED_TS);
-            if (item.ts > local_ts || local_ts === undefined) {
-                console.log(`remote timestamp (${item.ts}) is newer than local timestamp (${local_ts}), flushing cache`);
+            if (item.timestamp > local_ts || local_ts === undefined) {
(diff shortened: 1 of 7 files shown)


XiaochenCui:
@jelveh this PR is ready for merge, I'll address the remaining issues later

XiaochenCui:
@jelveh the "stale cache" bug has been solved, now it's ready for review/merge

XiaochenCui:
still WIP, will be ready later

XiaochenCui:
@jelveh now it's ready for merge, the new logic will be in the next PR:

`client.last_valid_ts`
Definition: the last valid time of local cache.
Update logic:
- update every second
- reset to 0 on "local fs update", after the purge action
- reset to 0 on "remote fs update", after the purge action

PR #1635, opened 2025-10-01
Add 'Set as Desktop Background' context menu item to images

### LINKED ISSUE #1608: Enable "set as background" for photos
When right-clicked, the user would be able to set a photo as the desktop background

bomberfish:
Fixes #1608 
Screenshot 2025-09-30 at 20 55 34


KernelDeimos:
Thought you might be interested to know about a regression this PR caused:
https://github.com/HeyPuter/puter/pull/1917

I can definitely imagine how files without extensions were missed during testing - that's not something that comes up very often. In the future though this is probably a good thing to test.

PR #1638, opened 2025-10-01
Update Indonesian (Bahasa Indonesia) translation of Puter

skeith:
Added missing translations and re-translated some of existing translations. 
Should solve #1522

_Trivia: Puter means spin in Indonesia (casual form)_


jelveh:
Merging 🎉

PR #1639, opened 2025-10-01
Complete Igbo translation for missing keys (#1521)

pranav200408:
This PR completes all missing Igbo translations in the file.
src/gui/src/i18n/translations/ig.js as described in issue #1521.
No more undefined or missing translations exist in this file.


jelveh:
Hi there. This is not a translation PR, seems to be another PR achieving something else.

PR #1646, opened 2025-10-02
Add Puter.js docs

reynaldichernando:


reynaldichernando:
Some things we might want to do @jelveh 
- make sure this directory license doesn't mess with the main repo license
- do we want to remove V1 docs?
- other hidden pages to be removed if any

jelveh:
For reference, we ended up releasing docs under another repo: https://github.com/HeyPuter/docs

PR #1687, opened 2025-10-05
Improved public folder file links

### LINKED ISSUE #1633: Linking to files in public folders
Currently linking to files in public folders doesn't work very well.

The idea behind linking to Public folders is like this:
- Create/copy/upload some file to your "Public" folder
- Navigate to `<origin>/@your-username/<filename>`
- The file opens in the default app associated with that filetype
- The window in GUI has metadata so that when it's in focus the URL is `<origin>/@your-username/<filename>` instead of `<origin>/app/<app-name>`

Currently some file types work while others don't, and the URL is immediately replaced with the app URL.

In the production `puter.com` environment we can see these issues:
- Go to `https://puter.com/@ed/test.txt` to see that my `test.txt` files **does** work
- Go to `https://puter.com/@ed/README.md` to see that my `README.md` file does **not** work

bomberfish:
Fixes #1633. Also added is an error dialog when the requested file does not exist. Currently blocked by a (pre-existing) bug where UIDesktop would not finish initializing when using a direct file link

### DIFF
--- a/src/gui/src/UI/UIDesktop.js
+++ b/src/gui/src/UI/UIDesktop.js
@@ -44,7 +44,6 @@ import item_icon from "../helpers/item_icon.js"
 import UIWindowSearch from "./UIWindowSearch.js"
 
 async function UIDesktop(options) {
-    console.log('UIDesktop()');
     // start a transaction if we're not in embedded or fullpage mode
     let transaction;
     if (!window.is_embedded && !window.is_fullpage_mode) {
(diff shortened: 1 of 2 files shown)


jelveh:
I'm testing this right now, but firs thing I noticed that needs to be fixed is the extra console.logs that were put in for debugging

jelveh:
need to use i18n function instead of hardcoding string literals. e.g. `'User or path not found.'` -> `i18n('User or path not found.')`

PR #1692, opened 2025-10-06
the first release of client replica

XiaochenCui:
WIP


XiaochenCui:
@KernelDeimos @jelveh this PR is ready for review/merge

src/backend/src/filesystem/hl_operations/hl_mkdir.js
KernelDeimos, line 1: (1) Why is this file changed? There do not appear to be any functional changes here - I think these changes were included by accident.

src/backend/src/filesystem/hl_operations/hl_mklink.js
KernelDeimos, line 1: See previous comment "(1)"

src/backend/src/modules/puterfs/lib/PuterFSProvider.js
```diff
@@ -220,6 +220,21 @@ class PuterFSProvider extends putility.AdvancedBase {
...
+
+        // ================== client-replica patch start ==================
+        // patch: the parent_uid should be updated after move operation
+        {
+            const new_parent_uid = await new_parent.get('uid');
+            node.entry.parent_uid = new_parent_uid;
+
+            const db = services.get('database').get(DB_WRITE, 'filesystem');
+            await db.write(
+                'UPDATE fsentries SET parent_uid = ? WHERE uuid = ?',
```
KernelDeimos, line 241: It's a bit concerning to see this change. If the `parent_uid` was not already being updated, that would mean `move` is currently broken. However, I just tried moving a file in prod - move is not broken.

A second area of concern is I was informed that the relica code won't do anything if it's not enabled, however this looks like a database operation that's happening every time a file is moved without any feature flag being checked.

src/backend/src/filesystem/hl_operations/hl_write.js
```diff
@@ -431,6 +432,22 @@ class HLWrite extends HLFilesystemOperation {
...
+        // ================== client-replica hook start ==================
+        // "write" hook
+        (async () => {
+            try {
+                const user_id = values?.user?.id;
+                if ( ! user_id ) {
+                    console.error('user_id is missing');
+                    return;
+                }
+                await sendFSNew(user_id, response);
```
KernelDeimos, line 443: It isn't necessary to add these hooks into the filesystem implementation provider. This puts the responsibility on file system providers to inform the client-replica system of what's happening, requiring duplication of effort any time a new provider is added. This sort of thing is why we have filesystem events like `fs.create.*`, for example [here](https://github.com/HeyPuter/puter/blob/206633293c5300138e287c83138d926b78ad5de4/src/backend/src/modules/puterfs/lib/PuterFSProvider.js#L696). If the events don't have enough information or some events are missing (it looks like we don't have an event for delete yet) then the events should be updated to support that.

src/backend/src/modules/web/lib/eggspress.js
```diff
@@ -187,9 +187,6 @@ module.exports = function eggspress (route, settings, handler) {
         return next();
       }
     }
-    if ( config.env === 'dev' && process.env.DEBUG ) {
-      console.log(`request url: ${req.url}, body: ${JSON.stringify(req.body)}`);
-    }
```
KernelDeimos, line 192: Why is this conditional debug log being removed? This only prints of the `DEBUG` environment variable is set - it allows easier debugging of endpoints when Puter is initiated with a terminal input like: `DEBUG=1 npm start`

src/backend/src/routers/filesystem_api/rename.js
```diff
@@ -176,21 +173,49 @@ module.exports = eggspress('/rename', {
...
+            const errors = req.services.get('error-service').create(log);
+            errors.report('emit.rename', {
+                alarm: true,
+                source: e,
+            });
+        }
+    })();
+
+    // ================== client-replica hook start ==================
+    // "rename" hook
```
KernelDeimos, line 193: If we can use events instead of this hook here that would be better.

src/fs_tree_manager/go/fs_tree_manager_grpc.pb.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+func (c *fSTreeManagerClient) FetchReplica(ctx context.Context, in *FetchReplicaRequest, opts ...grpc.CallOption) (*MerkleTree, error) {
+	cOpts := append([]grpc.CallOption{grpc.StaticMethod()}, opts...)
+	out := new(MerkleTree)
+	err := c.cc.Invoke(ctx, FSTreeManager_FetchReplica_FullMethodName, in, out, cOpts...)
+	if err != nil {
+		return nil, err
+	}
+	return out, nil
```
KernelDeimos, line 76: I'm nitpicking here but why not just `return out, nil` since `out` is already a nil-able `*MerkelTree` and `err` will have the appropriate value?
```suggestion
	return out, err
```

KernelDeimos (changes requested):
Biggest concerns are:
- filesystem events should be used instead of adding hooks everywhere
- database `UPDATE` added to `move` operation; I don't understand what it's fixing

Go code looks good to me, but I reviewed it very quickly since it's more outside the concern of the core backend.

src/backend/src/modules/puterfs/lib/PuterFSProvider.js
XiaochenCui: it's not broken but the `parent_uid` is wrong without this patch, see here for details:

https://github.com/XiaochenCui/puter/blob/fs/client-replica/doc/RFCS/20250821_client_replica_file_system.md#fsentry-parent

the code below this patch doesn't set the `parent_uid` correctly:
https://github.com/HeyPuter/puter/blob/847b3a07a4ec59e724063f460a4c26cb62b04d42/src/backend/src/modules/puterfs/lib/PuterFSProvider.js#L223-L232

src/backend/src/modules/web/lib/eggspress.js
XiaochenCui: oh, this log was added by me. I don't know others are using it - will revert this

src/fs_tree_manager/go/fs_tree_manager_grpc.pb.go
XiaochenCui: `src/fs_tree_manager/go` and `src/fs_tree_manager/js` are generated by protobuf, no need to review

src/backend/src/modules/web/lib/eggspress.js
KernelDeimos: Honestly I thought I added it, but let's keep it - it seems useful

src/backend/src/modules/puterfs/lib/PuterFSProvider.js
XiaochenCui: I found `parent_uid` is correct after moving, I'll look into this later

PR #1695, opened 2025-10-07
fix: remove verbose parameters log

KernelDeimos:


KernelDeimos:
apitest is currently broken. I tested a few things manually (including the start menu in light of recent events) and nothing is broken as far as I can tell. This is an important change so I'm merging it despite not being able to run apitest here (which shouldn't be applicable anyway)

PR #1696, opened 2025-10-07
more reliable suspended

KernelDeimos:


KernelDeimos:
apitest is broken. We can merge this when deemed necessary or when apitest is working again.

XiaochenCui:
apitest now works

jelveh:
I'm merging this now because we need it and it's passing tests

PR #1697, opened 2025-10-07
fix: backward compatibility for puter kv incr/decr

Salazareo:


src/puter-js/index.d.ts
```diff
@@ -297,8 +297,8 @@ interface KV {
     set(key: string, value: string | number | boolean | object | any[]): Promise<boolean>;
     get(key: string): Promise<any>;
     del(key: string): Promise<boolean>;
-    incr(key: string, pathAndAmount: { [key: string]: number }): Promise<number>;
-    decr(key: string, pathAndAmount: { [key: string]: number }): Promise<number>;
+    incr(key: string, pathAndAmount?: { [key: string]: number } | number): Promise<number>;
+    decr(key: string, pathAndAmount?: { [key: string]: number } | number): Promise<number>;
```
reynaldichernando: question: should we do multi types or method overloading with different types?

tests/puterJsApiTests/kv.test.ts
```diff
@@ -18,24 +18,67 @@ describe('Puter KV Module', () => {
     });
 
     it('should increment a key success', async () => {
-        const getRes = await puter.kv.get(TEST_KEY);
-        expect(getRes).toBe(0);
+        await puter.kv.set(TEST_KEY, 0);
         const incrRes = await puter.kv.incr(TEST_KEY, { '': 5 });
```
reynaldichernando, line 22: for 'should increment a key success' do we intend to have the argument? since other test below already include the argument
i wonder if this test is for without the argument

tests/puterJsApiTests/kv.test.ts
```diff
@@ -18,24 +18,67 @@ describe('Puter KV Module', () => {
...
         expect(incrRes).toBe(5);
         const finalGet = await puter.kv.get(TEST_KEY);
         expect(finalGet).toBe(5);
     });
 
     it('should decrement a key success', async () => {
-        const getRes = await puter.kv.get(TEST_KEY);
-        expect(getRes).toBe(5);
+        const getRes = await puter.kv.set(TEST_KEY, 0);
         const decrRes = await puter.kv.decr(TEST_KEY, { '': 3 });
```
reynaldichernando, line 30: same as above, i wonder if this test is for the one without pathAndAmount param

tests/puterJsApiTests/kv.test.ts
```diff
@@ -18,24 +18,67 @@ describe('Puter KV Module', () => {
...
-        expect(decrRes).toBe(2);
+        expect(decrRes).toBe(-3);
         const finalGet = await puter.kv.get(TEST_KEY);
-        expect(finalGet).toBe(2);
+        expect(finalGet).toBe(-3);
     });
+
+    it('should increment a key with second argument', async () => {
+        await puter.kv.set(TEST_KEY, 0);
+        const incrRes = await puter.kv.incr(TEST_KEY);
```
reynaldichernando, line 38: i think this test is missing the second argument

tests/puterJsApiTests/kv.test.ts
```diff
@@ -18,24 +18,67 @@ describe('Puter KV Module', () => {
...
+        expect(incrRes).toBe(1);
+        const finalGet = await puter.kv.get(TEST_KEY);
+        expect(finalGet).toBe(1);
+    })
+
+    it('should increment a key with second argument', async () => {
+        await puter.kv.set(TEST_KEY, 0);
+        const incrRes = await puter.kv.decr(TEST_KEY);
```
reynaldichernando, line 46: i think this test is missing the second argument

src/puter-js/index.d.ts
reynaldichernando: btw this is just me wanting to know more about the decision 😓 
i don't know or have preference for either

reynaldichernando:
other than the test cases, i think this is good to go 🚀

src/puter-js/index.d.ts
Salazareo: yeah can update

tests/puterJsApiTests/kv.test.ts
Salazareo: yes will rename

PR #1708, opened 2025-10-08
Create A New Option to Install Puter as  Linux Desktop Environment

canulasunal:
This contribution sets up a Tkinter window to install Puter as a desktop environment based on sway for various Debian and RPM based Linux Distributions for a more integrated experience.


jelveh:
Very cool idea.
I've assigned this to @ProgrammerIn-wonderland for review since he's pretty knowledgeable in this area. ✌

ProgrammerIn-wonderland:
Not against this, I think it's quite cool. I wonder if it's a better idea to use an electron window since you can install electron from npm. Thoughts?

canulasunal:
Thank you for the feedback. I also think that using something like electron is more convenient because of npm integration and general polish. Would it make sense to merge this a temporary solution and then switch to electron or make a prototype using electron first and then merge? If we decide to go with the electron installer, I could write the backend as have little experience with front end. We could also package the Python version using something like pyinstaller so we can make the installer a separate package.

PR #1712, opened 2025-10-08
devex: webpack via module instead of subprocess

KernelDeimos:
This was really really really (100 more "really"s) tedious compared to what I had expected.

rollup will be next.


KernelDeimos:
PR #1723 include these changes

PR #1715, opened 2025-10-09
fix: quick fix for config context issue

KernelDeimos:
I don't know what causes this


XiaochenCui:
It doesn't fix the issue on my env. I suggest close this PR

KernelDeimos:
Fixed by #1730

PR #1783, opened 2025-10-20
Add `dedupeName` to `CreateAppOptions`

### LINKED ISSUE #1777: Add `dedupeName` param for `puter.apps.create`
puter.apps.create supports dedupe_name field (via drivers call), might be a good idea to add this to puter js client lib

reynaldichernando:
Fixes #1777

### DIFF
--- a/src/puter-js/index.d.ts
+++ b/src/puter-js/index.d.ts
@@ -146,6 +146,7 @@ interface CreateAppOptions {
     icon?: string;
     maximizeOnStart?: boolean;
     filetypeAssociations?: string[];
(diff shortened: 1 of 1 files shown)


reynaldichernando:
hi @Salazareo , for puter js change, do i need to do build or something before it can be published to npm?

utkubakir:
```
WORKERS ERROR: Preamble has not been built! Workers will not have access to puter.js
To fix this cd into src/backend/src/worker and run npm run build
```

^ from the worker log

Salazareo (approved):

PR #1839, opened 2025-10-27
Update files.js and deleteUser to use ll_read

KernelDeimos:


KernelDeimos:
GitHub is erroneously showing repeated changes from the last PR. I need to investigate why this is happening.

**update:** that was worth investigating; this branch had the wrong base

PR #1860, opened 2025-10-29
ci: expose server logs

XiaochenCui:


XiaochenCui:
this pr is ready for merge

it adds the runtime log to artifacts so people can check logs when ci failed

PR #1922, opened 2025-11-06
Add Google OAuth login support

jelveh:
Introduces Google OAuth authentication by adding backend routes, configuration defaults, and frontend UI integration. Enables users to log in or sign up using their Google account if enabled in config, with domain restrictions and session management.


Salazareo:
nit: should probably be an extension

PR #1940, opened 2025-11-07
[puterfs] move write operations to extension

KernelDeimos:


KernelDeimos:
I'm going to merge this Tuesday morning in case any issues come up. I have another branch ready that depends on this but I want these changes to go into prod first, because the migration is now entering the parts where the risk of breaking changes is relatively high.

PR #1941, opened 2025-11-07
fix: bug caused by rollup release

KernelDeimos:
I was going to wait for 4.53.1 but got impatient - this will fix it for now.


KernelDeimos:
This was fixed by instead removing rollup entirely from the Puter monorepo. Rollup can't hurt us anymore.

Issue #1764, opened 2025-10-16
Missing published/attached website icon for folder on different views


reynaldichernando:
We have 3 views for file explorer, I'm not sure what they are called, so i'm naming them
- large
- small
- detail

we only have the website icon on large

<img width="1360" height="356" alt="Image" src="https://github.com/user-attachments/assets/d1e39b56-c1ae-4bea-bb44-e123697c22fb" />

but not on small or detail

<img width="1363" height="771" alt="Image" src="https://github.com/user-attachments/assets/df4ea6d2-1e68-4163-a106-68ee1cad79a1" />

<img width="1354" height="784" alt="Image" src="https://github.com/user-attachments/assets/cf34016c-20d2-483a-8a9e-a43b2525363d" />

reynaldichernando:
can i get assigned for this?

jelveh:
Go for it! assigned.

Hetsavani:
can i get assigned for this?

reynaldichernando:
hi @Hetsavani, go ahead if you want to give this a shot, my progress so far was gathering context where this website ui element is located

i found it here
https://github.com/HeyPuter/puter/blob/4b14ef863c436eaedcdbf14c2cf98a47ce573157/src/gui/src/UI/UIItem.js#L127-L138

i haven't continued to look where the file explorer views yet, like the screenshot above

i think it should just be a matter of adding the ui element to the other views

Hetsavani:
Thank you!
I made the same progress. 😄

PR #1945, opened 2025-11-08
Fix: missing published/attached website icon for folder on different views (#1764)

Hetsavani:
Fix : #1764 

Screenshot from 2025-11-08 12-51-29

Screenshot from 2025-11-08 12-51-45

### DIFF
--- a/src/gui/src/css/style.css
+++ b/src/gui/src/css/style.css
@@ -614,12 +614,27 @@ span.header-sort-icon img {
 }
 
 .item-container-list .item-badges {
-    display: none;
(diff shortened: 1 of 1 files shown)


jelveh:
Looks pretty great! We probably need to make the badges a bit larger at some point but this will do for now! 
Thanks again, merging now.

Hetsavani:
> Looks pretty great! We probably need to make the badges a bit larger at some point but this will do for now! Thanks again, merging now.

Thanks for reviewing and merging! I appreciate the feedback.

PR #1957, opened 2025-11-11
Add Mistral OCR support and imrpove `img2txt` API

jelveh:
Introduces Mistral OCR integration in backend and updates the `AIInterfaceService` and `MistralAIService` to support new OCR options and response normalization. Updates metering cost map for OCR and annotation usage. Refactors `DriverService` to support interface-specific service aliases. Expands the puter.js `AI.img2txt` API to support flexible options and provider selection, including Mistral OCR.


jelveh:
passes all existing puter.js tests. I'm merging now and testing on staging

PR #1961, opened 2025-11-12
PuterFSProvider decoupling

KernelDeimos:


KernelDeimos:
Copy Cartesian case 48 does not fail locally. Attempts to reproduce have not revealed any issue. I have no idea why apitest is failing but only on the GitHub action. I will investigate a little more but if I'm still unable to reproduce it tomorrow I will merge this anyway and frequently test copy and related operations as I continue working on this.

KernelDeimos:
I've verified that I am neither able to reproduce this when running apitest on a production deployment - copy cart 48 still passes. At this point I believe it's safe to assume this is an issue specific to the CI action somehow.

PR #1972, opened 2025-11-14
Add missing together image model prefixes

### LINKED ISSUE #1970: Inconsistent together ai model prefix
hi team, i noticed this problem when i tried to access ByteDance Seedream 4.0 model (from together ai)
it should be available looking at together ai, and the costmap

the error

Image

tracing parts of the code, it looks like the js lib failed to determine if this should use the together ai service, because there isn't the bytedance prefix (adding it should fix it?)

image prefix
https://github.com/HeyPuter/puter/blob/545b5a071e5c236e055d4d060ac883dd33b8a05c/src/puter-js/src/modules/AI.js#L13-L23

video prefix
https://github.com/HeyPuter/puter/blob/545b5a071e5c236e055d4d060ac883dd33b8a05c/src/puter-js/src/modules/AI.js#L33-L41

this is where it determines which AI service it chooses, by default because it doesn't found any, it just uses the one it initialized with, which is the openai, which explains why error message shows gpt image, dalle, etc
https://github.com/HeyPuter/puter/blob/545b5a071e5c236e055d4d060ac883dd33b8a05c/src/puter-js/src/modules/AI.js#L892-L918

reynaldichernando:
Fixes #1970 

- this should fix the proper selection for together ai model service
- only adding the missing ones here, instead of removing existing ones (even though their models are nonexistent)
  - the reason is, trying to be conservative, and don't want to break stuff
- the new model prefix is added from looking at the costmap btw, and seeing which are missing

maintainers feel free to edit this PR

### DIFF
--- a/src/puter-js/src/modules/AI.js
+++ b/src/puter-js/src/modules/AI.js
@@ -20,6 +20,13 @@ const TOGETHER_IMAGE_MODEL_PREFIXES = [
     'sg161222/',
     'wavymulder/',
     'prompthero/',
+    'bytedance-seed/',
+    'hidream-ai/',
+    'lykon/',
+    'qwen/',
+    'rundiffusion/',
+    'google/',
+    'ideogram/',
 ];
 
 const TOGETHER_IMAGE_MODEL_KEYWORDS = [


ProgrammerIn-wonderland:
this looks fine

PR #1963, opened 2025-11-13
puterfs decoupling continued

KernelDeimos:

PR #1982, opened 2025-11-18
PuterFS - Move storage into filesystem provider

KernelDeimos:
To be merged after #1963 - yes that means I'm two PRs deep right now. I need to start rigorous testing on Zenpacket very soon because I have been so far unable to reproduce the failures reported by apitest in either #1961 or #1963. This could also incur further delay.

Generally the following pieces of the filesystem refactor and simplification remain:
- move the rest of the storage code to the `puterfs` and `prodfs` extensions
- create simplified filesystem interface on `FilesystemService` adhereing to the [design document](https://github.com/HeyPuter/issues/blob/main/documents/doc-2510-001-fs-design.md) (not available publicly at present)


KernelDeimos:
I think I finally figured this out - apitest test fails here just like in #1963 but it's not because puterfs tests are failing, it's because memoryfs tests are failing. The memory filesystem will be updated later.

Issue #1862, opened 2025-10-30
toolbar is shown and hidden quickly when a window is snapped while the cursor is on the toolbar


jelveh:
This video demonstrates the annoying behaviour.

https://github.com/user-attachments/assets/2df35271-24db-4d77-b167-7d7af7b5d633

JohnBoulanger:
Hi there. I will look into a fix for this issue. Do you mind assigning me to it?

jelveh:
Awesome! assigned :)

PR #1989, opened 2025-11-20
fix: Add debounce to toolbar reveal when dragging windows to prevent accidental activation

JohnBoulanger:
Fix for issue: [#1862 ](https://github.com/HeyPuter/puter/issues/1862)

Fixed issue where toolbar would pop up briefly when a user drags a window to maximize over the toolbar

This PR introduces a small UX improvement to the desktop toolbar behavior. Previously, when a window was being dragged to the top of the screen to maximize it, the toolbar would pop up briefly causing an annoying unwanted behaviour.

Key Changes
- Added a debounce timer (dragReleaseDebounceTimer) that activates whenever a window drag operation ends.
- During the debounce period, the toolbar will not reveal itself, even if the mouse enters the toolbar area.
- Updated the toolbar mouseenter handler to respect the new debounce state.
- Ensures smoother drag interactions and prevents accidental toolbar activation during or immediately after window drag.
Result
Dragging windows feels more consistent, avoids unintended toolbar pop-ups.

https://github.com/user-attachments/assets/06f84a40-c9ee-4d18-94fe-2c491f0c835f

### DIFF
--- a/src/gui/src/UI/UIDesktop.js
+++ b/src/gui/src/UI/UIDesktop.js
@@ -1548,21 +1548,21 @@ async function UIDesktop(options) {
     };
 
     // debounce timer to prevent toolbar showing immediately after drag ends
-    window.dragReleaseDebounceTimer = null;
+    window.drag_release_debounce_timer = null;
     const DRAG_RELEASE_DEBOUNCE_MS = 300; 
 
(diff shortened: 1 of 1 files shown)


jelveh:
Nailed it! Thank you @JohnBoulanger. merging now.

PR #1991, opened 2025-11-20
"test mode" method overrides for services

KernelDeimos:


KernelDeimos:
This PR of mine is too stale to merge. I seem recall there was some backlash when I presented this idea. This is an interesting idea but it's not happening here. I'm closing this.

PR #2012, opened 2025-11-23
Add support for Gemini 3 Pro image preview model

Nihhaar0002:
This pull request introduces support for the new Gemini 3 Pro image preview model within Puter’s AI image generation system.

What’s Included

Frontend (AI.js)

Added a new alias: "nano-banana-pro" → "gemini-3-pro-image-preview"

Ensured the new model correctly routes through the existing gemini-image-generation service

Backend (GeminiImageGenerationService.js)

Added "gemini-3-pro-image-preview" to the list of supported models

Updated ratio validation so the new model uses valid dimensions

Modified the image generation call to use the model passed in the request rather than a fixed default

Additional Information

The pricing for the new model is currently aligned with the existing gemini-2.5-flash-image-preview (0.039 per 1024×1024). This can be adjusted by the maintainers if a different pricing structure is preferred.

I’m happy to make any adjustments or improvements based on feedback.

### DIFF
--- a/src/backend/src/modules/puterai/GeminiImageGenerationService.js
+++ b/src/backend/src/modules/puterai/GeminiImageGenerationService.js
@@ -37,9 +37,13 @@ class GeminiImageGenerationService extends BaseService {
     static MODULES = {
     };
 
-    _construct() {
+      _construct() {
         this.models_ = {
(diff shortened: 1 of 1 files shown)


ProgrammerIn-wonderland:
Merged this manually!

PR #2034, opened 2025-11-26
fix: add missing 'it' import from vitest in MeteringService.test.ts

parth-117:
## Summary

This PR fixes a bug in `MeteringService.test.ts` where the `it` function from vitest was being used without being imported.

## Problem

The test file was using `it('should have some services', ...)` on line 6, but the import statement on line 1 only imported `{ describe, expect }` from vitest, causing a `ReferenceError: it is not defined` when running tests.

## Solution

Added `it` to the import statement:
```typescript
import { describe, expect, it } from 'vitest';
```

This is a simple one-line fix that resolves the test failure.


Salazareo:
closing since we've cleaned up tests some more and don't think this is needed now

PR #2056, opened 2025-11-27
fix(emulator): correct byte/virtio stream helpers and wisp framing

2400060033:
Fixes multiple bugs in the browser emulator wiring:
- implement to_int correctly (little-endian)
- fix callback byte-stream async iterator
- correct ATStream carry handling
- improve virtio frame assembly and minor robustness

No behavior changes; improves stability of virtio/wisp packet handling.

### DIFF
--- a/src/puter-wisp/basic.html
+++ b/src/puter-wisp/basic.html
@@ -5,7 +5,7 @@
         font-size: 12px;
         line-height: 16px;
     }
-    BODY {
+    body {
         background-color: #111;
         display: flex;
         justify-content: center;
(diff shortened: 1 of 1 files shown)


ProgrammerIn-wonderland:
@r58Playz does this look correct?

src/puter-wisp/basic.html
```diff
@@ -19,393 +19,377 @@
...
-        }
-        async run_ () {
-            const handlers_ = {
+        init() { this.run_(); }
+
+        async run_() {
+            const handlers = {
                 [WispPacket.INFO.id]: ({ packet }) => {
-                    // console.log('guess we doing info packets now', packet);
+                    // reflect INFO packet back
```
r58Playz, line 363: This is technically incorrect according to spec but it doesn't really matter for the extensions puter is using since they have no payload on either side

r58Playz (commented):
The actual v86 emulator doesn't use this code. You should probably modify `src/emulator/src/main.js` and `src/puter-wisp/src/exports.js`.

PR #2060, opened 2025-11-27
More test updates and cleanup

KernelDeimos:


KernelDeimos:
This PR went unmerged, I assume erroneously. This branch name is not in the file where I keep track of everything I'm doing - I have `25BR0-tests-core`. That said, these changes don't look incredibly important; most of it is formatting changes because of eslint and this is likely also the cause of conflicts. I'm closing this because it is likely not worth the effort to bring it back.

Issue #1658, opened 2025-10-03
[Feature Request] Auto-delete files from the recycle bin


wuniversales:
Add a configuration menu to enable automatic deletion of files from the recycle bin.

jelveh:
I think this is a pretty cool idea! I need to think more about it but we'll consider it.

Pahul-Sidhu:
I would like to implement this feature if you guys want

hanifjilani:
I would like to implement this feature as a part of my university course. It will include the feature implementation, relevant unit tests and mocking if necessary.

jelveh:
Hi @hanifjilani, thank you for your interest in contributing to Puter. I just assigned this issue to you. Let me know if I can help in any way.

hanifjilani:
Hey, I have started working on this project. I will let you know once I am done. Once again, thanks for assigning it to me!

hanifjilani:
I’m finalizing the auto-delete contextual menu for the trash bin and have implemented a worker in the backend to handle the auto-delete checks. I wanted to confirm where you’d prefer the related tests to be placed — the worker code is currently located at puter/src/backend/src/services/worker.

Additionally, for the UI components that set user preferences via KV (e.g., choosing 24 hours, 7 days, or 30 days for auto-delete), I’d like to add some tests as well. Could you please let me know your preferred location for those frontend tests?

Apologies if these are basic questions — I’m still getting familiar with the contribution structure and want to make sure everything aligns with the project’s conventions.

Thanks

hanifjilani:
Just submitted a PR solving this feature request. Thank you so much for your help in the discord. #2079

hanifjilani:
I would be extremely grateful, if you could review my PR and provide feedback and if possible accept it. Accepting the PR will provide my extra credit for my college project. Thanks!

hanifjilani:
https://github.com/user-attachments/assets/3bf53b01-8585-4f66-a0ec-13e58f40a5fc


https://github.com/user-attachments/assets/7c673b6e-e0b1-4f5f-be51-f531d71c6407

PR #2079, opened 2025-12-02
Auto-delete Feature for Trash (Backend + Frontend + Tests) (Solving ISSUE: #1658)

hanifjilani:
# ✨ Auto-delete Feature for Trash (Backend + Frontend + Tests)

This PR introduces a complete Auto-delete system for the Trash, integrating both backend logic and frontend behavior while preserving Puter’s existing UX patterns.

Solved the **[Feature Request] Auto-delete files from the recycle bin** #1658 

## Backend: Periodic Auto-Delete Service

Added a new `TrashAutoDeleteService` running every 30 minutes.

For each user, the service:

- Reads their auto_delete_days preference from kv.

- Finds files in the user’s Trash folder.

- Determines which entries are expired based on the configured time window.

- Emits a trash.auto_delete socket event for each expired file.

- No files are deleted directly in the backend — it only emits events.

- The frontend handles deletion using the same existing UX flow as manual deletes.

### Context Menu Integration

Auto-delete settings are now accessible from three key places:

- Trash window context menu

- Trash desktop icon context menu

- Trash taskbar icon context menu

Users can choose:

- Disabled

- 24 hours

- 7 days

- 30 days

The selection is stored using Puter’s kv system, allowing the backend service to access user preferences.

### Trash UI Reaction

When the backend emits a trash.auto_delete event:

The frontend seamlessly removes the entry from the Trash UI.

This uses the same code path as normal Trash deletions, ensuring consistent behavior.

### Backend (Vitest)

Added a full unit test suite covering:

- Single expired item deletion

- Multiple expired items

- Disabled preference (0)

- Missing or malformed timestamps

- Multiple users with independent Trash folders

- DB failure logging behavior

- Periodic scheduling logic

These tests avoid requiring a full Kernel environment by fully mocking DB, helpers, and socket services.

### Frontend (Playwright)

Added a Playwright test to verify that the frontend correctly handles socket-emitted auto-delete events.

## 📁 Summary

This PR adds a complete, fully tested Auto-delete system that:

- Respects per-user preferences

- Uses Puter’s existing architecture (KV, services, socket events)

- Integrates smoothly into the frontend with minimal changes

- Maintains UX consistency with existing Trash operations

Let me know if you'd like the PR broken into smaller commits or if any part should be refactored!

### DIFF
--- a/src/backend/src/CoreModule.js
+++ b/src/backend/src/CoreModule.js
@@ -307,6 +307,9 @@ const install = async ({ context, services, app, useapi, modapi }) => {
     const { EdgeRateLimitService } = require('./services/abuse-prevention/EdgeRateLimitService');
     services.registerService('edge-rate-limit', EdgeRateLimitService);
 
+    const { TrashAutoDeleteService } = require('./services/TrashAutoDeleteService');
+    services.registerService('trash-auto-delete', TrashAutoDeleteService);
+    
     const { CleanEmailService } = require('./services/CleanEmailService');
     services.registerService('clean-email', CleanEmailService);
 
--- a/src/backend/src/services/TrashAutoDeleteService.js
+++ b/src/backend/src/services/TrashAutoDeleteService.js
@@ -1,0 +1,140 @@
+/*
+ * Copyright (C) 2024-present Puter Technologies Inc.
+ *
+ * This file is part of Puter.
+ *
+ * Puter is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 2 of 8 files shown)


hanifjilani:
https://github.com/user-attachments/assets/3bf53b01-8585-4f66-a0ec-13e58f40a5fc


https://github.com/user-attachments/assets/7c673b6e-e0b1-4f5f-be51-f531d71c6407

KernelDeimos:
Hello, thanks for submitting this PR. The effort here is appreciated. This will require changes before we're able to merge it into the repo, however. I'm aware that you were originally hoping to have files deleted from backend rather than frontend, and this implementation reflects that. Either the backend or the frontend should be responsible for this, not both; here's why:

We're spending resources (database mostly) iterating over all users and determining which files are expired, to then send a socket event to a user who probably isn't online, expecting them to handle the operation.

We need either proper batching on backend, or a fully-frontend implementation. For this I would actually recommend a fully-frontend implementation because it will be simper to implement and resource consumption grows with user sessions.

### Fully Backend Approach
- Files are also deleted on backend, no socket event
- The 30-minute "scan all users" interval does **not** work. You don't know how many users there are, and therefore you cannot make an assumption about how much time is enough for the interval. Also, you probably should not scan all users here either way; better to do so in chunks of 100-or-so users. Parameters like idle time and user chunk size should be managed by ParameterService because they may need to be tuned.

### Fully Frontend Approach
- Interval can correspond to user's configuration. The operation can be expected not to take too long.
- Schedule and iteration through files are on frontend.

PR #2086, opened 2025-12-02
chore: add typedefs for puter.js

Salazareo:


src/puter-js/types/shared.d.ts
```diff
@@ -0,0 +1,39 @@
+export type PuterEnvironment = 'app' | 'gui' | 'web' | 'web-worker' | 'service-worker' | 'nodejs';
```
reynaldichernando, line 1: confirming: is it expected that we add another environment (web-worker) here? because it didn't exist previously

src/puter-js/types/modules/ai.d.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+export interface ChatResponse {
+    message?: ChatMessage;
+    messages?: ChatMessage[];
+    choices?: unknown;
+    [key: string]: unknown;
+}
+
+export type ChatStreamChunk = Record<string, unknown>;
```
reynaldichernando: are we sure we want to rename this type? if that's the case we need to update our docs for it

https://docs.puter.com/AI/chat/#return-value
https://docs.puter.com/Objects/chatresponsechunk

src/puter-js/types/modules/ai.d.ts
```diff
@@ -0,0 +1,122 @@
...
+    temperature?: number;
+    max_tokens?: number;
+    stream?: boolean;
+    vision?: boolean;
+    [key: string]: unknown;
+}
+
+export interface ChatResponse {
+    message?: ChatMessage;
+    messages?: ChatMessage[];
```
reynaldichernando: haven't tested this yet, but i'm surprised it's possible that we return a "messages" field, like it being an array, instead of a single message

src/puter-js/types/modules/ai.d.ts
reynaldichernando: hmm also this should be 
```
{
  text?: string,
  reasoning?: string
}
```

in our previous type we just have the reasoning be the arbitrary key, but i think we should atleast have a fixed type for this

Salazareo: ill update

Salazareo: fixed

src/puter-js/types/modules/ai.d.ts
Salazareo: hmm i believe this is correct, but could be wrong

src/puter-js/types/shared.d.ts
Salazareo: we do have this, but I think it may not be used anywhere yet? @ProgrammerIn-wonderland do you know?

ProgrammerIn-wonderland: if you use puter.js in a normal web worker it *should* detect as "web-worker" but I haven't tested this

ProgrammerIn-wonderland: image
yeah

src/puter-js/types/modules/ai.d.ts
Salazareo: fixed, was not correct, we coerce it as messages

src/puter-js/types/shared.d.ts
reynaldichernando: thanks!

PR #2100, opened 2025-12-05
Fix AI types

reynaldichernando:
- remove constructor
- remove global methods, setAuthToken, setAPIOrigin
- use stricter type (removing `[key: string]: unknown;`)
- fix streaming intellisense
- add missing imageURLArray handling
- fix all positional testmode and options
- simplify txt2speech
- add missing options everywhere
- add missing return type for chat
- fix return type speech2speech

and probably some more that i didn't write here


reynaldichernando:
after some more thinking, even if we miss a type or two, it won't be a breaking change for end user, because they can still disable the typescript check for a particular type

PR #2105, opened 2025-12-07
dev: limit concurrent gemini image generations

KernelDeimos:


Salazareo (changes requested):
this can create a potentially problematic request queue just waiting for this...

we're no longer seeing OOM issues, so not sure if this is the solution we should have

src/backend/src/modules/puterai/GeminiImageGenerationService.js
```diff
@@ -46,6 +49,10 @@ class GeminiImageGenerationService extends BaseService {
                 '1024x1024': 0.156,
             },
         };
+        // Queue for pending generation tasks
+        this.generationQueue_ = [];
```
Salazareo, line 53: why the lading trail? if its private it can just be private?

Salazareo:
nvm on my comment before, see that we just don't generate if the queue is full. 

I think we should revisit this properly though, this would make our service kinda shitty I feel.

also this has changed now so might need to be moved over

src/backend/src/modules/puterai/GeminiImageGenerationService.js
KernelDeimos: I'm still used to this convention. Are we enforcing the use of `#` to mark private members? I think we could accomplish the same thing by adding a linter rule to disallow access to members like this from other files - then we could also access them inside tests more easily and we would not be restricted from using a hack for an urgent hotfix if that was ever necessary.

PR #2133, opened 2025-12-11
Fix speech2speech options

reynaldichernando:
- removing the camel case version, and only have the snake case
- aligned with what we have on https://docs.puter.com/AI/speech2speech/


Salazareo (approved):

ProgrammerIn-wonderland:
I think the input actually accepts both types

reynaldichernando:
yeah, this is expected so we only have one version to not confuse users

PR #2151, opened 2025-12-12
Add missing AI types

reynaldichernando:
- went through this one more time to see anything missing
- mostly test_mode
- and some in txt2vid and speech2txt


Salazareo (commented):
can you actually also update the  

width?: number;
height?: number;
aspect_ratio?: string;

their now just funelled through `ratio`

Salazareo (approved):

PR #2164, opened 2025-12-13
Adjust costs for image generation models

AkshayCoder48:
Updated costs for image generation models to reflect new pricing. To add wider access to expensive models and free models.


jelveh:
these are nto accurate

PR #2181, opened 2025-12-15
Fix more puter.js typescript types

reynaldichernando:
- networking
  - rename the types
  - adjust on error param type
- os (we don't expose this, but might as well)
  - fix import
  - remove constructor and global function
- ui
  - add missing params and methods
- util (we don't expose this aswell)
- workers
  - remove constructor and global function
- puter global
  - remove constructor
  - fix weird networking import
- apps
  - update AppRecord -> App
  - remove constructor and global function

notes:
- i'm doing changes that are obvious here, haven't gotten into the actual implementation, where usually there are more params and overloads, etc
- will do another pass on this to make sure i didn't miss any
- other than this, i haven't touched perms and drivers

### DIFF
--- a/src/puter-js/types/modules/networking.d.ts
+++ b/src/puter-js/types/modules/networking.d.ts
@@ -8,7 +8,7 @@ export type SocketEvent =
     | 'tlsopen'
     | 'tlsclose';
 
-export class Socket {
+export class PSocket {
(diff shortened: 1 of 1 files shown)


reynaldichernando:
thanks for the review @ProgrammerIn-wonderland ,i've gone ahead and updated them

PR #2184, opened 2025-12-15
telemetry: config telemetry to run by default

Salazareo:


src/backend/src/modules/perfmon/TelemetryService.js
```diff
@@ -78,12 +76,20 @@ class TelemetryService extends BaseService {
     }
 
     #getConfiguredExporter () {
-        if ( config.jaeger ?? this.config.jaeger ) {
-            return new OTLPTraceExporter(config.jaeger ?? this.config.jaeger);
+        if ( jaeger ?? this.config.jaeger ) {
+            return new OTLPTraceExporter(jaeger ?? this.config.jaeger);
+        }
+        if ( this.config.devConsole ) {
```
KernelDeimos: I would just call this `console` rather than `devConsole`

src/backend/src/modules/perfmon/TelemetryService.js
```diff
@@ -38,13 +37,13 @@ class TelemetryService extends BaseService {
                         }));
 
         const exporter = this.#getConfiguredExporter();
-        this.exporter = exporter;
+        const metrictExporter = this.#getMetricExporter();
```
KernelDeimos: Should this be `metricExporter`?

src/backend/src/modules/perfmon/TelemetryService.js
```diff
@@ -78,12 +76,20 @@ class TelemetryService extends BaseService {
...
+            return new ConsoleSpanExporter();
         }
-        const exporter = new ConsoleSpanExporter();
-        return exporter;
     }
-}
 
-module.exports = TelemetryService;
+    #getMetricExporter () {
+        if ( jaeger ?? this.config.jaeger ) {
```
KernelDeimos: What is `jaeger` here? `config.jaeger` was much clearer. Maybe we should remove support for a global config entry for this - I don't think anybody else was using this.

PR #2186, opened 2025-12-16
Add optional requestCallbacks options to getUser

reynaldichernando:
- the getUser method has the legacy callback (success, error), adding it here for completeness

note:
- i'm not gonna document the callback in docs, because i think it will just confuse users
- same with all the fs stuff that has callbacks


reynaldichernando:
hi @Salazareo , could you help to review this PR, thank you 🙏

Salazareo (approved):

PR #2189, opened 2025-12-17
Add flux.2 max for together provider

reynaldichernando:
- add flux.2 max to provider model.ts and costMap
- i'm not sure if we are still using the costMap (?)
- although i believe we used to allow all models from together directly, but has changed to explicitly add it before its enabled

info:
- together hasn't published pricing for this actually
- i got this number $0.07 per MP from black forest labs
- looking at previous flux.1 and flux.2 models it's the same between bfl and together, so i predict it will be the same
- if someone can review and cross check this, i would appreciate that a lot


jelveh:
yeah, we don't use cost maps any longer. 

I'm testing right now.

jelveh:
works like a charm. Merging.

PR #2197, opened 2025-12-18
Fix apps types

reynaldichernando:
- add CreateAppResult return type for create method
- add getDeveloperProfile overload
- remove unused RequestCallbacks (those implementation don't use callbacks)


reynaldichernando:
hi @Salazareo , could you help to review this PR, thank you 🙏

PR #2199, opened 2025-12-18
Update gemini models config

reynaldichernando:
for some reason the numbers we have are inaccurate, compared to google own docs
i found these while adding the gemini 3 flash

- gemini 3 pro preview should have 65,536 max tokens (source: https://ai.google.dev/gemini-api/docs/models#gemini-3-pro)
- gemini 2.5 pro should also have 65,536 max tokens (https://ai.google.dev/gemini-api/docs/pricing#gemini-2.5-pro)
- gemini 2 flash should have 1M context (https://ai.google.dev/gemini-api/docs/models#gemini-2.0-flash)

please help cross check and review this

although nothing is wrong/broken yet with these models? 
they still work, but idk what the implication is having these incorrect information

### DIFF
--- a/src/backend/src/services/ai/chat/providers/GeminiProvider/models.ts
+++ b/src/backend/src/services/ai/chat/providers/GeminiProvider/models.ts
@@ -4,7 +4,7 @@ export const GEMINI_MODELS: IChatModel[] = [
     {
         id: 'gemini-2.0-flash',
         name: 'Gemini 2.0 Flash',
-        context: 131072,
+        context: 1_048_576,
         costs_currency: 'usd-cents',
         input_cost_key: 'prompt_tokens',
         output_cost_key: 'completion_tokens',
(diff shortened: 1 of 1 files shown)


reynaldichernando:
hi @ProgrammerIn-wonderland  , could you help to review this PR, thank you 🙏

ProgrammerIn-wonderland:
This looks fine but I want to make sure there aren't weird effects with max_tokens first since max_tokens means something completely different for Gemini (maximum output tokens) than it does for the OpenAI models (max tokens in conversation). I think this change has the potential to cause issues

reynaldichernando:
noted, btw for the gemini 3 flash, i followed the number in the docs, which shows 65,536

> I want to make sure there aren't weird effects with max_tokens

okay, i'm not sure how the internals work yet for this, but just bringing this to your attention 🫡

Salazareo:
so I think these should be good to be different than openai ones, but i won't merge in case @ProgrammerIn-wonderland finds something off with it

ProgrammerIn-wonderland:
Yeah the issue here is still that we associate Max tokens with context length internally and do math based on that, we probably need to do a secondary, alternate calculation to properly follow geminis.

PR #2203, opened 2025-12-18
add user iteration to all methods which return an App object

ProgrammerIn-wonderland:


ProgrammerIn-wonderland:
I hate this test, @Salazareo what does fail to get admin password even mean

PR #2211, opened 2025-12-22
Entity Refactor (part 0)

KernelDeimos:
No significant changes in this PR - I just want to get my first couple of commits in `main` in case `Apps.js` is going to have more modifications in the near future so there's no conflict because of the linter.

Subsequent PR will introduce a new service that will eventually replace `es:app`.


KernelDeimos:
API Test is broken again, but puter-js test which is relevant to this change passes, so I am merging this.

PR #2255, opened 2026-01-09
Model singularity

ProgrammerIn-wonderland:


Salazareo (approved):
LGTM!

follow up though could be making the puterId be the default display in puter.com/puterai/*/models

ProgrammerIn-wonderland:
> puter.com/puterai/*/models

to my knowledge it is 
image

PR #2266, opened 2026-01-12
Improve UI for App Deployment Options in Dev Center

### LINKED ISSUE #604: [Dev Center] Improve UI for App Deployment Options in Dev Center
The Dev Center currently offers two methods for deploying apps:

1. Uploading assets (HTML, JS, CSS, etc.)
2. Setting the app's index URL to any URL

While the second option is powerful and appreciated by developers, the current UI does not clearly present both options. This leads to confusion among developers, with many believing that uploading assets is the only way to create apps.

We need to improve the UI to make both deployment methods more understandable and equally visible. This may include:

- Redesigning the app creation/deployment flow
- Adding clear, visual separation between the two methods
- Providing brief explanations or tooltips for each option
- Ensuring that the custom index URL option is as prominent as the asset upload option

rowin-C:
resolves #604
- Added CSS styling for deployment option cards
- Implemented active state switching between deployment methods with guardrails
- Abstracted saveEditApp for reuse across Deploy and Edit tabs
- Added error handling for invalid URLs in the Deploy tab

### DIFF
--- a/src/dev-center/js/apps.js
+++ b/src/dev-center/js/apps.js
@@ -297,13 +297,16 @@ $(document).on('click', '.deploy-app-card', function () {
     if ( $(this).hasClass('deploy-app-active') ) {
         return;
     }
+    $('.deploy-btn').addClass('disabled');
     if ( $(this).hasClass('deploy-app-card-files') ) {
         $('#deploy-app-index-url').val('');
     }
     if ( $(this).hasClass('deploy-app-card-link') ) {
         reset_drop_area();
+        if ( $('#deploy-app-index-url').val() != '' ) {
+            $('.deploy-btn').removeClass('disabled');
+        }
     }
-    $('.deploy-btn').addClass('disabled');
     $('.deploy-app-card').removeClass('deploy-app-active');
     $(this)
         .addClass('deploy-app-active')


jelveh:
Thanks @rowin-C, have been very much looking forward to this PR 🎉

A few things:
- Both options have this description: "Some description about the mode"
- It seems like the URL "https://dev-center.puter.com/coming-soon.html" is not deployable unless it's changed first. I'd say it should be deployable as it's a good default URL until the developer deploys the real version
- Entering a URL and pressing deploy, throws the following eror:
Screenshot 2026-01-11 at 8 16 51 PM

rowin-C:
Thanks @jelveh , Glad to hear that.

I’ll update the descriptions to:

Upload Assets: “Upload and deploy static assets like HTML, CSS, and JavaScript directly to host your app.”

External URL: “Set the app’s index to any publicly accessible URL where your app is already deployed.”

Regarding https://dev-center.puter.com/coming-soon.html: the URL is deployable, but currently the input needs to be interacted with first. If we don’t do that, the Deploy button would be active by default since the input already has a value.

Do we want the Deploy button to be active every time Dev Center loads, or only after the user explicitly interacts with the input?

jelveh:
> Do we want the Deploy button to be active every time Dev Center loads, or only after the user explicitly interacts with the input?

For "URL mode", the deploy button should be enabled by default unless the user removed the URL

jelveh:
🎉🎉🎉

PR #2278, opened 2026-01-14
feat(gui): add keyboard shortcuts guide dialog

RahulAswal01:
**## Summary**
Adds an initial Keyboard Shortcuts guide UI to the Puter GUI.

**## Details**
- Introduces a new UI component to display keyboard shortcuts.
- Uses the existing dialog/window system (UIWindow).
- Does not change shortcut logic or bindings.
- Keeps the change isolated to GUI.

**## Notes**
This PR focuses on the UI foundation only.
Further wiring (Help menu, shortcuts like F1/Ctrl+?) can be added in a follow-up if desired.

### DIFF
--- a/src/gui/src/UI/keyboardshortcut.js
+++ b/src/gui/src/UI/keyboardshortcut.js
@@ -1,4 +1,6 @@
-export function KeyboardShortcuts () {
+import UIWindow from './UIWindow.js';
+
+export async function openKeyboardShortcuts () {
     const shortcuts = [
         { key: 'Ctrl + ,', action: 'Open Settings' },
         { key: 'Ctrl + K', action: 'Open Search' },
@@ -21,5 +23,11 @@ export function KeyboardShortcuts () {
     container.appendChild(title);
     container.appendChild(list);
 
-    return container;
+    await UIWindow({
+        title: 'Keyboard Shortcuts',
+        body_content: container,
+        width: 500,
+        height: 400,
+        init_center: true,
+    });
 }


RahulAswal01:
Thanks for reviewing!
This PR adds the initial Keyboard Shortcuts guide UI only.
Happy to adjust structure, naming, or placement based on feedback.

jelveh:
thanks for the pr. why did you change PuterDialog?!

RahulAswal01:
Thanks for the feedback!
I changed PuterDialog because I initially used it as a quick way to open the Keyboard Shortcuts UI, but I understand this may not be the right place for it.
I’m happy to revert the PuterDialog changes and keep the keyboard shortcuts guide fully self-contained (or move it to a more appropriate entry point if you prefer). Please let me know which approach you’d like.

jelveh:
yeah puter dialog is reserved for a special notice. Can you add a specific dialog for it? nothing fancy, just a simple dialog with the shortcuts listed

RahulAswal01:
Got it, thanks for clarifying 👍  
I’ll revert the PuterDialog changes and add a simple, dedicated dialog just for listing keyboard shortcuts.

RahulAswal01:
Thanks! I’ve reverted the PuterDialog changes and updated the shortcuts guide to use a dedicated, simple dialog.

RahulAswal01:
Thanks for the merge earlier!
I noticed the change was later reverted. I completely understand priorities can shift — I just wanted to ask so I can learn: was the revert mainly due to UX direction, timing, or placement of the feature?
Happy to rework or reintroduce it in the preferred place if helpful.

PR #2280, opened 2026-01-14
Revise README.ro.md for updates and corrections

sn-o-w:
My mother tongue is Romanian, and I wasn’t a fan of the capitalization in the Romanian text.

In Romanian, capitalizing lots of words like that isn’t really correct.


jelveh:
Thank you!

PR #2294, opened 2026-01-16
more accurate tokenization

ProgrammerIn-wonderland:


ProgrammerIn-wonderland:
The main reason for this PR originally was an openrouter bug but now that that is mitigated, it may not be necessary or worth the RAM overhead of loading these tokenizers into memory

Issue #2125, opened 2025-12-10
[Feature Request]: User Friendly System Information Overview


hassan689:
Introduce a System Information & Resource Manager inside Puter to display client-side details, server specs, and user resource usage .This feature improves transparency, diagnostics, and gives Puter a more complete _operating system_ feel.

**Users currently lack visibility into:**

- What device/browser they are using
- What server resources the Puter instance has
- Their storage limits and usage
- Their performance metrics and network conditions
- A centralized location for system diagnostics

**Adding a System Information window will:**

- Reduce support questions
- Improve debugging and performance troubleshooting
- Make Puter feel more like a real OS
- Enable future features like virtual storage and resource limits

**Use Cases**

**Device Information**
Users can see browser, OS, screen resolution, CPU cores, RAM, network info, etc.

**Server Specifications**
Users can view the current Puter instance’s CPU, RAM, disk usage, and uptime.

**Resource Monitoring**
Users can view their storage usage, limits, and available capacity.

**Storage Management**
Users can request additional storage space from admins . (This can be further implementable in the future)

**System Diagnostics**
Users can view performance metrics, network speed, load time, and other diagnostics.

Implementation (Client-Side):

- window.performance
- screen.width
- screen.height
- navigator.hardwareConcurrency
- navigator.deviceMemory

Implementation (Server-Side):

**Server OS
Server CPU model & core count
Server RAM (free/total)
Server disk usage (no paths)
Server uptime**

UI will look like this:
 
<img width="930" height="857" alt="Image" src="https://github.com/user-attachments/assets/c626f0eb-0952-498e-9e9c-00cfa22b48ed" />
<img width="899" height="888" alt="Image" src="https://github.com/user-attachments/assets/6aaaa5fe-1477-4c0a-8609-3a742b6cd433" />
<img width="888" height="888" alt="Image" src="https://github.com/user-attachments/assets/82003bdd-2a76-4124-b6fb-36eaf27566c4" />

I request you to allow me to work on this one :)

jelveh:
Hi there,
Thank you for the awesome suggestion and your patience while we clear up the backlog. I think this is a great idea and I'd love to see it implemented in Puter. I also think this would be an excellent candidate for contributors. So I'm going to assign "help-wanted" to see if anybody is going to pick it up.

R0h1tAnand:
Hi — I’d like to pick this up.

Plan: add a “System Information” window (client + server).

[...]

PR #2311, opened 2026-01-21
Feat: Add system info (Client + Server metrics)

llpingll:
This PR adds a System Information window with client-side and server-side information. Accessed via the toolbar -> profile -> System Information.

Screenshot 2026-01-23 001754
Screenshot 2026-01-23 001733

(Showed settings window to display similarity in style)

It displays:
- **Client-side**:
  - Browser name/version (User-Agent Client Hints + fallback)
  - OS name/version
  - CPU cores
  - Approximate RAM
  - Screen resolution, pixel ratio, color depth
- **Server-side**:
  - OS (platform/type/release)
  - CPU model & logical core count
  - RAM (free/total in GB)
  - Disk usage (root filesystem, used/total in GB)
  - Uptime (formatted as d/h/m)
  - Exposed via `/getServerInfo` API endpoint

It features:
- Unified rendering logic for both client & server info
- Graceful error handling (empty list on API failure → no crash)
- Refresh buttons with animation
- Initial load + on-demand refresh

Testing:
- Tested locally in self-hosted mode
- API endpoint returns correct data (logs + browser response)
- Refresh buttons work with animation
- Error path (e.g., API down) shows empty section gracefully

**Regarding the existing PR #2286**
I noticed @R0h1tAnand submitted a PR (#2286) that adds a System Information window with both client-side and server-side metrics. I wasn't sure if it was fully complete or if there were still open review points, so I went ahead and completed a full implementation independently.

I'm still happy to:
- Close this PR if #2286 already covers this fully
- Rebase/merge my changes into #2286 (happy to help resolve conflicts)
- Collaborate on combining the approaches

Just let me know what direction you'd like to go in, goal is to get a solid System Information feature merged in the best shape possible.

Would love any feedback on the client-side code, styling, or if you'd prefer this integrated differently. Thanks for reviewing!

Solves #2125

PR #2337, opened 2026-01-24
feat: Implement Backend SystemInfoService for Server Specifications

rajashekarpatha07:
## Description
This PR implements the backend service required for **Issue #2125**.
It introduces a new `SystemInfoService` that exposes server-side specifications (CPU, RAM, OS, Disk) to the kernel.

## Changes
- Created `src/backend/src/services/SystemInfoService.js` (with JSDoc & lint compliance).
- Registered the service in `CoreModule.js` as `system-info`.

## Technical Details
The service utilizes Node.js `os` and `child_process` to return a unified JSON object containing:
- **OS:** Platform, Distro, Release
- **CPU:** Model, Cores, Load Average
- **Memory:** Total, Free, Used
- **Disk:** Total, Used, Available (Root partition)

### Design Decision
I noticed `HostDiskUsageService` exists, but it is limited to disk stats and uses synchronous execution.
My new `SystemInfoService`:
1. Aggregates **CPU, RAM, OS, and Disk** data into a single response (required for the System Info UI).
2. Uses **asynchronous** `exec` (`util.promisify`) to prevent event loop blocking, unlike the synchronous implementation in `HostDiskUsageService`.

## Testing
- Verified locally on Ubuntu Linux.
- Service returns valid JSON object with correct hardware stats.

### DIFF
--- a/src/backend/src/services/SystemInfoService.test.js
+++ b/src/backend/src/services/SystemInfoService.test.js
@@ -1,0 +1,75 @@
+const { describe, it, expect, vi, beforeEach } = require('vitest');
+const { SystemInfoService } = require('./SystemInfoService');
+const os = require('os');
+const child_process = require('child_process');
+
+// 1. Mock Node.js built-ins
+vi.mock('os');
+vi.mock('child_process', () => ({
+    exec: vi.fn(),
+}));
+
+describe('SystemInfoService', () => {
+    let service;
+    let mockExec;
+
+    beforeEach(() => {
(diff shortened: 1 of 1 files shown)


Salazareo:
thanks for the contribution. I've seen that https://github.com/HeyPuter/puter/pull/2311 this PR had already been assigned so will be taking that one just to avoid further confusion.

We'll do better to properly assign tasks in the future since a few people seemed to have added the same feature, sorry about that

PR #2338, opened 2026-01-24
fix: skip CI tests when only docs are changed (#2335)

### LINKED ISSUE #2335: Skip github actions test if working on `src/docs`
We recently moved [puter.js](https://github.com/heyputer/docs) docs to this repo, and with the first docs PR it's running like api test and backend tests.
It's not needed for docs, so we can just skip the github aciton workflows

serilevanjalines:
## Description
Skip test workflows (backend tests, API tests, vitest) when only documentation files in `src/docs/` are modified.

## Changes
- Added `paths-ignore: ["src/docs/**"]` to both `push` and `pull_request` triggers in `.github/workflows/test.yml`

## Why This Change?
After moving puter.js docs to this repo, CI workflows run unnecessarily for docs-only changes, wasting resources and time.

## Testing
- Documentation-only changes will now skip all test jobs
- Code changes continue to trigger full test suite as before

Fixes #2335

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,8 +3,12 @@ name: test
 on:
   push:
     branches: ["main"]
+    paths-ignore:
+      - "src/docs/**"
   pull_request:
     branches: ["main"]
+    paths-ignore:
+      - "src/docs/**"
 
 jobs:
   test-backend:


reynaldichernando:
in this PR its still running the github actions
although this is just due to how it works, plus since we are working on the non docs folder

it will only be ignored on subsequent docs PRs

(i encountered this in the past)

serilevanjalines:
@reynaldichernando Thanks for the quick review and merge! 
Yeah, that makes sense , the paths-ignore filter will only apply to future PRs since this one was modifying the workflow itself. Appreciate the explanation !

PR #2344, opened 2026-01-26
Fix issues with URL fetch

KernelDeimos:


KernelDeimos:
Needed to do a rebase atop a branch with a revert for disabling URL input

PR #2346, opened 2026-01-26
Separate subdomains from readir

KernelDeimos:
These changes separate readdir from fetching subdomains to improve performance of displaying directory items in the ui. This was AI-assisted and needs further testing so I'm putting this in draft for now.

### DIFF
--- a/src/backend/src/filesystem/hl_operations/hl_readdir.js
+++ b/src/backend/src/filesystem/hl_operations/hl_readdir.js
@@ -39,7 +39,7 @@ class HLReadDir extends HLFilesystemOperation {
         });
     }
     async __run () {
-        const { subject: subject_let, user, no_thumbs, no_assocs, actor } = this.values;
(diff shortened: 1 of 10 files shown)


Salazareo (dismissed):
nit: can we move this to extensions? feels more appropriate there than as eggspress

also, can you use esm if possible, we gotta start modernizing that

KernelDeimos:
> nit: can we move this to extensions? feels more appropriate there than as eggspress
> 
> also, can you use esm if possible, we gotta start modernizing that

We can't move this to an extension currently. This is core functionality - Puter would be broken with the changes this PR adds to gui if the new `/readdir-subdomains` endpoint was moved to an extension.

Salazareo:
> > nit: can we move this to extensions? feels more appropriate there than as eggspress
> > also, can you use esm if possible, we gotta start modernizing that
> 
> We can't move this to an extension currently. This is core functionality - Puter would be broken with the changes this PR adds to gui if the new `/readdir-subdomains` endpoint was moved to an extension.

oh why would it not work if the route was in extension?

Issue #2342, opened 2026-01-26
Fix resume analyzer example in docs


reynaldichernando:
We have a resume analyzer example in docs playground: https://docs.puter.com/playground/ai-resume-analyzer/

it's supposed to be able to read PDF, DOC, or TXT, but in reality only PDF is supported, because that's the only format natively supported by the model we are using. (sonnet 4)

We need to do some preprocessing for doc and txt to make the example work.

some ideas:
- parse doc file with library such as mammoth.js and pass the text in the prompt
- parse the txt file and pass it as text in the prompt instead of sending file
- also update our model to latest (sonnet 4.5 is good)

doesn't strictly have to be these, but above are some general idea

reynaldichernando:
here are example files that can be used for testing

[Resume.pdf](https://github.com/user-attachments/files/24857682/Resume.pdf)
[joma.txt](https://github.com/user-attachments/files/24857680/joma.txt)
[resume.docx](https://github.com/user-attachments/files/24857681/resume.docx)

HithaBadikillaya:
Hey! @reynaldichernando can I be assigned to this issue? I’m happy to take it on and push a fix.

reynaldichernando:
assigned! @HithaBadikillaya

Devansh-18:
Hi @reynaldichernando ,
I understand this issue is currently assigned. I’ve previously implemented a similar functionality - example here:
https://github.com/Devansh-18/pariksha-app/blob/master/src/lib/pdfParser.ts

Please let me know if I can take this up. I will surely try to fix this too.

reynaldichernando:
hi @Devansh-18 , thank you for your interest 
since right now the issue is already assigned to @HithaBadikillaya , i will let her finish the work
or if she decided she can't continue with the issue, then i can simply unassign it

don't worry, there are lots more issues need help coming where you can contribute!

HithaBadikillaya:
Hi @reynaldichernando thanks for assigning this issue to me. I ran into persistent environment-specific build issues on Windows (native dependency compilation with Node 24) and won’t be able to make progress in a reasonable timeframe. 

Please feel free to unassign me so someone else can take this forward. Appreciate the understanding.

reynaldichernando:
noted, and thanks for the confirmation, i've reassigned the issue to @Devansh-18 🫡

Devansh-18:
I will try to do it asap @reynaldichernando

reynaldichernando:
[...]

PR #2359, opened 2026-01-28
fix(resume-analyzer): add preprocessing for DOCX and TXT files

Devansh-18:
### Summary
This PR updates the **AI Resume Analyzer** example in the Puter Docs Playground so that it behaves correctly when users upload **DOCX or TXT** files.  
Previously, the example attempted to upload all file types directly to `puter.ai.chat`, but **Claude Sonnet 4 only supports native PDF parsing**, resulting in errors for DOCX/TXT.

This update ensures that the example actually works as advertised.

---

#### 1. Added preprocessing for DOCX
- Integrated `mammoth.js` to extract raw text from `.docx` files.  
- The extracted text is now passed directly in the prompt instead of sending the file.

#### 2. Updated TXT handling
- TXT files are read using `File.text()` and sent as plain text.  
- This avoids the previous issue where TXT uploads were treated as binary data by the AI model.

#### 3. Updated UI & validation
- Removed unsupported `.doc` from the file input.  
- Updated user instructions to reflect supported formats: **PDF, DOCX, TXT**.

#### 4. Added Conventional Comments
Following the project's commenting standards:
- `track:` — documents new logic flow  
- `compare:` — highlights differences from the original implementation  
- `todo:` — added for potential future `.doc` support  

#### 5. Preserved original implementation
- The old upload-based approach remains in the file, commented out and annotated with `compare(previous-implementation)` for reference.

---

### Notes
- `.doc` support was intentionally not implemented. Parsing `.doc` requires heavy binary parsing or large WASM dependencies (e.g., LibreOffice WASM ~80–100MB), which would significantly increase the size and complexity of this simple documentation example.
- A `todo:` comment has been added to indicate where `.doc` support could be added in the future.

---

### Testing
- Verified that all three supported formats (**PDF, DOCX, TXT**) work correctly in the Docs Playground.  
- Confirmed **no regressions** for Appspace.  
- Streaming responses remain fully functional.

---

### Issue Reference
Fixes **#2342**

### DIFF
--- a/src/docs/src/playground/examples/ai-resume-analyzer.html
+++ b/src/docs/src/playground/examples/ai-resume-analyzer.html
@@ -92,11 +92,6 @@
 
     
     <script>
-       // track: file-processing
-       // The AI model (claude-sonnet-4) only supports native PDF parsing.
-       // compare(original): original implementation attempted to upload DOC/DOCX/TXT directly.
-       // We now extract DOCX/TXT to text before analysis because the model treats them as binary.
-        
         let uploadedFile = null;
         let extractedText = null;
         
@@ -143,11 +138,6 @@
 
         async function processFile(file) 
         {
-            // track: file-type-handling
-            // PDF => upload directly (native support).
-            // DOCX => extracted via Mammoth (AI cannot parse DOCX binaries).
-            // TXT => read text because AI cannot parse raw .txt file uploads.
-
             if (!file) return;
             
             uploadedFile = file;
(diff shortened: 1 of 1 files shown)


reynaldichernando:
hi @Devansh-18 , thanks for the PR! i can confirm the example is working 🎉 
i made minor adjustments which is to cleanup the comments as well as updating the model to haiku 4.5, since it's faster and cheaper

PR #2365, opened 2026-01-28
docs: add example for streaming with function calling

### LINKED ISSUE #2360: Add example of function call with streaming mode
Right now in our docs, the example for function call uses non streaming mode
https://docs.puter.com/AI/chat/#function-calling

We would like to add example to use it in streaming mode

this should be possible given all the providers behind the scenes we use also support it

reference of function call with streaming

- https://openrouter.ai/docs/guides/features/tool-calling#streaming-with-tool-calls
- https://platform.openai.com/docs/guides/function-calling#streaming
- https://platform.claude.com/docs/en/build-with-claude/streaming#streaming-request-with-tool-use

AryaXDG:
## Description
This PR adds a missing example to the `puter.ai.chat()` documentation demonstrating how to handle **Function Calling** when `stream: true` is enabled.

Fixes #2360

## Context
Currently, the docs cover "Function Calling" and "Streaming" separately. Developers trying to combine them might assume they need to manually accumulate partial JSON strings (standard behavior for OpenAI/Anthropic SDKs).

However, after analyzing `services/ai/utils/Streaming.js` in the backend, I confirmed that `AIChatToolUseStream` buffers and parses the tool call internally. It only emits a chunk to the client when the tool call is complete.

## Changes
- Updated `src/docs/src/AI/chat.md`.
- Added a new code example that:
  1. Requests a chat completion with `{ stream: true, tools: [...] }`.
  2. Iterates through the stream.
  3. Listens for the `part.type === 'tool_use'` event.
  4. Accesses the fully parsed `part.input` object directly (simplifying the client-side logic).

### DIFF
--- a/src/docs/src/AI/chat.md
+++ b/src/docs/src/AI/chat.md
@@ -270,6 +270,88 @@ List of OpenAI models that support the web search can be found in their [API com
 </html>
 ```
 
+<strong class="example-title">Streaming Function Calling</strong>
+
+```html;ai-streaming-function-calling
+<html>
+<body>
+    <script src="https://js.puter.com/v2/"></script>
+    <script>
+        // Define the tool
+        const tools = [{
+            type: "function",
+            function: {
+                name: "get_weather",
+                description: "Get current weather for a location",
+                parameters: {
+                    type: "object",
+                    properties: {
+                        location: { type: "string", description: "City name" }
+                    },
+                    required: ["location"]
+                }
+            }
+        }];
+
+        // Mock weather function
+        function getWeather(location) {
+            return `The weather in ${location} is 22°C and Sunny.`;
+        }
+
+        (async () => {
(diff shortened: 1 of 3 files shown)


reynaldichernando:
hi @AryaXDG , thank you for your PR! i can confirm your function call example works with streaming!

i pushed a small change fixing minor issue in the PR
- minor adjust example text
- add missing code block closing
- move the example up, closer to the other function call

i also added the example to playground, in order to do this, we need to:
- add the example entry in `src/examples.js`
- as well as adding the code html to `src/playground/examples/**`

no worries about this, just something to look out for in the future

looking forward to more contributions!

AryaXDG:
@reynaldichernando Thanks for the review and the quick fixes! I appreciate the heads-up about the Playground files, I'll definitely keep that in mind for the future contributions :)

PR #2366, opened 2026-01-28
Add installation instructions for Debian and Ubuntu

stuffbymax:
Added installation requirements for Debian and Ubuntu.

i could make requirements.md or just add the requirements in first-run-issues.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,18 @@ Puter is an advanced, open-source internet operating system designed to be featu
 
 ## Getting Started
 
+### install requirements 
+
+#### (Debian, Ubuntu)
+
+```bash
+sudo apt install npm git curl
+
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.3/install.sh | bash
+source ~/.bashrc
+nvm install 24
+```
+
 ### 💻 Local Development
 
 ```bash


KernelDeimos:
We had specific instructions for different Linux distributions. Were they removed? I don't want to merge this until I know what happened.

stuffbymax:
> We had specific instructions for different Linux distributions. Were they removed? I don't want to merge this until I know what happened.

sorry for the late response. 
i coudnt find any instructions how to install requirements so shoud i make a md requiremets md page?

KernelDeimos:
I found what I was looking for: https://github.com/HeyPuter/puter/blob/eead0fdfa956e0edb0904c1adc0051c6a031e7f3/doc/self-hosters/first-run-issues.md?plain=1

First-run issue specific to Debian and Ubuntu (such as the correct node version not being installed) should be noted there. Obviously this documentation is too difficult to find, so we need a way to make it easier to find when it's needed.

stuffbymax:
> I found what I was looking for: https://github.com/HeyPuter/puter/blob/eead0fdfa956e0edb0904c1adc0051c6a031e7f3/doc/self-hosters/first-run-issues.md?plain=1
> 
> First-run issue specific to Debian and Ubuntu (such as the correct node version not being installed) should be noted there. Obviously this documentation is too difficult to find, so we need a way to make it easier to find when it's needed.

i made new doc called install.md as new pull request

https://github.com/HeyPuter/puter/pull/2405

and i am goanna close this one

PR #2367, opened 2026-01-28
fix: app-under-user access token ACL

KernelDeimos:
There were a couple issues with ACL where access would be denied to app tokens within access tokens, caused by:
- incorrect recursion: only `false` was considered when in fact the result from the recursive call is appropriate
- files without a known path are incorrectly handled, which can happen when iterating over parent directories


KernelDeimos:
Backend tests are currently broken from 89c49902b6951007f648ccca8ae6d12768980750 so I'm merging this because the test errors are still consistent

PR #2377, opened 2026-01-29
Improve docs search with fuzzy matching and smart navigation

### LINKED ISSUE #2362: Docs search improvement

https://docs.puter.com/

satyansh2004:
This PR improves the docs search UX by replacing exact string matching with
Fuse.js-based fuzzy search.

- Supports typo-tolerant and similar-word matching
- Preserves existing text-fragment navigation for exact matches
- Falls back to nearest page/section for fuzzy matches
- No backend changes; minimal client-side impact

Fixes #2362

### DIFF
--- a/src/docs/build.js
+++ b/src/docs/build.js
@@ -263,6 +263,8 @@ function generateDocsHTML (filePath, rootDir, page, isIndex = false) {
     // create the HTML file
     html += '<head>';
     html += '<meta charset="utf-8">';
+    html += '<script src="https://cdn.jsdelivr.net/npm/fuse.js@6.6.2"></script>';
+
     // Title
     if ( isIndex ) {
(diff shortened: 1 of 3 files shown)


reynaldichernando:
hi @satyansh2004 , thank you for the PR
let me test this thoroughly first, since its a major change

satyansh2004:
Thanks for reviewing this. Happy to make any changes if you find issues during testing.

satyansh2004:
ok, I will resolve these issues asyou described here as soon as possible

satyansh2004:
Restarting this PR with a cleaner implementation.

I'll open a new PR that fixes:

* text fragment matching
* fuzzy search navigation
* autocomplete highlight

Closing this one to keep the review simple.

PR #2418, opened 2026-02-05
Document getAuthToken for web based login

reynaldichernando:
- document getAuthToken
- get rid of the beta notice for node.js


reynaldichernando:
justification for removing beta notice
this was before we have all the framework integrations, but now since its more mature, we'll just remove it

PR #2426, opened 2026-02-06
Fix opus 4.6 spec

### LINKED ISSUE #2425: Inconsistent opus 4.6 context window and max tokens
hi team, seems like the info we have for opus 4.6 direct integration is inconsistent with their official announcement
https://github.com/HeyPuter/puter/blob/15bd2ec5ac3ecc5f6b372e2bb98a97ef6fdb6479/src/backend/src/services/ai/chat/providers/ClaudeProvider/models.ts#L20-L21

https://www.anthropic.com/news/claude-opus-4-6

- "1M token context (beta). Opus 4.6 is our first Opus-class model with 1M token context. Premium pricing applies for prompts exceeding 200k tokens ($10/$37.50 per million input/output tokens)."
- "128k output tokens. Opus 4.6 supports outputs of up to 128k tokens, which lets Claude complete larger-output tasks without breaking them into multiple requests."

it should be
```
context: 1000000,
max_tokens: 128000,
```

although to note, context window is a bit weird here, they have different pricing exceeding 200k token (?)

reynaldichernando:
Fixes #2425

### DIFF
--- a/src/backend/src/services/ai/chat/providers/ClaudeProvider/models.ts
+++ b/src/backend/src/services/ai/chat/providers/ClaudeProvider/models.ts
@@ -17,8 +17,8 @@ export const CLAUDE_MODELS: IChatModel[] = [
             cache_read_input_tokens: 500 * 0.1,
             output_tokens: 2500,
         },
-        context: 200000,
-        max_tokens: 64000,
+        context: 1000000,
+        max_tokens: 128000,
     },
     {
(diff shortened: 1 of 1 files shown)


ProgrammerIn-wonderland:
This is already done except for the context length thing which I decided not to do in accordance to what we did with other models regarding split length billing

Thanks for the PR though, I just didn't see this earlier

Issue #2430, opened 2026-02-06
Collapse playground menu when clicking link on mobile


reynaldichernando:
when playground (https://docs.puter.com/playground/) is in mobile mode, clicking on any of the menu item doesn't collapse the menu, requiring user to click the menu toggle again

we can improve this UX by always auto collapsing the menu when user clicks on any link while in mobile mode

sanjeevafk:
Hi @reynaldichernando, would love to work on this!

reynaldichernando:
assigned! @voidcommit-afk

sanjeevafk:
hi @reynaldichernando 
just opened a PR, pls check 
https://github.com/HeyPuter/puter/pull/2431

reynaldichernando:
merged! thank you for your contribution

sanjeevafk:
sure! would love to help out more, thank you!

PR #2431, opened 2026-02-06
fix: auto-collapse playground menu on mobile (#2430)

sanjeevafk:
This PR improves the user experience of the Puter.js Playground on mobile devices. Previously, when a user selected an example from the sidebar menu in mobile view, the menu would remain open and cover the editor, requiring the user to manually click the toggle button to close it.

The sidebar now automatically collapses once a link is clicked, allowing the user to see the selected example immediately.

### Changes

*   **UX Improvement**: Added a check in app.js to automatically collapse the sidebar when a .sidebar-item is clicked if the window width is 768px or less.
    
*   **Code Polish**:
    
    *   Added /\* global \*/ comments to avoid linting errors for external libraries like Monaco and Clarity.
        
    *   Simplified popstate handling by using window.location.reload() instead of re-assigning window.location.href.
        
    *   Removed an unused variable (code) in app.js.

### demo vid of the changes made

https://github.com/user-attachments/assets/d1352608-8027-49fb-8ddb-226bbb7d9c37

### DIFF
--- a/src/docs/src/playground/assets/js/app.js
+++ b/src/docs/src/playground/assets/js/app.js
@@ -87,7 +87,6 @@ document.addEventListener('keydown', function (e) {
     }
 });
 
-document.getElementById('code');
 var run = document.getElementById('run');
 run.addEventListener('click', function () {
     loadStringInIframe(editor.getValue());


src/docs/src/playground/assets/js/app.js
```diff
@@ -86,7 +87,7 @@ document.addEventListener('keydown', function (e) {
     }
 });
 
-var code = document.getElementById('code');
+document.getElementById('code');
```
reynaldichernando: nice catch!

let's remove this line entirely

reynaldichernando:
hi @voidcommit-afk , i left a small comment, but otherwise this PR is good!
could you help take a look at the comment, and afterwards we can merge this

sanjeevafk:
hi @reynaldichernando 
done, removed the said line! thanks for catching it

PR #2446, opened 2026-02-08
Add shell reload instructions for nvm setup

stuffbymax:
Added instructions to reload the shell after loading nvm.  i forgot to add it because i am stupid


ProgrammerIn-wonderland:
Good idea

PR #2460, opened 2026-02-11
Login/Signup with OIDC

KernelDeimos:
**let's not merge on a Friday night**, _but_ this is ready for review.

These changes implement login and signup for OIDC flows, as well as consequential changes because of how these changes disrupt other parts of authentication, such as:
- user protected endpoint need a new re-auth flow for OIDC-signed-up users
- GUI tokens and http-only tokens need to be separated to support redirect flows safely


src/backend/src/api/APIError.js
```diff
@@ -459,6 +463,10 @@ class APIError {
             status: 403,
             message: 'Password does not match.',
         },
+        'oidc_revalidation_required': {
```
Salazareo, line 472: are these errors handled nicely now on front end, as in not showing the whole error class?

src/backend/src/middleware/configurable_auth.js
```diff
@@ -47,23 +48,26 @@ const configurable_auth = options => async (req, res, next) => {
     const optional = options?.optional;
 
     // Request might already have been authed (PreAuthService)
-    if ( req.actor ) next();
+    if ( req.actor ) return next();
```
Salazareo, line 51: nit: returning void methods feels a bit weird to me, just cuz could change if signature for next (not that it will) changes, can we do {next(); return;}

extensions/whoami/routes.js
```diff
@@ -88,6 +89,20 @@ extension.get('/whoami', { subdomain: 'api' }, async (req, res, next) => {
...
         desktop_bg_fit: req.user.desktop_bg_fit,
         is_temp: (req.user.password === null && req.user.email === null),
+        oidc_only,
+        ...(oidc_only ? await (async () => {
+            try {
+                const svc_oidc = req.services.get('oidc');
+                const providers = await svc_oidc.getEnabledProviderIds();
+                const origin = (svc_oidc.global_config?.origin || '').replace(/\/$/, '');
+                const provider = providers && providers[0];
+                return provider
```
Salazareo: nit: can we just do if blocks for clarity?

src/backend/src/middleware/configurable_auth.js
Salazareo: oh but good catch here, this likely was executing even after next was called yikes

src/backend/src/middleware/configurable_auth.js
```diff
@@ -110,6 +117,20 @@ const configurable_auth = options => async (req, res, next) => {
     const services = context.get('services');
     const svc_auth = services.get('auth');
 
+    // Debug: log token source and decoded type before creating Actor (for session_required / hasHttpPowers debugging)
+    if ( process.env.DEBUG ) {
+        let decodedForLog;
+        try {
+            decodedForLog = jwt.decode(token);
+        } catch ( _ ) { /* ignore */ }
+        console.log('decodedForLog?', decodedForLog);
```
Salazareo: please use debug for debug logs

also heads up that decoded jwt even if behind env flag might be bad

src/backend/src/modules/web/WebServerService.js
```diff
@@ -337,6 +337,34 @@ class WebServerService extends BaseService {
...
+                    actor.type.user,
+                    actor.type.session,
+                );
+                res.cookie(config.cookie_name, session_token, {
+                    sameSite: 'none',
+                    secure: true,
+                    httpOnly: true,
+                });
+            } catch ( _e ) {
+                // Invalid or expired token; do not set cookie
```
Salazareo: plz add an info log so we can see this in case of an attack vector

src/backend/src/modules/web/WebServerService.js
```diff
@@ -627,7 +655,7 @@ class WebServerService extends BaseService {
                 req.co_isolation_enabled
                 ;
 
-            if ( req.path === '/signup' || req.path === '/login' || req.path.startsWith('/extensions/') ) {
+            if ( req.path === '/signup' || req.path === '/login' || req.path.startsWith('/extensions/') || req.path.startsWith('/auth/oidc') ) {
```
Salazareo, line 659: nit: might be worthwhile pulling these into a helper method?

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+const config = require('../../config');
+const jwt = require('jsonwebtoken');
+const { get_user } = require('../../helpers');
+
+const REVALIDATION_COOKIE_NAME = 'puter_revalidation';
+const REVALIDATION_EXPIRY_SEC = 300; // 5 minutes
+
+/** If Accept includes text/html, set session cookie and redirect to app; otherwise send JSON. */
+const finishOidcSuccess_ = async (req, res, user, stateDecoded) => {
+    console.log('okay finishOidSuccess_ is happening');
```
Salazareo: plz debug with nice messages if debug logs needed

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+    const svc_auth = req.services.get('auth');
+    const { session, token: session_token } = await svc_auth.create_session_token(user, { req });
+    res.cookie(config.cookie_name, session_token, {
+        sameSite: 'none',
+        secure: true,
+        httpOnly: true,
+    });
+    console.log('what are these values?', {
```
Salazareo: same as above

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+        stateDecoded,
+    });
+    let target = stateDecoded.redirect_uri || config.origin || '/';
+    const origin = config.origin || '';
+    console.log('okay what\'s target though?', { target, origin });
```
Salazareo: here too

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+const router = new express.Router();
+const config = require('../../config');
+const jwt = require('jsonwebtoken');
+const { get_user } = require('../../helpers');
+
+const REVALIDATION_COOKIE_NAME = 'puter_revalidation';
+const REVALIDATION_EXPIRY_SEC = 300; // 5 minutes
+
+/** If Accept includes text/html, set session cookie and redirect to app; otherwise send JSON. */
+const finishOidcSuccess_ = async (req, res, user, stateDecoded) => {
```
Salazareo, line 41: can we make these take just needed args and return just the cookie and target if needed

we should let router method handle routing and respoding. avoids this acting and looking like a middleware

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+    const origin = config.origin || '';
+    console.log('okay what\'s target though?', { target, origin });
+    if ( target && origin && !target.startsWith(origin) ) {
+        target = origin;
+    }
+    return res.redirect(302, target);
+};
+
+/** Exchange code for tokens, get userinfo; returns { provider, userinfo, stateDecoded } or sends error and returns null. */
+const oidcCallbackPreamble_ = async (req, res, callbackRedirectUri) => {
```
Salazareo: same comment regarding having this in a class and as just do business logic, not actual routing

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+    if ( !userinfo || !userinfo.sub ) {
+        res.status(401).send('Could not get user info.');
+        return null;
+    }
+    return { provider, userinfo, stateDecoded };
+};
+
+// GET /auth/oidc/providers - list enabled provider ids for frontend
+router.get('/auth/oidc/providers', async (req, res) => {
+    if ( require('../../helpers').subdomain(req) !== 'api' ) {
```
Salazareo: esm please, and top level imports if not needed to be lazy imported

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+    if ( ! cfg ) {
+        return res.status(404).send('Provider not configured.');
+    }
+    const flow = req.query.flow ? String(req.query.flow) : undefined;
+    const flowRedirects = {
+        login: config.origin || '/',
+        signup: config.origin || '/',
+        revalidate: `${(config.origin || '').replace(/\/$/, '')}/auth/revalidate-done`,
+    };
+    const appRedirectUri = (flow && flowRedirects[flow]) ? flowRedirects[flow] : (config.origin || '/');
```
Salazareo, line 107: can we if block this (in an iife if you want), those parens could cause problems

src/backend/src/routers/auth/oidc.js
Salazareo: also nit: can we group these methods as private helpers in a class?

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+        return res.status(429).send('Too many requests.');
+    }
+    const svc_oidc = req.services.get('oidc');
+    const callbackRedirectUri = svc_oidc.getCallbackUrlForFlow('login');
+    const preamble = await oidcCallbackPreamble_(req, res, callbackRedirectUri);
+    if ( ! preamble ) return;
+    const { provider, userinfo, stateDecoded } = preamble;
+    const user = await svc_oidc.findUserByProviderSub(provider, userinfo.sub);
+    if ( ! user ) {
+        return res.status(400).send('No account found. Sign up first.');
```
Salazareo, line 144: probably should be a 404 no?

also, potentially, could we just make this case trigger signup flow instead? would ease friction of people not knowing if/if not signed up

src/backend/src/routers/auth/oidc.js
Salazareo: also nit: preamble feels a bit off as a name for this, is this not the actual code validation for the oauth provider?

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+    if ( existingUser ) {
+        return res.status(400).send('Account already exists. Log in instead.');
+    }
+    const outcome = await svc_oidc.createUserFromOIDC(provider, userinfo);
+    if ( outcome.failed ) {
+        console.log('it looks like the outcome failed...');
+        return res.status(400).send(outcome.userMessage);
+    }
+    const user = await get_user({ id: outcome.infoObject.user_id });
+    console.log('got user????', user);
```
Salazareo: remove plz

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+    const preamble = await oidcCallbackPreamble_(req, res, callbackRedirectUri);
+    if ( ! preamble ) return;
+    const { provider, userinfo, stateDecoded } = preamble;
+    const existingUser = await svc_oidc.findUserByProviderSub(provider, userinfo.sub);
+    if ( existingUser ) {
+        return res.status(400).send('Account already exists. Log in instead.');
+    }
+    const outcome = await svc_oidc.createUserFromOIDC(provider, userinfo);
+    if ( outcome.failed ) {
+        console.log('it looks like the outcome failed...');
```
Salazareo: remove please

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+        return res.status(429).send('Too many requests.');
+    }
+    const svc_oidc = req.services.get('oidc');
+    const callbackRedirectUri = svc_oidc.getCallbackUrlForFlow('signup');
+    const preamble = await oidcCallbackPreamble_(req, res, callbackRedirectUri);
+    if ( ! preamble ) return;
+    const { provider, userinfo, stateDecoded } = preamble;
+    const existingUser = await svc_oidc.findUserByProviderSub(provider, userinfo.sub);
+    if ( existingUser ) {
+        return res.status(400).send('Account already exists. Log in instead.');
```
Salazareo, line 177: can we not just log them in?

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+    const callbackRedirectUri = svc_oidc.getCallbackUrlForFlow('revalidate');
+    const preamble = await oidcCallbackPreamble_(req, res, callbackRedirectUri);
+    if ( ! preamble ) return;
+    const { provider, userinfo, stateDecoded } = preamble;
+    if ( stateDecoded.flow !== 'revalidate' || stateDecoded.user_id == null ) {
+        return res.status(400).send('Invalid revalidate state.');
+    }
+    const user = await svc_oidc.findUserByProviderSub(provider, userinfo.sub);
+    if ( ! user ) {
+        return res.status(400).send('No account found.');
```
Salazareo, line 215: same as before, can this flow into signup in these cases? or when would this even happen?

src/backend/src/routers/auth/oidc.js
```diff
@@ -0,0 +1,238 @@
...
+        httpOnly: true,
+        maxAge: REVALIDATION_EXPIRY_SEC * 1000,
+        path: '/',
+    });
+    const target = stateDecoded.redirect_uri || `${(config.origin || '').replace(/\/$/, '')}/auth/revalidate-done`;
+    return res.redirect(302, target);
+});
+
+// GET /auth/revalidate-done - landing page after OIDC revalidate; posts to opener and closes (for popup flow).
+router.get('/auth/revalidate-done', (req, res) => {
```
Salazareo, line 235: this feels to me, like it should belong in the GUI code, but I could be wrong

src/backend/src/routers/logout.js
```diff
@@ -51,7 +51,9 @@ router.post('/logout', auth, express.json(), async (req, res, next) => {
     //---------------------------------------------------------
     // DANGER ZONE: delete temp user and all its data
     //---------------------------------------------------------
+    console.log('wait... what are these?', req.user.password, req.user.email);
```
Salazareo: please no logs, definitely don't log passwords

src/backend/src/routers/logout.js
```diff
@@ -51,7 +51,9 @@ router.post('/logout', auth, express.json(), async (req, res, next) => {
...
     if ( req.user.password === null && req.user.email === null ) {
+        console.log('ACTUALLY DELETING A USER');
```
Salazareo: please make log more informational

src/backend/src/routers/signup_create_new_user.js
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Create a new user for signup. Common behavior shared by POST /signup and OIDC signup.
+ * Form-signup path is still handled in signup.js; this handles OIDC and will support form signup after refactor.
+ *
+ * @param {object} services - Backend services (from req.services)
+ * @param {object} options - Creation options. For OIDC: { providerId, userinfo }. For form signup: TBD (to be refactored from signup.js).
+ * @returns {Promise<object|null>} The created user, or null on failure (e.g. email already registered).
+ */
+async function signup_create_new_user (services, options) {
```
Salazareo, line 32: feels like these should be in like, UserService or AuthService maybe?

src/backend/src/services/auth/Actor.d.ts
```diff
@@ -12,8 +12,10 @@ export class SystemActorType {
 }
 
 export class UserActorType {
-    constructor (params: { user: IUser });
+    constructor (params: { user: IUser; session?: { uuid: string }; hasHttpPowers?: boolean });
     user: IUser;
+    /** When true, this actor can access user-protected HTTP endpoints (e.g. change password). GUI tokens set this false. */
+    hasHttpPowers: boolean;
```
Salazareo: might need a better name,  httpPowers is a bit vague

maybe 

canAccessAuthEndpoints? idk something a bit clearer

src/backend/src/services/auth/OIDCService.js
```diff
@@ -0,0 +1,255 @@
...
+    return username;
+}
+
+/**
+ * OIDC/OAuth2 service for sign-in with Google (and extensible to other providers).
+ * Uses config.oidc.providers only; no environment variables.
+ */
+export class OIDCService extends BaseService {
+    static MODULES = {
+        jwt,
```
Salazareo: can we just use the import, we can mock for tests, and this adds more mental stack overhead. also removes typehinting when doing this.modules....

src/backend/src/services/auth/OIDCService.js
```diff
@@ -0,0 +1,255 @@
...
+        if ( raw.authorization_endpoint && raw.token_endpoint && raw.userinfo_endpoint ) {
+            return {
+                ...raw,
+                scopes: raw.scopes ?? 'openid email profile',
+            };
+        }
+        return null;
+    }
+
+    async _getGoogleDiscovery_ () {
```
Salazareo: is this private? can just be #getGoogleDiscovery,

src/backend/src/services/auth/OIDCService.js
```diff
@@ -0,0 +1,255 @@
...
+ */
+export class OIDCService extends BaseService {
+    static MODULES = {
+        jwt,
+    };
+
+    async _init () {
+        this.db = await this.services.get('database').get(DB_WRITE, 'auth');
+        this.providers = this.config.providers ?? {};
+        this._googleDiscovery = null;
```
Salazareo: is this private? can just be #googleDiscovery

src/backend/src/services/abuse-prevention/EdgeRateLimitService.js
```diff
@@ -104,6 +104,10 @@ class EdgeRateLimitService extends BaseService {
                 limit: 10,
                 window: HOUR,
             },
+            ['/user-protected/change-username']: {
```
Salazareo: nit: string keys dont need []

src/backend/src/services/auth/OIDCService.js
```diff
@@ -0,0 +1,255 @@
...
+'use strict';
+import jwt from 'jsonwebtoken';
+import { username_exists } from '../../helpers.js';
+import { generate_identifier } from '../../util/identifier.js';
+import { OutcomeObject } from '../../util/outcomeutil.js';
+import BaseService from '../BaseService.js';
+import { DB_WRITE } from '../database/consts.js';
+import { CreatedUserOutcome } from './SignupService.js';
+
+const GOOGLE_DISCOVERY_URL = 'https://accounts.google.com/.well-known/openid-configuration';
```
Salazareo, line 28: these are probably best to come from a configuration file.

at the very least a const map of different configs in a separate file

anytime we wanna add new ones we'll need to add a new set of code here for endpoints, scopes, etc.

I'd recommend doing a config like {

[providerName]: {
relevant fields
}

src/backend/src/services/auth/OIDCService.js
Salazareo: also should again probably be more generic.

otherwise this is 

`GoogleOIDCProvider` which is ok if we want interfaces, but more hardcode-y

src/backend/src/services/auth/OIDCService.js
```diff
@@ -0,0 +1,255 @@
...
+        return await svc_get_user.get_user({ id: rows[0].user_id, cached: false });
+    }
+
+    /**
+     * Link an existing Puter user to an OIDC provider identity.
+     */
+    async linkProviderToUser (userId, providerId, providerSub, refreshToken = null) {
+        try {
+            await this.db.write('INSERT INTO user_oidc_providers (user_id, provider, provider_sub, refresh_token) VALUES (?, ?, ?, ?)',
+                            [userId, providerId, providerSub, refreshToken]);
```
Salazareo, line 204: what does providerSub mean? subject?

src/backend/src/services/auth/SignupService.js
```diff
@@ -0,0 +1,276 @@
...
+                return outcome.fail(
+                    'User signup is disabled.',
+                    'signup.user_signup_disabled',
+                );
+            }
+        }
+
+        // Emit the `puter.signup` event
+        // NOTICE: conditional early return
+        {
```
Salazareo, line 94: nit: not a fan of these blocks, if its distinct enough logic it should probably be in a helper method, if its not, then don't see a point much

src/backend/src/services/auth/SignupService.js
```diff
@@ -0,0 +1,276 @@
...
+
+            const user_id = insert_res.insertId;
+            outcome.infoObject.user_id = user_id;
+
+            const [user] = await db.pread(
+                'SELECT * FROM `user` WHERE `id` = ? LIMIT 1',
+                [user_id]);
+
+            // TODO(???): should user login happen here or by caller?
+            {
```
Salazareo, line 245: we should clean up commented out code like this, if its relevant, then remove comments

src/backend/src/services/database/BaseDatabaseAccessService.js
```diff
@@ -58,7 +58,7 @@ class BaseDatabaseAccessService extends BaseService {
     *
     * @returns {BaseDatabaseAccessService} The current instance of the service.
     */
-    get () {
+    get (_accessLevel, _scope) {
```
Salazareo, line 61: ? whats this for?

src/backend/src/services/repositories/DDBClientWrapper.js
Salazareo, line 1: please remove these ghost files, likely just remnants from folder move or renames

src/backend/src/services/web/UserProtectedEndpointsService.js
```diff
@@ -24,6 +24,19 @@ const { UserActorType } = require('../auth/Actor');
 const { Endpoint } = require('../../util/expressutil');
 const APIError = require('../../api/APIError.js');
 const configurable_auth = require('../../middleware/configurable_auth.js');
+const config = require('../../config');
+const jwt = require('jsonwebtoken');
+
+const REVALIDATION_COOKIE_NAME = 'puter_revalidation';
+
+async function revalidateUrlFields_ (svc, req, user) {
```
Salazareo: if this is meant to be a private method for the class, should probably belong in class as private method.  (I'm assuming thats what _ suffix implies still, but not sure)

src/backend/src/util/outcomeutil.js
```diff
@@ -0,0 +1,32 @@
+export class OutcomeObject {
```
Salazareo: ts transpile file?

src/backend/src/util/outcomeutil.ts
```diff
@@ -0,0 +1,77 @@
+/**
+ * Represents the outcome of a task that might fail or succeed.
+ */
+export class OutcomeObject<T> {
```
Salazareo, line 4: make sure to git ignore the file output, for now easiest way is to put in a folder and git ignore *.js in a file there

src/gui/src/UI/UIWindowChangePassword.js
```diff
@@ -84,59 +104,152 @@ async function UIWindowChangePassword (options) {
...
+            await myOpenRevalidatePopup(data.revalidate_url);
+            const r = await doSubmit();
+            if ( r.ok ) onSuccess();
+            else r.json().then((d) => onError(d.message || 'Request failed')).catch(() => onError('Request failed'));
+            return;
+        }
+        onError(data.message || res.statusText || 'Request failed');
+    });
+
+    // function openRevalidatePopup (revalidateUrl, onDone) {
```
Salazareo: nit: avoid leaving commented out code like this

src/backend/src/util/outcomeutil.ts
Salazareo, line 4: do we actually need this at all? can this not just be an interface we return through an object?

src/backend/src/services/auth/OIDCService.js
```diff
@@ -0,0 +1,255 @@
...
+        const svc_signup = this.services.get('signup');
+        const outcome = await svc_signup.create_new_user({
+            username: await generate_random_username(),
+            email: claims?.email ?? null,
+            password: null,
+            oidc_only: true,
+            assume_email_ownership: true,
+        });
+        const { user_id } = outcome.infoObject;
+        if ( outcome.success ) {
```
Salazareo: success is a method I'm pretty sure, don't think this would work properly.

also again, should outcome be its own class? it can just be a defined interface it returns as a json object

Salazareo (changes requested):
# Artisan, handmade, human review
So a lot of these are more stylistic and code structure changes.

however, please do remove noisy logs, bits that make the code footprint larger ( commented out code, pseudo-blocks and boilerplate classes if they can be avoided) and avoid the this.modules.* syntax from before as it breaks support from language servers.

aside from that. 2 things to note:

- that object outcome success looks suspect

- lets avoid hard coded X provider configs, like `Google_Scope`, and such, ideally we make this class follow spec and just pull config as needed from some storage, whether it be live or via config redeploy.

lastly I had some concerns about the change with gui token and http-only token.

fully in favour of moving to http-only tokens, and giving granular restrictions to tokens, but we'll need to verify to make sure existing features (like switching acocunts from gui) still function with that.


Additionally, and I'll say to take this with a grain of salt, but running codex 5.3 on this flagged some things too. Some I already mentioned in my artisan, human review, but some others I missed and definitely worth looking at:

# LLM slop review

## Findings (ordered by severity)

1. **Critical**: OIDC signup linking uses `outcome.success` incorrectly, so the link step always executes.
`src/backend/src/services/auth/OIDCService.js:237`
`OutcomeObject` exposes a `success()` method and a `failed` flag. Checking `if (outcome.success)` is always truthy (function reference), so `linkProviderToUser(...)` runs even for failed outcomes. That can attempt linking with `user_id = null`, turning expected signup validation failures into DB errors/500s.

2. **High**: Change password request payload is broken for non-OIDC path and can throw in OIDC retry path.
`src/gui/src/UI/UIWindowChangePassword.js:174`
`doSubmit` expects an object (`{ new_password, current_password }`), but code passes a string (`doSubmit(current_password)`), so both fields serialize as `undefined`. Also `doSubmit()` is called with no args on revalidation retry (`src/gui/src/UI/UIWindowChangePassword.js:183`), which throws due destructuring from `undefined`. This can break password change for normal users and OIDC users.

3. **High**: Change email OIDC retry can throw at runtime.
`src/gui/src/UI/Settings/UIWindowChangeEmail.js:167`
`doSubmit` destructures `{ new_email, password }`, but retry path calls `doSubmit()` with no argument. That throws before the request is sent, so OIDC revalidation flow can fail after popup.

4. **Medium**: Username revalidation flow does not await popup completion before retry.
`src/gui/src/UI/UIWindowChangeUsername.js:131`
and `src/gui/src/UI/UIWindowChangeUsername.js:151`
`myOpenRevalidatePopup(...)` is invoked without `await`, and a protected request is sent immediately afterward. The retry can race before cookie issuance and fail spuriously.

5. **Medium**: Sensitive debug logging was added in auth paths.
Files: `src/backend/src/middleware/configurable_auth.js:126`, `src/backend/src/routers/auth/oidc.js:31`, `src/backend/src/routers/logout.js:54`
These log decoded token data, redirect state, and account password/email nullness to stdout. Even if some are gated by `DEBUG`, this is high-risk auth telemetry and should be removed or replaced with structured/redacted service logs.

6. **Low**: Wrong rate-limit key used for username changes.
`src/backend/src/routers/user-protected/change-username.js:49`
Route checks `svc_edgeRateLimit.check('change-email-start')` instead of a username-specific key. This ties username throttling behavior to the email flow by mistake and undermines intended controls.

## Open Questions / Assumptions

1. I assumed this review target is behavior/correctness/security; I did not run full backend/frontend test suites in this pass.
2. I also noticed some incidental cleanup debt (debug comments, unused imports, duplicated utility additions), but I excluded style-only notes from findings.


# Last notes

This was all static analysis, so we'll need to test this in depth before deploying after any changes.

src/backend/src/util/outcomeutil.ts
KernelDeimos: can't return interfaces, I don't understand this. That said I'd rather focus on the bigger issues

src/backend/src/middleware/configurable_auth.js
KernelDeimos: This is wrapped in `if ( process.env.DEBUG )` - it will not be logged unless this env flag is explicitly set.

src/backend/src/routers/auth/oidc.js
KernelDeimos: looks nicer than most log messages to me, but this one was left by accident

src/backend/src/api/APIError.js
KernelDeimos: I don't believe handling of these errors on frontend is centralized/DRY'd, so it's hard to say. If you saw something like that somewhere it's likely specific to that particular where. In the flows I tested I did not see any "whole error class" error messages.

src/backend/src/middleware/configurable_auth.js
KernelDeimos: This definitely looks odd at first, but I adopted this style from Nariman's earlier route handlers because I noticed it prevents errors that occur when copy-pasting sections of code that call `next()` or `res.send()` - you always take the return with it which is almost certainly what you want.

src/backend/src/modules/web/WebServerService.js
KernelDeimos: Good catch, adding that

src/backend/src/modules/web/WebServerService.js
KernelDeimos: Individually or a helper method that checks all of these? I think that's too much jumping around - the condition isn't that complicated and I think it'll be easier to read the code if we don't extract this.

extensions/whoami/routes.js
KernelDeimos: Where? Can you give an example? GitHub might be showing the wrong section of code, but this is my argument against what I think you're suggesting:

first rewrite:
```javascript
let origin;
if ( svc_oidc.global_config?.origin ) {
    origin = origin.replace(/\/$/, '');
} else {
    origin = '';
}
```
problem: changes a `const` to a `let`, fixable by:
```javascript
cosnt origin = (() => {
    if ( svc_oidc.global_config?.origin ) {
        return origin.replace(/\/$/, '');
    } else {
        return '';
    }
})();
```
problem: unnecessary IIFE - probably more confusing to read for most

src/backend/src/routers/auth/oidc.js
KernelDeimos: refactored as described - your logic here is sound. As far as grouping these methods as private helpers in a class, I don't see how that's necessary or beneficial - I would rather defer that until such a time when that's clear

src/backend/src/routers/auth/oidc.js
KernelDeimos: What kind of problems do you foresee? I find this snippet is very clear - I'm not sure I would make it clearer by trying to address your comment because I'm not seeing what you're seeing.

src/backend/src/routers/auth/oidc.js
KernelDeimos: This is tricky... HTTP semantics are very counter-intuitive. I think `400` is still correct here with the reasoning that "404" doesn't mean "something wasn't found", it means "the requested resource was not found". This however is an application-level thing that isn't found, and this isn't in a RESTful context.

src/backend/src/routers/auth/oidc.js
KernelDeimos: Do you have an alternative name suggestion?

src/backend/src/routers/auth/oidc.js
KernelDeimos: I have a backlog TODO in my notes for this. We have never before logged a user in through the signup flow - I didn't want to introduce this potential complication on the first release.

src/backend/src/routers/auth/oidc.js
KernelDeimos: This is the redirect URL sent to the oauth provider - this is the closest we can get to it being in gui code

src/backend/src/services/auth/Actor.d.ts
KernelDeimos: `canAccessAuthEndpoints` is also inaccurate - technically it's `canRequestUserProtectedEndpoints` but that's very verbose. I'll leave this open for further suggestions.

src/backend/src/services/auth/OIDCService.js
KernelDeimos: Fixed this

src/backend/src/util/outcomeutil.js
KernelDeimos: deleted

src/backend/src/util/outcomeutil.ts
KernelDeimos: Just added the filename to a `.gitignore` - less shotgun surgery of finding out-of-sync imports

extensions/whoami/routes.js
Salazareo: this is a return, no need for let nor IIFE

src/backend/src/api/APIError.js
Salazareo: i meant whether or not the response for these still write as the whole class serialized or not

src/backend/src/middleware/configurable_auth.js
Salazareo: idk if copy-paste-ability of something is a good reason, but its fine to keep as is if you want, mainly an issue if signature changes

src/backend/src/middleware/configurable_auth.js
Salazareo: still use debug logs, just good practice

src/backend/src/routers/auth/oidc.js
Salazareo: verifyOidcCode? not sure, I just don't think preamble is what this is

src/backend/src/routers/auth/oidc.js
Salazareo: if someone changes to 

`flow && flowRedirects[flow] ? flowRedirects[flow] : config.origin || '/'`

is this flow && (flowRedirects[flow] ? flowRedirects[flow] : config.origin || '/')

or the prior one.

src/backend/src/routers/auth/oidc.js
Salazareo: thats fair

src/backend/src/routers/auth/oidc.js
Salazareo: oh in the past I've done it so that the oauth provider response url is handled by the gui.

src/backend/src/services/auth/Actor.d.ts
Salazareo: allowUserProtectedReqs?

src/backend/src/modules/web/WebServerService.js
Salazareo: sure thats valid

src/backend/src/routers/auth/oidc.js
Salazareo: thats fine sure. functionally methods not exported by a file are private methods for anything else in that file

src/backend/src/util/outcomeutil.ts
Salazareo: I just meant it could be just an json object instead of a class, but yeah, was more of a question if it needs to be class or not

src/backend/src/routers/auth/oidc.js
KernelDeimos: Updated

src/backend/src/routers/auth/oidc.js
KernelDeimos: Ternary always has the lower precedence in c-like languages but I agree that it's confusing without the parens. My solution to that confusion would be to add the parens. The parens are here, so I don't know why we're hypothesizing if someone removes them - that sounds like a logical fallacy to me

extensions/whoami/routes.js
KernelDeimos: Not seeing how this would be simpler or easier to read. Can you show me using "Add a suggestion"?

src/backend/src/api/APIError.js
KernelDeimos: I haven't seen this issue while testing - marking this one as resolved

src/backend/src/util/outcomeutil.ts
KernelDeimos: Oh, that makes sense. I'm going to say yes, because that means the state is controlled (invalid states won't happen)

src/backend/src/routers/auth/oidc.js
KernelDeimos: Name changed to `processOIDCCallbackRequest_`

src/backend/src/routers/auth/oidc.js
KernelDeimos: Marking as resolved as in out-of-scope for this PR in particular

src/backend/src/routers/auth/oidc.js
KernelDeimos: Marking as resolved as in out-of-scope for this PR in particular (same reason as previous comment - want to release without this complication initially)

src/backend/src/routers/signup_create_new_user.js
KernelDeimos: I want to leave this here for now because it doesn't seem to _need_ to be in UserService or AuthService. I find it's better to do this kind of refactor when it's clear that it's needed - i.e. as soon as we make it depend on some state in UserService or AuthService, that will give us more information about where it belongs. Might as well not put it in the wrong class from the start.

src/backend/src/services/auth/OIDCService.js
KernelDeimos: Making the method private. This is definitely **OIDCProvider** and it can have a method like this for each provider. When there is more than one provider we can look at the similarities/differences and make it more declarative, but I don't think it's time to do that right now.

src/backend/src/services/auth/OIDCService.js
KernelDeimos: We can hardcode values like this for known providers - if we don't we're adding burden on users when they configure OIDC. I also think this is a premature refactor, so out-of-scope for this PR.

src/backend/src/services/auth/OIDCService.js
KernelDeimos: It stands for subject, but OpenID calls it `sub` everywhere so I think it would be more confusing if it wasn't called `sub` - consistency and familiarity over verbosity in this case.

src/backend/src/services/auth/SignupService.js
KernelDeimos: I disagree. Static analysis of source files is easier and faster when scope limiters are used where possible because you don't need to keep track of symbols where they're not going to be used. Extracting to a separate helper method would require jumping between files to analyze the code unnecessarily and also imply to other developers that this code should be reused, when in fact it's specific to the implementation of `create_new_user`.

src/backend/src/services/auth/SignupService.js
KernelDeimos: This "commented code" is for the sake of example - a placeholder to say "this is where this will go if we decide to implement it". What can I do to make that more clear?

src/backend/src/services/database/BaseDatabaseAccessService.js
KernelDeimos: Implementation of an implied interface for the database service. The `_` prefix is recognized by eslint and other static analysis tools to indicate that the argument is unused. This makes it possible to name the arguments which helps developers and AI agents understand how this method is meant to be used. It may seem pointless to pass access level and scope if we're not using them, but in this case I made the decision many moons ago to pass these parameters to this method because it would be a very painful refactor to add this later on and I can see a lot of reasons we may eventually want to do this. These properties might never be used, but if that happens to be the case it didn't cost us much to have these values just in case.

src/backend/src/services/repositories/DDBClientWrapper.js
KernelDeimos: What are ghost files? Why did this happen? I didn't touch anything under your `repositories` directory.

src/backend/src/routers/auth/oidc.js
KernelDeimos: Name was fixed, decided to keep this where it is for now to avoid a premature refactor.

src/backend/src/services/auth/Actor.d.ts
KernelDeimos: I'm going with `hasHttpOnlyCookie` because it better states what it is rather than what it's for, but is clearer than `hasHttpPowers`

src/backend/src/services/auth/OIDCService.js
KernelDeimos: Yes it is. I made the change.

src/backend/src/services/web/UserProtectedEndpointsService.js
KernelDeimos: Yes it is. I made the change.

src/backend/src/services/abuse-prevention/EdgeRateLimitService.js
KernelDeimos: This should be an eslint rule. I added the eslint rule and applied it elsewhere in backend as well.

src/backend/src/services/auth/OIDCService.js
KernelDeimos: Yep, I removed it.

extensions/whoami/routes.js
Salazareo:
```suggestion
               if (provider) { 
                 return {oidc_revalidate_url: `${origin}/auth/oidc/${provider}/start?flow=revalidate&user_id=${req.user.id}`
               }
               
               return {}
                 
```

src/backend/src/routers/auth/oidc.js
Salazareo: just to avoid people changing the logic and breaking it.

codes not exactly forever static, so building around making maintainability and extendability of the code easier and safer is worthwhile IMO. Otherwise why even build abstractions

src/backend/src/services/auth/OIDCService.js
Salazareo: sure, but means every oatuh provider requires a code change, which was what I wanted to avoid. can stay like this for now until we add more

src/backend/src/services/auth/SignupService.js
Salazareo: you can make it a private helper, but thats fine if, I just don't think this is common practice

src/backend/src/services/auth/SignupService.js
Salazareo: I feel TODO is enough, when/if this is implemented, whoever works on it will implement it as require anyway

extensions/whoami/routes.js
KernelDeimos: I see now, GitHub wasn't showing the entire section of the code you were referring to as I suspected earlier. For the same reason your suggestion won't replace the entire expression; it will only replace thee `return provider` of:
```javascript
return provider
  ? { oidc_revalidate_url: `${origin}/auth/oidc/${provider}/start?flow=revalidate&user_id=${req.user.id}` }
  : {};
```

GitHub allows commenting on a selection of lines of code btw - I would recommend doing that for future PR comments like this one.

I made this change because I agree that it's more clear. I still like the ternary expression better because it's easier to move around the expression as a whole because it's not interlaced with control-flow statements.

Salazareo (approved):

PR #2462, opened 2026-02-11
Worker badge and info

### LINKED ISSUE #2429: Add workers attached info
hi team, right now it's hard to tell which URL a worker is currently being attached

unlike with publish website where we can simply hover over the folder or right click properties to get the information

Image

Image

with workers we have to do this via the dev center

ideally we have some sort of indicator that shows the worker url

miikakuisma:
Adds worker badge to .js files which have been registered as workers.
Hovering on worker badge shows popover with worker URL link. 
File properties modal also shows link to worker URL.

Fixes #2429


jelveh:
I'm going to merge this with a side note. This has a blocking await to retrieve workers list which will be slow if the user has a lot of workers. I don't blame this on the GUI code, and it's something that should be optimized in Puter.js and the backend, so for that reason, LGTM.

PR #2487, opened 2026-02-13
Miika/put 500

### LINKED ISSUE #2429: Add workers attached info

miikakuisma:
Add worker badge on worker JS files and list Worker URL in file properties modal.

Fixes #2429


src/gui/src/helpers/get_html_element_from_options.js
```diff
@@ -36,7 +36,9 @@ const get_html_element_from_options = async function (options) {
     options.immutable = (options.immutable === false || options.immutable === 0 || options.immutable === undefined ? 0 : 1);
     options.sort_container_after_append = (options.sort_container_after_append !== undefined ? options.sort_container_after_append : false);
     const is_shared_with_me = (options.path !== `/${window.user.username}` && !options.path.startsWith(`/${window.user.username}/`));
-
+    const workers = await puter.workers.list();
```
ProgrammerIn-wonderland: shouldn't use puter.workers.list() for one file, as it fetches all workers. You should ideally just use the same fsentry for this information (.workers array) but if you don't have access to that information and must fetch a singular file's workers, use `await puter.fs.stat({path: "~/Desktop/New Worker.js", returnWorkers: true})`

ProgrammerIn-wonderland (changes requested):
Mostly good! But there is one unnecessary call to puter.workers.list() that I'd rather have replaced with a stat

miikakuisma:
These things had been merged already from somewhere else. This PR could be just ignored and closed..

PR #2498, opened 2026-02-16
Fix: When a maximized window gets resized, Puter apps should be resized as well

### LINKED ISSUE #2471: Resizing windows
when a maximized window gets resized, Puter apps should be resized as well

Image

miikakuisma:
Fixes #2471 

When global resize listener is triggered, go through all mazimized windows and re-apply their maximize functionality.


jelveh:
@miikakuisma the problem is not resolved:


https://github.com/user-attachments/assets/0a3bfff9-b396-4c28-be64-7a246092a98e

PR #2499, opened 2026-02-16
Permission Request Modal Update

### LINKED ISSUE #2482: Core UI modal popups design
Image

Permission and authentication popups can easily scare end users.

Right now, the messaging is either too aggressive or too vague. Combined with the fact that the modal design doesn’t match the app they just opened, it creates distrust.

To a new user, it doesn’t feel like “this app is asking for access.”
It feels like “something is asking for access to my entire computer.”

Most new users don’t even know what Puter is yet, or that apps run inside it. So when a system-looking permission dialog suddenly appears, it looks suspicious or unsafe.

Those first few seconds of interaction matter more than anything else. If trust is broken there, users leave.

Developers should be able to:
• customize the wording
• explain why permission is needed
• scope the request clearly
• style the modal to match their app (CSS/theme/branding)

The same problem applies to login and registration screens. Generic, platform-styled auth pages feel disconnected and untrustworthy. They should feel like a natural part of the app experience, not an external system prompt.

Permissions and authentication should feel:
clear, specific, branded, and reassuring — not generic or alarming.

miikakuisma:
Fixes #2482 

To improve permission request modal

1) Use the app's actual display name and icon instead of the raw identifier puter-app-builder-ai
2) Create separate text strings for read and write access requests


miikakuisma:
Some prototyping was done on how developers could provide reasons for access requests in Dev Center. I also tested custom styles for the request modal. However, after talking with @jelveh, it was decided that those modal customizations won't be implemented or further developed at this point.

PR #2501, opened 2026-02-16
fix nano banana pro pricing and resolutions

P3il4:
fixes two issues with current implementation of nano banana pro in togetherai:
1. not converting aspect ratio values into actual width and height parameters for togetherai, resulting in generation errors
2. incorrect pricing for 2K and 4K quality
https://www.together.ai/models/nano-banana-pro
1K and 2K cost 13.4, 4K costs 24. current implementation does 13.51 * quality value

testing:
```js
puter.drivers.call("puter-image-generation", "together-image-generation", "generate", {
  "prompt": "generate an image of a cat",
  "ratio": {
    "w": 21,
    "h": 9
  },
  "quality": "2K",
  "model": "togetherai:google/gemini-3-pro-image"
})
```
should give correct width + height values and correct cost amount. other models should not be affected by this
wasn't able to test this with a real key as i don't have one but logging the request gave correct values.

### DIFF
--- a/src/backend/src/services/ai/image/providers/TogetherImageGenerationProvider/TogetherImageGenerationProvider.ts
+++ b/src/backend/src/services/ai/image/providers/TogetherImageGenerationProvider/TogetherImageGenerationProvider.ts
@@ -104,11 +104,13 @@ export class TogetherImageGenerationProvider implements IImageProvider {
         const isGemini3 = selectedModel.id === 'togetherai:google/gemini-3-pro-image';
 
         let costInMicroCents: number;
+        let usageAmount: number;
         const qualityCostKey = isGemini3 && quality && selectedModel.costs[quality] !== undefined ? quality : undefined;
 
         if ( qualityCostKey ) {
             const centsPerImage = selectedModel.costs[qualityCostKey];
             costInMicroCents = centsPerImage * 1_000_000;
+            usageAmount = 1;
         } else {
             const priceKey = '1MP';
             const centsPerMP = selectedModel.costs[priceKey];
@@ -117,6 +119,7 @@ export class TogetherImageGenerationProvider implements IImageProvider {
             }
             const MP = (ratio.h * ratio.w) / 1_000_000;
             costInMicroCents = centsPerMP * MP * 1_000_000;
+            usageAmount = MP;
         }
 
         const usageType = `${selectedModel.id}:${quality || '1MP'}`;
@@ -145,7 +148,7 @@ export class TogetherImageGenerationProvider implements IImageProvider {
                 throw new Error('Together AI response did not include image data');
             }
 
-            this.#meteringService.incrementUsage(actor, usageType, 1, costInMicroCents);
+            this.#meteringService.incrementUsage(actor, usageType, usageAmount, costInMicroCents);
 
             const first = response.data[0] as { url?: string; b64_json?: string };
             const url = first.url || (first.b64_json ? `data:image/png;base64,${ first.b64_json}` : undefined);


src/backend/src/services/ai/image/providers/TogetherImageGenerationProvider/TogetherImageGenerationProvider.ts
```diff
@@ -49,6 +49,19 @@ const CONDITION_IMAGE_MODELS = [
...
 ];
 
+const GEMINI_3_IMAGE_RESOLUTION_MAP: Record<string, Record<string, { w: number; h: number }>> = {
+    '1:1':   { '1K': { w: 1024, h: 1024 }, '2K': { w: 2048, h: 2048 }, '4K': { w: 4096, h: 4096 } },
+    '2:3':   { '1K': { w: 848,  h: 1264 }, '2K': { w: 1696, h: 2528 }, '4K': { w: 3392, h: 5096 } },
+    '3:2':   { '1K': { w: 1264, h: 848  }, '2K': { w: 2528, h: 1696 }, '4K': { w: 5096, h: 3392 } },
+    '3:4':   { '1K': { w: 896,  h: 1200 }, '2K': { w: 1792, h: 2400 }, '4K': { w: 3584, h: 4800 } },
+    '4:3':   { '1K': { w: 1200, h: 896  }, '2K': { w: 2400, h: 1792 }, '4K': { w: 4800, h: 3584 } },
+    '4:5':   { '1K': { w: 928,  h: 1152 }, '2K': { w: 1856, h: 2304 }, '4K': { w: 3712, h: 4608 } },
+    '5:4':   { '1K': { w: 1152, h: 928  }, '2K': { w: 2304, h: 1856 }, '4K': { w: 4608, h: 3712 } },
```
ProgrammerIn-wonderland: can this be moved to the models.ts somehow? I'm somewhat against model specific logic here but I understand why its here too since gemini billing is weird

src/backend/src/services/ai/image/providers/TogetherImageGenerationProvider/TogetherImageGenerationProvider.ts
```diff
@@ -101,37 +114,49 @@ export class TogetherImageGenerationProvider implements IImageProvider {
...
         if ( ! usageAllowed ) {
             throw APIError.create('insufficient_funds');
         }
 
-        const request = this.#buildRequest(prompt, { ...options, ratio, model: selectedModel.id.replace('togetherai:', '') }) as unknown as Together.Images.ImageGenerateParams;
+        // Resolve abstract aspect ratios to actual pixel dimensions for models that need it
+        let resolvedRatio = ratio;
+        if ( quality && ratio ) {
+            const ratioKey = `${ratio.w}:${ratio.h}`;
+            const resolutionEntry = GEMINI_3_IMAGE_RESOLUTION_MAP[ratioKey]?.[quality];
```
ProgrammerIn-wonderland, line 137: what does this code do if you're not using Gemini? I dont see any gating

P3il4: other models don't have an allowedQualityLevel
https://github.com/HeyPuter/puter/blob/44bb5953b4afe2854454f3381e938d30d1839521/src/backend/src/services/ai/image/providers/TogetherImageGenerationProvider/models.ts#L176-L184
so this part
https://github.com/HeyPuter/puter/blob/44bb5953b4afe2854454f3381e938d30d1839521/src/backend/src/services/ai/image/AIImageGenerationService.ts#L281-L289
makes quality = '' and this block never runs
i guess an exception is the flash-image-2.5 model:
https://github.com/HeyPuter/puter/blob/44bb5953b4afe2854454f3381e938d30d1839521/src/backend/src/services/ai/image/providers/TogetherImageGenerationProvider/models.ts#L212-L234
which does run through this block, however it uses actual width+height values instead of aspect ratios, so nothing changes. passing a gemini-3-pro ratio value into flash-image-2.5 is also impossible because of a similar allowedRatios check in AIImageGenerationService

src/backend/src/services/ai/image/providers/TogetherImageGenerationProvider/TogetherImageGenerationProvider.ts
P3il4: moved it into models.ts

src/backend/src/services/ai/image/providers/TogetherImageGenerationProvider/TogetherImageGenerationProvider.ts
ProgrammerIn-wonderland: I think its probably still a good idea to put this behind and if(model == "gemini-3-pro")ish block just for code clarity reasons for now

P3il4: alright, i also put the price logic behind a similar check since it's only for gem 3 for now

ProgrammerIn-wonderland:
Aiming to merge this this week after testing to make sure there aren't billing bugs

ProgrammerIn-wonderland:
LGTM

PR #2507, opened 2026-02-18
Add permission requests for directories of subdomains associated with apps owned by the user

KernelDeimos:
This allows an app to request permission to access the directory for one of the user's apps. The user must have permission to access the directory in order for the app to have permission. The subdomain associated with the app is determined in one of two ways:
1. Subdomain has an associated app set (typical deployment)
2. Subdomain URL matches (create a site and then set index url of an app)

Permission is checked primarily based on filesystem permission of the directory. App ownership is also checked as an additional precaution.

### DIFF
--- a/src/backend/src/api/APIError.js
+++ b/src/backend/src/api/APIError.js
@@ -155,7 +155,9 @@ class APIError {
         },
         'forbidden': {
             status: 403,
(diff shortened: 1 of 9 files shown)


KernelDeimos:
I hate ESM/CJS interop ***soooo much***

PR #2515, opened 2026-02-19
feat: cleanup cache invalidation and pull out keys to be easily used in other places that need the same cache

Salazareo:


Salazareo:
this looks big, but its literally just the package-lock I had to delete to fix up

PR #2540, opened 2026-02-24
fix(puter.js): give assignment to defaultGUIOrigin precedence

KernelDeimos:
This change gives precedence to the assignment of `defaultGUIOrigin` over global values typically set by the bundler. This is helpful when testing puter.js behavior in third-party sites with a locally running instance of Puter, where this value can be incorrect.

In the following code snippet, the behavior before this change is the output "default gui origin? https://puter.com". After this change, the output is "default gui origin? https://example.com" as expected.

```javascript
puter.defaultGUIOrigin = 'https://example.com';
console.log('default gui origin?', puter.defaultGUIOrigin);
```

### DIFF
--- a/src/puter-js/src/index.js
+++ b/src/puter-js/src/index.js
@@ -101,6 +101,7 @@ const puterInit = (function () {
 
         #defaultAPIOrigin = 'https://api.puter.com';
         #defaultGUIOrigin = 'https://puter.com';
+        #overrideDefaultGUIOrigin = undefined;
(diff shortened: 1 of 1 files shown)


KernelDeimos:
Set this to draft to avoid confusion. This is an idea for a change we could make to make testing easier, and is a "fix" because I think the current behavior would be unexpected, but it is not a fix for a bug related to anything currently being investigated.

PR #2543, opened 2026-02-25
fix(gui): add missing awaits in calls to update_auth_data

KernelDeimos:
This didn't fix the bug I was investigating. Maybe these awaits should still be here though. I'm leaving this open and unmerged for now.


KernelDeimos:
Setting this to draft to avoid confusion. This should not be merged unless it's found that this behavior is better.

Issue #2553, opened 2026-02-26
Can't change username (Missing authentication token)


reynaldichernando:
hi team, as the title suggests, i can't change username, it shows a missing authentication token error

<img width="592" height="418" alt="Image" src="https://github.com/user-attachments/assets/b6aa1d75-dc50-4840-aca3-56026e0785d6" />

how to repro:
1. go to `https://puter.com/dashboard#account`
2. change username
3. get error message

note the request indeed seems to not have any auth token

<img width="663" height="506" alt="Image" src="https://github.com/user-attachments/assets/2096324e-018e-48f5-999b-8e2953b151e4" />

(truncated the cookie for the screenshot)

4nshuman:
Hey @reynaldichernando, I looked into this issue and I believe it's a 2-tier fault. First, my assumptions:
1. I assume you encountered the authentication issue not on production but rather on your local system.
2. If not on local, then I suppose you were somehow running http rather than https.
The issue is that recently there was a change by @KernelDeimos where they added a re-authentication flow for protected actions. This means that authentication now goes through the cookie itself rather than as an auth token. 

There are two issues:
1. On an insecure connection, the cookies will never be saved — see login.js, line 35 (secure: true). So if you're running locally, you'd need to either run HTTPS with a self-signed cert, or as a quick fix, comment out that line.
2. Once you get past issue 1, you'll start seeing a 500 error with a "request failed" prompt. This one also affects production. The cause is that the rate limit scope change-username-start is not registered in EdgeRateLimitService. The handler should be using the existing scope /user-protected/change-username instead.

I have a fix and will raise a PR soon.
Please do let me know in case I might've missed out on something.

Reference snaps : 
<img width="1013" height="604" alt="Image" src="https://github.com/user-attachments/assets/43ce4f78-e4fc-4f9d-afe9-aaf4d50c10bc" />
<img width="642" height="767" alt="Image" src="https://github.com/user-attachments/assets/eac3fd11-bb28-4918-8161-7d918766e5cc" />

reynaldichernando:
thank you for the explanation @4nshuman , although this case is on production and using https

4nshuman:
That's correct @reynaldichernando so if somehow you saw authentication failure on PROD, then your cookies were missing on the session, not sure how though, cause login creates the cookies.
At the moment on PROD, I see Request Failed error, do you see it differently and if yes, could you please try "logging off and on again?" Classic IT response I know.. but just to be sure!! :D

[...]

PR #2559, opened 2026-02-27
fix: corrected service reference

4nshuman:
This PR will solve the username change issue raised at [2553](https://github.com/HeyPuter/puter/issues/2553)
The change includes reference correction on the EdgeRateLimitService along with a couple of formatting fixes.


KernelDeimos:
Yep, I see what happened here. Better fix is to make EdgeRateLimitService have a reasonable default and not throw an error here, but I'm going to merge this as well.

PR #2568, opened 2026-02-28
fix: break loop when keys.length is 0

KernelDeimos:
I don't know if this is the correct thing to do, but `keys.length` appears to be `0` for every iteration when this goes haywire - i.e. loops an astonishing number of times for apparently no reason.


KernelDeimos:
This was solved differently

PR #2569, opened 2026-02-28
fix gemini image provider implementation

P3il4:
fixed the current txt2img implementation for the gemini models:
1. replaced the in-prompt resolution logic with passing proper aspectratio and imagesize config as per the [gemini docs](https://ai.google.dev/gemini-api/docs/image-generation#aspect_ratios_and_image_size)
2. added support for passing [multiple images](https://ai.google.dev/gemini-api/docs/image-generation#use-14-images) with "input_images" array, backwards compatible with old "input_image"
3. added automatic mime type detection for input images, supporting both raw data and full data uris
4. replaced hardcoded quality:resolution with [token-based billing](https://ai.google.dev/gemini-api/docs/pricing#gemini-3-pro-image-preview) for both text and image

side fix: fixed puter.ai.txt2img not selecting the correct model even when providing exact puterId

notes:
1. all of this was made based on the google documentation and tested using my best-effort mock server of the gemini api made from the source of google/genai npm package and logs from free tier of aistudio as i don't have a paid gemini api key. would appreciate if someone could test with a real api key.
2. google documentation is inconsistent for 3.1 flash: [official model page](https://ai.google.dev/gemini-api/docs/pricing#gemini-3.1-flash-image-preview) has 4 tiers of pricing (747, 1120, 1680, 2520 for 0.5K, 1K, 2K and 4K accordingly), while [aspect ratio table](https://ai.google.dev/gemini-api/docs/image-generation#aspect_ratios_and_image_size) has 747, 1120, 1120, 2000 for the same quality levels. my code uses the official model page pricing.
3. 560 input tokens per input image is based on [gemini 3 pro image pricing](https://ai.google.dev/gemini-api/docs/pricing#gemini-3-pro-image-preview) page. other models don't specify a particular input image price but have "text/image" type input billing. there is a way to calculate [image input tokens for other models](https://ai.google.dev/gemini-api/docs/image-understanding#token_calculation), however in my opinion that's rather compute heavy and a simple count * 560 is fair enough for a pre-request check. assuming the real api returns actual prompt tokens for actual billing.
4. this changes output metering keys from a single "output" to "output:text" + "output:image". in case any analytics system was reading that key.

### DIFF
--- a/src/backend/src/services/ai/image/providers/GeminiImageGenerationProvider/models.ts
+++ b/src/backend/src/services/ai/image/providers/GeminiImageGenerationProvider/models.ts
@@ -52,13 +52,14 @@ export const GEMINI_IMAGE_GENERATION_MODELS: IImageModel[] = [
         name: 'Gemini 2.5 Flash Image',
         version: '1.0',
         costs_currency: 'usd-cents',
-        index_cost_key: 'output_image',
+        index_cost_key: '1x1',
         index_input_cost_key: 'input',
         allowedQualityLevels: [''],
         costs: {
             input: 30, // $0.30 per 1M input tokens (text/image)
             output: 250, // $2.50 per 1M output tokens (text and thinking)
             output_image: 3000, // $30.00 per 1M output image tokens
+            '1x1': 3.9,
         },
         allowedRatios: [
             { w: 1, h: 1 },
@@ -79,7 +80,7 @@ export const GEMINI_IMAGE_GENERATION_MODELS: IImageModel[] = [
         name: 'Gemini 3 Pro Image',
         version: '1.0',
         costs_currency: 'usd-cents',
-        index_cost_key: 'output_image',
+        index_cost_key: '1K:1x1',
         index_input_cost_key: 'input',
         aliases: [
             'gemini-3-pro-image-preview',
@@ -105,6 +106,7 @@ export const GEMINI_IMAGE_GENERATION_MODELS: IImageModel[] = [
             input: 200, // $2.00 per 1M input tokens (text/image)
             output: 1200, // $12.00 per 1M output tokens (text and thinking)
             output_image: 12000, // $120.00 per 1M output image tokens
+            '1K:1x1': 13.4,
         },
     },
     {
@@ -113,7 +115,7 @@ export const GEMINI_IMAGE_GENERATION_MODELS: IImageModel[] = [
         name: 'Gemini 3.1 Flash Image',
         version: '1.0',
         costs_currency: 'usd-cents',
-        index_cost_key: 'output_image',
+        index_cost_key: '1K:1x1',
         index_input_cost_key: 'input',
         aliases: [
             'gemini-3.1-flash-image-preview',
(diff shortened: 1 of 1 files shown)


ProgrammerIn-wonderland:
The costs array and keys need to be there as it's used for the onlime models browser https://developer.puter.com/ai/models/

ProgrammerIn-wonderland:
If I recall correctly Gemini image provider also already did meter based on tokens since https://github.com/HeyPuter/puter/blame/main/src/backend/src/services/ai/image/providers/GeminiImageGenerationProvider- /GeminiImageGenerationProvider.ts#L138 so I'm not sure if some of the billing changes are necessary

P3il4:
> If I recall correctly Gemini image provider also already did meter based on tokens since https://github.com/HeyPuter/puter/blame/main/src/backend/src/services/ai/image/providers/GeminiImageGenerationProvider- /GeminiImageGenerationProvider.ts#L138 so I'm not sure if some of the billing changes are necessary

current billing system uses hardcoded values for image dimensions and then adds text output tokens on top
https://github.com/HeyPuter/puter/blob/f2926c948ee2087881a554abb0d11b93b74b457d/src/backend/src/services/ai/image/providers/GeminiImageGenerationProvider/GeminiImageGenerationProvider.ts#L91-L92
https://github.com/HeyPuter/puter/blob/f2926c948ee2087881a554abb0d11b93b74b457d/src/backend/src/services/ai/image/providers/GeminiImageGenerationProvider/GeminiImageGenerationProvider.ts#L130-L149

by using entire candidatesTokenCount as output, current system bills user for text + image tokens as pure text output and adds image cost on top, resulting in extra charges
seperating output into outputTextCostInCents and outputImageCostInCents allows for more accurate billing. thought tokens (if present) aren't included in candidates count
image
plus not having to hardcode every dimension is a nice bonus

> The costs array and keys need to be there as it's used for the onlime models browser https://developer.puter.com/ai/models/

is this open source? i wasn't able to find it in the repo. i could revert the index_cost_key change and add one hardcoded dimension as that seems to be what it's pulling, but it wouldn't be complete docs. or put all dimension costs on the page in tokens and pull image output cost from models.ts

ProgrammerIn-wonderland:
@Salazareo can you confirm what @P3il4 said about metering in Gemini service?

ProgrammerIn-wonderland:
Staging this for either today or this weekend

ProgrammerIn-wonderland:
never mind the cost removals haven't been reverted so this will still break the models page. To answer the other question, I dont believe its opensource right now, but the code does rely on those keys

ProgrammerIn-wonderland:
@reynaldichernando can likely provide more details as he maintains it

P3il4:
> never mind the cost removals haven't been reverted so this will still break the models page.

fixed that so it's possible to merge now without breaking anything

ProgrammerIn-wonderland:
Thanks will look into it again

reynaldichernando:
hi there  @P3il4 , 
thank you for the PR, and what you see in the developer site is indeed based upon whatever data is in the repo
(which is exposed via https://puter.com/puterai/image/models/details)

the data shown in developer site are regularly updated (not automatic), so i will need to update those if there's new data

and yes you are right it is relying on `index_cost_key` to determine what's shown

we are still figuring out how to best show this (gemini), especially since with image models there are lots of variation based on sizes, also input/output tokens

PR #2578, opened 2026-03-03
Eric/26323 revert 26322

KernelDeimos:
This didn't work. I tested it wrong.


KernelDeimos:
Wow, okay - really inconvenient issue with Squash and Merge. **NEVER** Squash and merge a revert. If your branch happens to contain also the commit from your feature branch that you're reverting, which is entirely likely, it's going to just make a stupid PR like this one that makes zero changes and if you don't notice you'll be scratching your head a while trying to figure out why `main` still has the bug you thought you introduced and reverted.

PR #2617, opened 2026-03-06
clean(apps): merge fns only called by get_app_icon_url

KernelDeimos:
Each of these methods is only being called by get_app_icon_url. By changing that ternary operator into an `if` block with an early return we can move the bodies of these functions here. If one of the former functions ever needs to be used in more than one place, it's easy to extract it out of here agian - for now though let's not pay the cost of jumping to another function definition while trying to understand this code if we're not gaining anything from it.

### DIFF
--- a/src/backend/src/helpers.js
+++ b/src/backend/src/helpers.js
@@ -145,37 +145,6 @@ const isBase64AppIcon = (app) => {
     return isRawBase64ImageString(trimmed);
 };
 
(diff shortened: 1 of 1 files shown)


KernelDeimos:
I wanted to add more to this PR but I'll get this in for now and PR further changes later, since this approach reduces the chances of a merge conflict in the future.

PR #2660, opened 2026-03-13
fix coercion service not selecting correct rule

P3il4:
currently coerce method always returns a value even when no coercion is found. this causes a "string:url:data" type result to run through a "string:url:web -> stream" coercion instead of the intended "string:url:data -> stream", resulting in issues such as undefined content_type and passing a base64 data url into the web url processing logic. 
changed the method to return undefined in case no valid coercion was found and to allow it to try others.

side note: TypeSpec.equals() only checks the base value ($) ignoring content_type. this means a content_type: video result can get routed through a content_type: image rule. this doesn't break anything right now as they're almost the same, but something to keep in mind.
https://github.com/HeyPuter/puter/blob/87672802fe3b0bc7b4052b9c1b892798556458b6/src/backend/src/services/drivers/meta/Construct.js#L156-L168

### DIFF
--- a/src/backend/src/services/drivers/CoercionService.js
+++ b/src/backend/src/services/drivers/CoercionService.js
@@ -66,11 +66,13 @@ class CoercionService extends BaseService {
             coerce: async typed_value => {
                 console.debug('coercion is running!');
 
-                const response = await secureAxiosRequest(CoercionService.MODULES.axios,
-                                typed_value.value,
-                                {
-                                    responseType: 'stream',
-                                });
+                const response = await secureAxiosRequest(
+                    CoercionService.MODULES.axios,
+                    typed_value.value,
+                    {
+                        responseType: 'stream',
+                    },
+                );
 
                 return new TypedValue({
                     $: 'stream',
@@ -89,11 +91,13 @@ class CoercionService extends BaseService {
                 content_type: 'video',
             },
             coerce: async typed_value => {
-                const response = await secureAxiosRequest(CoercionService.MODULES.axios,
-                                typed_value.value,
-                                {
-                                    responseType: 'stream',
-                                });
+                const response = await secureAxiosRequest(
+                    CoercionService.MODULES.axios,
+                    typed_value.value,
+                    {
+                        responseType: 'stream',
+                    },
+                );
 
                 return new TypedValue({
                     $: 'stream',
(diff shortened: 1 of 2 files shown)


ProgrammerIn-wonderland:
@KernelDeimos does this look correct?

KernelDeimos:
This does look correct. How do we test this? (i.e. through which driver/feature was the problem observed and what steps were taken to produce it?)

P3il4:
> This does look correct. How do we test this? (i.e. through which driver/feature was the problem observed and what steps were taken to produce it?)

original issue came out of the gemini image generation driver causing this error when requesting a 2K quality image with flash-image-3.1:
```
{"success":false,"error":{"key":"url","expected":"web URL","got":"Maximum call stack size exceeded","code":"field_invalid","$":"heyputer:api/APIError","message":"Field `url` is invalid. Expected web URL. Got Maximum call stack size exceeded.","status":400}}
```
i traced the issue to secureAxiosRequest in securehttp.js. i wasn't able to reproduce the crash locally, but later realized that it shouldn't have went this way at all, as the gemini service was returning a string:url:data, while secureAxiosRequest was only supposed to be for string:url:web types.
https://github.com/HeyPuter/puter/blob/e28a5b01d85c69ffffa23c9796e60274f775b88d/src/backend/src/services/drivers/CoercionService.js#L162-L175
putting a few console logs around this spot i noticed that it was using the
```
produces: {
    $: 'stream',
    content_type: 'image',
},
consumes: {
    $: 'string:url:web',
    content_type: 'image',
},
```
coercion, resulting in odd behavior on smaller images too, such as content-type being "undefined"
applying my fix it now tries all available coercions until it finds the one it needs, no longer going into securehttp at all and responding with the correct content-type.

before the fix:
(logging desired_type and result in driverservice and the selected coercions in coercionservice)
image
after the fix:
image

you can test this on live puter.com right now:
```js
puter.ai.txt2img('generate an image', options= {
    ratio: {"w": 1, "h": 1},
    quality: "1K",
    model: "google:google/gemini-3.1-flash-image-preview",
})
```
going into network tab, the /call will have content-type: undefined.

```js
puter.ai.txt2img('generate an image', options= {
    ratio: {"w": 1, "h": 1},
    quality: "2K",
    model: "google:google/gemini-3.1-flash-image-preview",
})
```
and this will result in the error mentioned above

PR #2664, opened 2026-03-14
feat: add puter.peer to sdk, create PeerService in backend

velzie:


ProgrammerIn-wonderland:
I think that there is some rigidity here that I don't fully like. I see that the stun URLs are hardcoded in puterjs. I think this should probably come from the signalling worker in some capacity so stun servers can be changed or TURN creds can be retrofitted/changed. 

The main reason for this is that once we deploy a version of puterjs, we support it pretty much indefinitely. This is why for example in the wisp service, the wisp URLs are given from the backend instead of being hardcoded in.

ProgrammerIn-wonderland:
I think a good example of what I want is basically what speed.cloudflare.com does in that the full webrtc config is sent from the server
image

src/puter-js/src/modules/Peer.js
```diff
@@ -0,0 +1,413 @@
+let iceServers = [
```
ProgrammerIn-wonderland: make these be served from the signaling worker

src/puter-js/src/modules/Peer.js
```diff
@@ -0,0 +1,413 @@
...
+    #signallerUrl;
+    #datachannel;
+    connected = false;
+    closed = false;
+    #bufferedMessages = [];
+    constructor (signallerUrl, authToken) {
+        super();
+        this.#signallerUrl = signallerUrl;
+        this.#authToken = authToken;
+        this.rtconn = new RTCPeerConnection({ iceServers });
```
ProgrammerIn-wonderland: main reason for the signalling worker serving the config is so here, the worker can determine if you should be given turn creds or not, possibly depending on how much account balance is left

ProgrammerIn-wonderland:
otherwise, the actual core code seems pretty good and solid for the first iteration

src/puter-js/src/modules/Peer.js
ProgrammerIn-wonderland: probably replace the { iceServers } with a fetch() to the signalling worker and that's the fix here

src/puter-js/src/modules/Peer.js
velzie: didn't bother because it would be getting replaced with TURN anyway, but I suppose it should probably be kept in there in case you don't want it or something

PR #2665, opened 2026-03-14
fix flash image quality level

P3il4:
missed this in my last pr
turns out it's 512 not 0.5K for flash image 3.1 quality level
https://ai.google.dev/gemini-api/docs/image-generation#use-14-images -> click "Python"
currently causes "Request contains an invalid argument" when requesting 0.5K (default)
```js
puter.ai.txt2img('generate an image', options= {
    model: "google:google/gemini-3.1-flash-image-preview",
})
```
other quality levels are unaffected and work as intended
not able to verify this locally as i don't have a gemini key but i assume we can trust gemini docs


ProgrammerIn-wonderland:
Makes sense

PR #2690, opened 2026-03-19
Fix double slash in canonical URL #2672

### LINKED ISSUE #2672: Fix docs index page canonical link
right now in our docs canonical link have a double slash in the url

Image

it should just be https://docs.puter.com/

nandinikashyap1902:
Fixes #2672 

### Description
This pull request fixes a minor structural issue in the Puter.js documentation where the `<link rel="canonical">` tag was generating a double slash `//` at the end of the root URL (`https://docs.puter.com//`).

###solution
The metadata builder (`src/docs/build.js`) previously made an assumption that the underlying URL would not have a trailing slash and blindly appended one via string concatenation (`/">`). Because Node.js's `URL` constructor evaluates the root index automatically with a trailing slash resulting in `https://docs.puter.com/`, the manual string concatenation triggered the double slash. 

This commit updates the canonical tag string interpolation to dynamically check if the URL already resolves ending with `/`, ensuring exactly one trailing slash across both root and sub-pages.

### DIFF
--- a/src/docs/build.js
+++ b/src/docs/build.js
@@ -273,7 +273,11 @@ function generateDocsHTML (filePath, rootDir, page, isIndex = false) {
         html += `<meta name="title" content="${removeTags(page.title_tag ?? page.title)}" />`;
     }
     // Self referencing canonical
-    html += `<link rel="canonical" href="${new URL(page.path, site).href}/">`;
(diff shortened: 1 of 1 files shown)


reynaldichernando:
hi @nandinikashyap1902 , thank you for the PR!
the code works as expected, i am merging this

PR #2698, opened 2026-03-19
wip: ai video gen ai refactor

ProgrammerIn-wonderland:

PR #2711, opened 2026-03-23
refactor ai video gen

P3il4:
refactored ai video generation into a single service
- created main service file similar to [image service](src/backend/src/services/ai/image/AIImageGenerationService.ts)
- added two guards against bad duration and dimension values (from #2698 )
- changed costs logic from pulling costmaps to having it in models.ts and using costOverrides
- added duration and seconds values to /models/details of openai models for consistency

smaller stuff:
- added sora pro xxl ($0.70 tier) https://developers.openai.com/api/docs/models/sora-2-pro
- added options.test_mode for video missing from [docs](https://docs.puter.com/AI/txt2vid/)
- added exact model id match because of openai/together sora alias

tested with a mock server implementation of openai video and togetherai video by changing baseurl
.getMonthlyUsage() logged costkeys remain the same
legacy driver hints openai/together-video-generation still work
possibly breaking change for apps: in togetherai durationSeconds is now an array instead of a number
and some togetherai models ids are now lowercase 
`togetherai:ByteDance/Seedance-1.0-pro` -> `togetherai:bytedance/seedance-1.0-pro`


ProgrammerIn-wonderland:
GitHub quite literally will not let me merge this PR because it timed out the first time
image

ProgrammerIn-wonderland:
you might need to just close this one and make another one if it doesn't resolve itself

ProgrammerIn-wonderland:
but everything seems fine

PR #2714, opened 2026-03-23
Revamp networking api with epoxy

r58Playz:
~~Need to finalize epoxy CI/deployment first before merging~~


ProgrammerIn-wonderland:
is this stable? how is browser support?

r58Playz:
Minmium chrome version is somewhere in between 90 and 100. Ungoogled chrome archives are pretty sparse and it works on ugc r100 but not ugc r90

Minimum firefox version is ~100 (before that puter just doesn't let you log in thanks to turnstile)

Minimum safari version is probably 17 or 18 I don't think there's a good way to test older versions

ProgrammerIn-wonderland:
@r58Playz can you fix the merge conflict? auth requirements in PSocket changed

ProgrammerIn-wonderland:
after that I'll merge, every browser seems to work

ProgrammerIn-wonderland:
crap this is actually causing issues with puter worker builds

PR #2732, opened 2026-03-26
Getting rid of Endpoint

ProgrammerIn-wonderland:


ProgrammerIn-wonderland:
This is probably the first of a few PRs but we should get this in tree before there's any diversion

PR #2739, opened 2026-03-27
docs(kv): add sorting section with examples to list() API reference

RaffyLeong:
Added sorting section with examples for lexicographic order and zero-padding for numbers.

### DIFF
--- a/src/docs/src/KV/list.md
+++ b/src/docs/src/KV/list.md
@@ -128,7 +128,6 @@ If the user has no keys, the array will be empty.
     </script>
 </body>
 </html>
-
 ```
 
 <strong class="example-title">Sort numeric keys with zero-padding</strong>
@@ -166,5 +165,4 @@ If the user has no keys, the array will be empty.
     </script>
 </body>
 </html>
-
 ```


reynaldichernando:
hi @RaffyLeong , thank you for the PR!
for the examples, could you help follow the existing format? where
- code examples are full HTML code
- are under the #examples section
- adding the entries in `src/docs/src/examples.js` and `src/docs/src/playground/examples/**`

and the mention about keys being sorted lexicographically, i think we can just have this included in the first sentence/overview sentence

feel free to clarify if there's anything unclear

reynaldichernando:
^ the main idea with the examples are so that people can run it in the playground, and see the results immediately

RaffyLeong:
Hi, @reynaldichernando,  Thank you for the feedback! I'm new to open source and want to make sure I get this right.
Could you help me understand the exact structure I need to follow for adding the sorting example?

From what I see:
- I should remove the Sorting section from src/docs/src/KV/list.md and add one sentence about lexicographic sorting at the top.
- I need to add an entry in src/docs/src/examples.js under the Key-Value Store section.
- I need to create a new HTML file in src/docs/src/playground/examples/** (similar to ai-chatgpt.html).
Is that correct? If so, what should I name the HTML file and where exactly should the entry go in examples.js?
Thank you for your patience!"

reynaldichernando:
> I should remove the Sorting section from src/docs/src/KV/list.md and add one sentence about lexicographic sorting at the top.

yes, this sounds good, something like "list returns array of keys sorted by the keys in lexicographic order" or other phrases that fits at the top

> I need to add an entry in src/docs/src/examples.js under the Key-Value Store section.
I need to create a new HTML file in src/docs/src/playground/examples/** (similar to ai-chatgpt.html).
If so, what should I name the HTML file and where exactly should the entry go in examples.js?

yes to both. you can name the file like kv-list-sort.html or something relevant, and the entry should go below the existing `kv-list`. keeping it near each other since they are related

also, if you are writing 2 different examples, then it should have 2 entries as well

lastly, don't forget to add the same example code under #examples section in KV/list.md, you can see how the example in the docs, matches the one in the playground

ProgrammerIn-wonderland:
@reynaldichernando do you approve?

reynaldichernando:
hi @RaffyLeong , thank you for the update!
everything's looking good i've pushed some minor refinements including:
- improving the copy 
- updating example for readability (and following existing format)

merging this right now, thank you for your contribution 🫡

PR #2764, opened 2026-04-02
Ns/simplify 1

ProgrammerIn-wonderland:


FixleCH:
Good morning,

Could you please clarify the purpose of this PR?

It appears that a significant number of useful functions (over 1000 lines of code) are being removed. I’d like to better understand the reasoning behind this change and whether these functionalities are being replaced or relocated elsewhere.

Kind regards,
Fixle

ProgrammerIn-wonderland:
It relates to removing dev console code. Dev console itself was already removed but this PR removes all the stubs that hooked into dev console. This code isn't being moved anywhere since it is all dead code now

FixleCH:
> It relates to removing dev console code. Dev console itself was already removed but this PR removes all the stubs that hooked into dev console. This code isn't being moved anywhere since it is all dead code now

Hi,

Thanks for the clarification.

I understand that the dev console has been removed for now, but is there a possibility it might be reintroduced in the future? If so, removing all related functions could make it more complex to restore that functionality later.

Also, I’d suggest adding a description to the PR itself. Since it’s currently empty, it’s difficult to understand the intent behind such a large change. A short explanation would really help provide better context and avoid confusion during review.

Kind regards,
Fixle

ProgrammerIn-wonderland:
It's a extremely unlikely thay dev console will be readded as it was a nightmare for maintainability. These stubs serve no real purpose now, and any future additions are likely to be to an admin dashboard which will require a very different set of stubs than this

PR #2771, opened 2026-04-06
add gemini image models in .chat

P3il4:
added support for gemini models in .chat that can edit images in a conversation
similar to how [openrouter](https://openrouter.ai/docs/guides/overview/multimodal/image-generation#basic-image-generation) does it (image_config, response formats, etc)
implementation based on [gemini image provider](https://github.com/HeyPuter/puter/blob/b0bd8a12a75d2b919951d2f640189ddaf12f7cde/src/backend/src/services/ai/image/providers/GeminiImageGenerationProvider/GeminiImageGenerationProvider.ts)
kept the same usage keys as txt2img (gemini:model:output:text/image or input)
testing:
```js
await puter.ai.chat([
    { role: 'user', content: 'Draw a cat' },
], { 
  model: "gemini-3.1-flash-image-preview", 
  stream: false, 
  image_config: { aspect_ratio: '16:9', image_size: '1K' } 
});
```
for stream: false, image is in result.message.images
for stream: true, in image type event
also works with openai compatible route, matching openrouter format exactly
```js
const apiBase = puter.APIOrigin;
const token = puter.authToken;
const resp = await fetch(`${apiBase}/puterai/openai/v1/chat/completions`, {
    method: 'POST',
    headers: {
        'Content-Type': 'application/json',
        'Authorization': `Bearer ${token}`,
    },
    body: JSON.stringify({
        model: "gemini-3.1-flash-image-preview",
        messages: [{ role: 'user', content: 'Draw a small red dot' }],
        stream: true,
        image_config: { aspect_ratio: '16:9', image_size: '1K' }
    }),
});
```
there are more examples in docs but those are the main ones
tested with a mock server, would appreciate testing on a real key.

### DIFF
--- a/src/backend/src/services/ai/chat/providers/GeminiProvider/GeminiChatProvider.ts
+++ b/src/backend/src/services/ai/chat/providers/GeminiProvider/GeminiChatProvider.ts
@@ -131,26 +131,54 @@ export class GeminiChatProvider implements IChatProvider {
             const role = msg.role === 'assistant' ? 'model' : 'user';
             const parts: Record<string, unknown>[] = [];
 
+            // First pass: collect parts and find if any has thoughtSignature
+            let sharedThoughtSignature: string | undefined;
+            if ( Array.isArray(msg.content) ) {
+                for ( const part of msg.content ) {
+                    if ( part.type === 'image_url' && part.thoughtSignature ) {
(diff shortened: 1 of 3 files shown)


ProgrammerIn-wonderland:
This is awesome

ProgrammerIn-wonderland:
I see thoughtsignature is fixed now, testing again

Issue #2814, opened 2026-04-16
Update opus 4.7 knowledge cut off to `2026-01`


reynaldichernando:
https://github.com/HeyPuter/puter/blob/4130fc4104edfbba6ea2720aa290f0aa9673217b/src/backend/src/services/ai/chat/providers/ClaudeProvider/models.ts#L11

source:
https://platform.claude.com/docs/en/about-claude/models/overview

<img width="609" height="693" alt="Image" src="https://github.com/user-attachments/assets/8523bb63-b661-438b-8b26-81a68d406ae2" />

btw: this does not affect anything since its just data, so not the highest priority

subhp03:
Hey @reynaldichernando just opened a PR #2816

reynaldichernando:
thank you for your contributions @subhp03 , merged!

PR #2816, opened 2026-04-16
fix: update knowledge date for Claude Opus 4.7 model

subhp03:
This Pull Request updates the knowledge cutoff date for Claude Opus 4.7 in the provider model configuration. Previously, the cutoff was set to 2025-08, which was inaccurate. It is hereby changed to 2026-01. This value is referenced in `ClaudeProvider/models.ts`

Closes issue #2814


reynaldichernando (approved):
LGTM

PR #2916, opened 2026-05-05
Block unconfirmed users from API endpoints server-side

jelveh:
Add a default-on email confirmation gate that rejects users with `requires_email_confirmation && !email_confirmed` on all authenticated routes, returning 403 with `email_confirmation_required`.

Previously this was only enforced client-side via a GUI modal, meaning direct API calls could bypass the check entirely.

Essential routes are exempted via `allowUnconfirmed: true`:
  `/whoami`, `/logout`, `/send-confirm-email`, `/confirm-email`,
  `/save_account`, `/get-anticsrf-token`, `/get-gui-token`,
  `/session/sync-cookie`, `/auth/revoke-session`,
  `/user-protected/delete-own-user`

No impact on temp users (`requires_email_confirmation` is false), self-hosted deployments without email (flag is never set), or unauthenticated routes (login, signup, password recovery, OIDC).

### DIFF
--- a/extensions/whoami.ts
+++ b/extensions/whoami.ts
@@ -25,7 +25,7 @@ const CLIENT_VISIBLE_FEATURE_FLAGS: ReadonlySet<string> = new Set([
 
 extension.get(
     '/whoami',
-    { subdomain: 'api', requireAuth: true },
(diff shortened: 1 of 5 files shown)


Salazareo (approved):
LGTM

PR #2921, opened 2026-05-05
expose costs in ai chat response

### LINKED ISSUE #2892: Add usage.usd_cents to AI chat response usage object
## Summary

Add a `usd_cents` field to the `usage` object returned by `puter.ai.chat()` so users can programmatically track their spending.

## Current behavior

The usage object currently returns token counts only:

```json
"usage": {
    "prompt_tokens": 15,
    "completion_tokens": 1750,
    "cached_tokens": 0
}
```

While token counts are useful, they don't directly tell users how much a request cost — especially since pricing varies per model and cached vs uncached tokens are priced differently.

## Proposed behavior

Add a `usd_cents` field to the usage object:

```json
"usage": {
    "prompt_tokens": 15,
    "completion_tokens": 1750,
    "cached_tokens": 0,
    "usd_cents": 0.42
}
```

This would represent the total cost of the request in US cents (fractions allowed), calculated server-side using the actual model pricing at the time of the request.

## Why

- Users building apps on Puter's AI API need to track costs accurately
- Token counts alone require users to maintain their own pricing tables per model, which is error-prone and goes stale as pricing changes
- Server-side cost calculation is the single source of truth since Puter already knows the model and its pricing
- Enables users to set budget alerts, show cost dashboards, or enforce spending limits in their apps

## Implementation notes

- The cost should be calculated server-side after the response is received from the upstream provider
- Should account for different input/output/cached token rates per model
- A value of `null` is acceptable if cost data is unavailable for a given provider

P3il4:
add `usd_cents` after a puter.ai.chat call:
```js
// non-streaming
"usage": {
    "prompt_tokens": 12,
    "completion_tokens": 10,
    "cached_tokens": 0,
    "usd_cents": 0.002352
},
```
```js
// streaming
{"type":"text","text":"hi."}
{"type":"usage","usage":{"prompt_tokens":11,"completion_tokens":9,"cached_tokens":0,"usd_cents":0.000705}}
```
usage is same as from puter.auth.getMonthlyUsage:
```js
// from the non stream case
"deepseek:deepseek-chat:prompt_tokens": {
    "units": 12,
    "cost": 672,
    "count": 1
},
"deepseek:deepseek-chat:completion_tokens": {
    "units": 10,
    "cost": 1680,
    "count": 1
},
// 672 + 1680 = 2352 / 1_000_000 = 0.002352
```
`ai.prompt.cost-calculated` event shares the same logic for consistency
testing:
```js
puter.ai.chat([{
        role: 'user',
        content: `message`,
        cache_control: { type: "ephemeral" } // mind the model minimal cache token length for testing cache cases
    },
], options={model: "claude-haiku-4-5"})
```
tested on openai (with, without cache), claude (with cache), openrouter (special "billedUsage" case) on live keys, deepseek/xai/gemini/mistral on mock servers
closes #2892

### DIFF
--- a/src/backend/drivers/ai-chat/ChatCompletionDriver.ts
+++ b/src/backend/drivers/ai-chat/ChatCompletionDriver.ts
@@ -357,13 +357,15 @@ export class ChatCompletionDriver extends PuterDriver {
             const init = res.init_chat_stream;
             const cleanup = res.finally_fn;
 
-            // Intercept `chatStream.end(usage)` so we can fire the same
-            // complete + cost-calculated events the non-streaming branch
-            // emits. Providers always terminate streams via `.end(usage)`;
-            // if they skip it, we just lose the cost event (no worse than
-            // not emitting).
+            // Intercept `chatStream.end(usage)` to fire complete + cost events
+            // (mirrors the non-streaming branch). Clone usage so providers that
+            // meter after this call (e.g. Claude) don't pick up `usd_cents`.
             const originalEnd = chatStream.end.bind(chatStream);
             chatStream.end = (usage?: Record<string, number>) => {
+                const enrichedUsage = usage ? { ...usage } : usage;
+                if (enrichedUsage) {
+                    this.#injectUsdCents(enrichedUsage, model);
+                }
                 this.clients.event.emit(
                     'ai.prompt.complete',
                     {
@@ -371,7 +373,7 @@ export class ChatCompletionDriver extends PuterDriver {
                         completionId,
                         intended_service: intendedProvider,
                         parameters: args,
-                        result: { usage, stream: true },
+                        result: { usage: enrichedUsage, stream: true },
                         model_used: model.id,
                         service_used: model.provider,
                     },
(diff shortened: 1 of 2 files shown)


Salazareo (changes requested):
can we call them microCents instead of Ucents

I know thats its meant to be the mu micro symbol, but bit less legible than microCents

P3il4:
> can we call them microCents instead of Ucents
> 
> I know thats its meant to be the mu micro symbol, but bit less legible than microCents

updated, the rest of the file + event uses ucents though, so this causes a bit of a naming inconsistency

Salazareo:
> > can we call them microCents instead of Ucents
> > I know thats its meant to be the mu micro symbol, but bit less legible than microCents
> 
> updated, the rest of the file + event uses ucents though, so this causes a bit of a naming inconsistency

well update as we touch it I guess, I didn't realize rest was also using ucents

Salazareo (approved):

P3il4:
ignore that, it's back at efd7b19

PR #2928, opened 2026-05-06
Update Z.AI vision model id to z-ai/z-ai for consistency

reynaldichernando:
Right now the vision model showing as `z-ai:zai/{id}`
image

compared to text model which are using `z-ai:z-ai/{id}`
https://github.com/HeyPuter/puter/blob/224ea5f7f835e57c65ac194f75c614246b276409/src/backend/drivers/ai-chat/providers/zai/models.ts#L44

to keep it consistent, i also updated the vision model id to use `z-ai:z-ai/{id}`

the zai/{id} version should still work since we have alias for it, plus this shouldn't really be a concern because we always document Z.AI model as z-ai (with a hyphen)


ProgrammerIn-wonderland:
Oops, thanks

PR #2930, opened 2026-05-06
Fix: support defaultValue in UIPrompt (app environment)

### LINKED ISSUE #2894: Add support for `defaultValue` option for `puter.ui.prompt` in puter app environment
Currently the `defaultValue` options only works in web environment. Ideally we match this functionality in puter apps env

app
Image

web
Image

Best to try and compare it here
https://ui-playground.puter.site/
https://puter.com/app/ui-playground

angelinazzz1809:
Closes #2894

I’d be happy to work on this issue. I’m new to contributing here, so any feedback or suggestions would be greatly appreciated.

In the app environment, the prompt input was initialized using `options.value`, so `defaultValue` was not reflected.

This change ensures that `defaultValue` is used when provided, while keeping backward compatibility with `options.value`.


jelveh:
Thank you for your contribution!

PR #2939, opened 2026-05-06
fix: thread referrer through OIDC state so signup records acquisition…

jelveh:
OIDC signups never had the referrer field set because the callback request from the IdP doesn't carry the frontend's JSON body. Fix by passing referrer as a query param to the OIDC start endpoint, embedding it in the signed state JWT, and forwarding it to createUserFromOIDC.


ProgrammerIn-wonderland (commented):
Code looks sane, need to test

PR #3013, opened 2026-05-08
Use GA version of gemini 3.1 flash lite

reynaldichernando:
- use the generally available version of gemini 3.1 flash lite
- the `-preview` prefix will still be supported via alias
- doing it now, because the preview one will be deprecated in ~2 weeks (May 25, 2026)

https://cloud.google.com/blog/products/ai-machine-learning/gemini-3-1-flash-lite-is-now-generally-available

image


reynaldichernando:
image

reynaldichernando:
old model id still working via alias
image

PR #3069, opened 2026-05-11
docs: add collapse/expand styling and handlers for long code snippets

jelveh:


reynaldichernando:
i think we are still missing the code to shrink long code snippets

PR #3106, opened 2026-05-13
Fix Docs search improvement: Improve fuzzy search

### LINKED ISSUE #2362: Docs search improvement

angelinazzz1809:
Fixes #2362

## Summary

This PR improves the docs search experience by integrating Fuse.js fuzzy search and refining search result behavior.

### Changes made

* Added Fuse.js based fuzzy search
* Improved exact match ranking
* Added support for nearest heading jump for fuzzy matches
* Reduced incorrect full-page highlighting behavior
* Added heading metadata into the search index
* Started restructuring search indexing toward section-level indexing

### Current behavior

* Exact matches and partial matches now behave more consistently
* Fuzzy matches can jump to the nearest heading section
* Search highlighting is more controlled and less noisy

### Notes

There are still some limitations with long-word typo matching (for example certain missing-character cases), especially when indexing large page content blocks.

I also started moving toward section-level indexing as suggested in the issue discussion(related to #2599), which should further improve fuzzy matching quality and relevance ranking.

Feedback is very welcome!


reynaldichernando:
letting you know that i'm still looking at this, please give me some time to review it
in the meantime, the other commiter yiqingzhao1216 needs to sign the CLA too, its standard for contribution 🙏

reynaldichernando:
hi @angelinazzz1809 , the PR seems to contain some merge conflict in `gui/src/UI`
this causes the project to not run, could you help take a look?

angelinazzz1809:
Thanks! The other account (yiqingzhao1216) is also mine — I accidentally used a different local git commit identity while pushing from my local environment, which caused the CLA mismatch. I've signed the CLA for that account as well, so it should be fully resolved now. 

And I synced with the latest upstream changes and resolved the merge conflict in gui/src/UI. The PR has been updated, so it should run correctly now. Please let me know if there are still any issues.

Appreciate you taking the time to review this!

reynaldichernando:
hi @angelinazzz1809 , thank you for your contributions
after some time reviewing the code, i don't think it fits with what we're looking for
there are still some areas where it still falls short:
- the fuse index creation should happen at build time, and not reusing the naive index
- fuzzy search still doesn't work
- and the code doesn't run

i'll close this PR for now, but you are welcome to open a new one if you are interested

PR #3109, opened 2026-05-13
Add new type of access token

ProgrammerIn-wonderland:


Salazareo:
closing for now since we went with middleware approach

PR #3122, opened 2026-05-16
feat(ai): add direct MiniMax provider integration

JasonDebnath001:
Adds direct MiniMax provider integration.

Includes:
- MiniMax provider implementation
- model registration
- provider registration
- config support
- chat completion support
- streaming-ready structure

Uses direct HTTP requests via axios instead of OpenAI SDK.

### DIFF
--- a/src/backend/drivers/ai-chat/providers/minimax/miniMaxProvider.ts
+++ b/src/backend/drivers/ai-chat/providers/minimax/miniMaxProvider.ts
@@ -1,0 +1,87 @@
+import axios, { AxiosInstance } from 'axios';
+
+type ChatMessage = {
+    role: 'system' | 'user' | 'assistant' | 'tool';
+    content: any;
+    name?: string;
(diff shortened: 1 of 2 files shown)


ProgrammerIn-wonderland:
Hey, I would actually prefer it if you used the OpenAI sdk so the code is more similar to our existing providers

ProgrammerIn-wonderland:
Notably absent is that this code doesn't seem like it works either. Streaming support is not here and metering is absent as well. I would start by copying an existing provider like z-ai

ProgrammerIn-wonderland (changes requested):
Please don't use axios. Use OpenAI library and implement streaming as well as metering. Refer to Z-ai

jelveh:
@JasonDebnath001 how about you run your trash, ai-generated, never-tested code at least once before spamming our repo?

JasonDebnath001:
You're right. I'm sorry. I'll work on myself.

PR #3167, opened 2026-05-27
Support PostgreSQL database backend

### LINKED ISSUE #3165: Support PostgreSQL as a database backend
As a self-hosted Puter operator, I want to run Puter with PostgreSQL instead of SQLite or MySQL/MariaDB, so that I can use the database platform I already operate and rely on PostgreSQL-native behavior.

Today Puter’s self-hosting path documents MariaDB/MySQL, while tests and local development primarily use SQLite. PostgreSQL users need a first-class backend option that can bootstrap a fresh database, run reliably at startup, and support normal Puter flows such as account creation, login, sessions, apps, permissions, and filesystem entries.

This should work by configuring Puter with:

```json
{
  "database": {
    "engine": "postgres",
    "host": "postgres",
    "port": 5432,
    "user": "puter",
    "password": "...",
    "database": "puter",
    "migrationPaths": ["./src/backend/clients/database/migrations/postgres"]
  }
}
```

or with a PostgreSQL connection string.

The implementation should use PostgreSQL directly rather than translating MySQL migrations or emulating MySQL broadly. It should include native PostgreSQL migrations, documented self-hosting configuration, and test coverage for the main Puter backend flows on a real disposable PostgreSQL database when a test URL is provided.

rhernaus:
## Summary

Closes #3165.

This adds PostgreSQL as a supported database backend for self-hosted Puter. It includes a native Postgres client, native migrations, SQL preparation for Postgres syntax, and the compatibility fixes needed for normal backend flows to work against a real Postgres database.

What's included:

- Postgres database client wiring in the existing database factory
- Native Postgres migration support and schema bootstrap
- Shared SQL helpers for identifiers, booleans, JSON extraction, upserts, and returning inserted IDs
- Store/controller updates for Postgres-compatible queries and result shapes
- Normalization for Postgres `int8` results so session timestamps and quota values match existing expectations
- Self-hosting docs for configuring PostgreSQL
- Unit tests plus real disposable-Postgres integration coverage

## Testing

- `npm run test:backend -- src/backend/clients/database/preparePostgresSql.test.ts src/backend/clients/database/PostgresDatabaseClient.test.ts src/backend/stores/fs/FSEntryStore.test.ts`
- `PUTER_TEST_POSTGRES_URL='postgres://postgres:postgres@127.0.0.1:54329/puter_test' npm run test:backend -- src/backend/clients/database/PostgresDatabaseClient.integration.test.ts`
- `npm run build:ts`


Salazareo:
Hi @rhernaus could you look into adding [pgmock](https://github.com/hexclave/pgmock?tab=readme-ov-file) to be able to run those postgres tests more easily, and if possible, add a configuration path to have [setupTestServer](https://github.com/HeyPuter/puter/blob/775c9c49e2cebb217049d0cd4451a2f647ac1f27/src/backend/testUtil.ts#L17) run the test suite with pgmock as an in memory postgres?

I realize we don't have this for mysql (mostly out of lack of good in memory mocks) but as we won't have an active environment to test changes against postgres im worried it might degrade and rot overtime without a test harness we can configure to run frequently

rhernaus:
Hi @Salazareo this is addressed now.

I added `pgmock` so `PostgresDatabaseClient.integration.test.ts` runs by default without an external Postgres instance.

Salazareo (approved):
reviewed and added some docs and test changes to make maintaining this a bit easier, should be good to go though

Salazareo:
Thanks @rhernaus for the contribution, merged now, will cut a new release for it too

PR #3171, opened 2026-05-27
add minimax provider

ProgrammerIn-wonderland:


src/backend/drivers/ai-chat/providers/minimax/models.ts
```diff
@@ -0,0 +1,89 @@
...
+const CENTS_PER_USD = 100;
+const MTOK = 1_000_000;
+const CONTEXT_WINDOW = 204_800;
+const MAX_OUTPUT_TOKENS = 196_608;
+
+type MiniMaxChatModel = IChatModel & {
+    apiModel: string;
+};
+
+const usdPerMToken = (
```
Salazareo: nit: maybe should be universal for other models

Salazareo (approved):

src/backend/drivers/ai-chat/providers/minimax/models.ts
ProgrammerIn-wonderland: Not a bad idea

PR #3184, opened 2026-05-28
fix(docs/search): handle search index loading and failure states

### LINKED ISSUE #3180: Add edge case handling for docs search when index hasn't loaded yet
based on copilot comment
> When the MiniSearch index hasn’t finished loading (miniSearch is null), this path calls updateSearchResults([]), which renders the “No results found” empty state. That’s misleading during initial load/slow network.

https://github.com/HeyPuter/puter/pull/3179

ideally how we handle this
- if user tried to search, but search index hasn't fully loaded -> show info that search index is loading
- then if the index has finished loading, it should automatically retrigger search, so user don't have to retry searching

Anuragp22:
## Summary
- Replaces the misleading "No results found" empty state shown when users searched before the MiniSearch index finished loading. The search now shows "Loading search index..." and replays the query automatically once the index is ready.
- Renders a distinct "Failed to load search index. Please refresh." message on fetch error so the loading state does not persist indefinitely.

Fixes #3180.

## Open question for review
The failure-state message is not in the issue description but is a necessary consequence of the loading-state change. Without it, a failed `fetchSearchIndex` would leave "Loading..." showing forever. Open to dropping or rephrasing if a different approach is preferred.

The pre-existing `openSearchUI()` call to `updateSearchResults([])` (which renders "No results found" on open before any typing) is out of scope and untouched.

## Test plan
`src/docs/src/assets/js/search.js` has no existing unit tests, so this change is verified manually. Built and served locally (`npm run build && npm run serve` in `src/docs`):
- [x] Normal speed: search results render as before, no flash of loading state.
- [x] 3G throttle, type before `/index.json` loads: "Loading search index..." appears, results auto-render when the index lands without retyping.
- [x] `/index.json` returning 404: brief "Loading..." then "Failed to load search index. Please refresh."
- [x] 3G throttle, type then backspace to empty input before the index loads: pending query is cleared, no stale results fire when the index arrives.

### DIFF
--- a/src/docs/src/assets/js/search.js
+++ b/src/docs/src/assets/js/search.js
@@ -5,6 +5,15 @@ let miniSearch = null;
 let searchTimeout = null;
 let selectedSearchResult = -1;
 
+// Last query the user typed while the index was still loading; replayed by
+// fetchSearchIndex() once the index becomes available so the user doesn't
+// have to retype. Cleared on success, failure, or when the query falls
+// below the minimum length.
+let pendingQuery = null;
+// True after fetchSearchIndex() has thrown at least once; lets performSearch
+// distinguish "still loading" from "load failed" without waiting forever.
+let indexLoadFailed = false;
+
 // Query-time tokenizer. Splits on whitespace AND punctuation only — does
 // NOT emit the joined-without-punctuation form. The index (built in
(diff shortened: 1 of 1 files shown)


reynaldichernando (approved):
thank you for your contribution

PR #3185, opened 2026-05-29
feat (put-1049): worker types

Salazareo:


ProgrammerIn-wonderland (approved):

ProgrammerIn-wonderland:
Ignore that approval, I misclicked!

src/worker-types/globals.d.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import type { Puter } from '@heyputer/puter.js';
+import type { Router } from './types/router.d.ts';
+import type { PuterContext } from './types/event.d.ts';
+
+declare global {
+    /** The deployer's Puter context, authenticated with the worker's own credentials. */
+    const me: PuterContext;
+    /** Alias for {@link me}. */
+    const my: PuterContext;
```
ProgrammerIn-wonderland, line 17: only me should be typed, my and myself were globals we were experimenting with (for DevEX) but likely wont be keeping

src/worker-types/globals.d.ts
```diff
@@ -0,0 +1,31 @@
...
+    /** Alias for {@link me}. */
+    const myself: PuterContext;
+
+    /** The router used to register HTTP route handlers. */
+    const router: Router;
+
+    /** The deployer's Puter auth token, provided as a Cloudflare secret binding. */
+    const puter_auth: string;
```
ProgrammerIn-wonderland, line 25: This is only global as a consequence of how Service worker syntax workers work. If we migrate to Cloudflare ESM syntax workers (which i do want to do) puter_auth and puter_endpoint wont be global anymore. I dont think these should be in the type personally

ProgrammerIn-wonderland: If we keep it, we should make them @deprecated and say subject to future removal
Edit: Did not realize there was someone on github named deprecated

src/worker-types/types/event.d.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Present only when the worker was invoked with a `puter-auth` header
+     * (e.g. via `puter.workers.exec()`). When present, the deployer can act
+     * on the caller's behalf through `user.puter`.
+     */
+    user?: PuterContext;
+
+    /** Alias for {@link WorkerEvent.user}. */
+    requestor?: PuterContext;
```
ProgrammerIn-wonderland, line 31: also *probably* shouldn't be documented since this was an experiment, but we can keep it. I just think it may be confusing

src/worker-types/types/event.d.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/**
+ * The event object passed to every route handler.
+ *
+ * `TParams` defaults to the open-ended `Params` record, but the `Router`
+ * methods infer it from the path literal so handlers get precisely-typed
+ * `params` automatically.
+ */
+export interface WorkerEvent<TParams extends Params = Params> {
```
ProgrammerIn-wonderland: WorkerEvent actually is a FetchEvent https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent with `user` and `params` added on. This is important *mainly* because of the event waitUntil. However, since we aren't in the browser making our own type for this isn't a bad idea. I think we should just add waitUntil onto this instead of using proper FetchEvent.

ProgrammerIn-wonderland (changes requested):
Changes requested: (I can handle these if you'd like)
* add waitUntil into WorkerEvent
* Remove (or mark as deprecated) globalThis.my, globalThis.myself
* Remove (or mark as deprecated) globalThis.puterAuth and globalThis.puter_endpoint
* Remove (or mark as deprecated) globalThis.requestor

ProgrammerIn-wonderland (approved):

PR #3209, opened 2026-06-04
feat: add jspaint as built-in app

S33G:


Salazareo:
Hi @S33G unfortunately I don't think we can merge this.

While its really cool it doesn't actually interface with puter at all.

you're ofcourse able to add it in your own deployment, or as a 3rd party app in puter.com, but we probably can't have it as a recommended app universally

PR #3223, opened 2026-06-08
Add docs fixes for deployment and workers pages

reynaldichernando:
- update info about puterjs apps needs to be hosted by a web server
- tidy up workers.create docs page
- fix internal linking
- emphasize wildcard routes requiring name


reynaldichernando:
checks are failing for some reason (?)
but there are no backend changes, this is just docs, plus the github actions shouldn't be running anyway
i'm gonna merge

PR #3235, opened 2026-06-09
ci: add docs build workflow

### LINKED ISSUE #3225: Add CI for docs changes
hi, for docs changes (`src/docs`) we should atleast have a test to make sure it builds

it should be pretty straight forward, we only need to run `npm ci` and `npm run build`

wwb1942:
Closes #3225.

  Adds a CI workflow that runs when documentation files under `src/docs/**` change.

  Changes:
  - Triggers on PRs touching `src/docs/**`
  - Uses Node 24
  - Runs `npm ci` at repo root
  - Runs `npm run build` in `src/docs`

  Verified locally:
  - workflow YAML parses
  - `npm run build` passes in `src/docs` under Node 24/npm 11

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - 'src/docs/**'
+      - '.github/workflows/docs-build.yml'
 
 permissions:
   contents: read
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -4,10 +4,6 @@ on:
   pull_request:
     paths:
       - 'src/docs/**'
-      - '.github/workflows/docs-build.yml'
-
-permissions:
-  contents: read
 
 jobs:
   build-docs:
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,0 +1,30 @@
+name: Docs Build
+
+on:
+  pull_request:
+    paths:
+      - 'src/docs/**'
+
+permissions:
+  contents: read
+
+jobs:
+  build-docs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,31 @@
+name: Docs Build
+
+on:
+  pull_request:
+    paths:
+      - 'src/docs/**'
+      - '.github/workflows/docs-build.yml'
```
reynaldichernando: we should only listen for docs directory changes, no need to listen to the actions workflow file itself

.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,31 @@
...
+
+permissions:
+  contents: read
```
reynaldichernando: i think our workflow does not require this permission

reynaldichernando (changes requested):

reynaldichernando:
hi @wwb1942 , thank you for your contribution, i've left some review for the changes!

.github/workflows/docs-build.yml
wwb1942: Fixed in 55d2a66:
  - removed the workflow file from pull_request.paths
  - removed the explicit permissions block

  Verified locally that the workflow YAML parses and the docs build passes under Node 24/npm 11.

.github/workflows/docs-build.yml
wwb1942: Fixed in 55d2a66:
  - removed the workflow file from pull_request.paths
  - removed the explicit permissions block

  Verified locally that the workflow YAML parses and the docs build passes under Node 24/npm 11.

wwb1942:
> hi @wwb1942 , thank you for your contribution, i've left some review for the changes!

Addressed the review in 55d2a66:
  - removed the workflow file from `pull_request.paths`
  - removed the explicit `permissions` block

  Verified locally that the workflow YAML parses and `npm run build` passes in `src/docs` under Node 24/npm 11.

reynaldichernando (approved):
LGTM

PR #3278, opened 2026-06-19
Add PUT and DELETE to preflight allowed methods

reynaldichernando:
- add put and delete to for workers preflight auto handling

i'm not sure about the mcp-connector code having `s2w-router.js` as well
is this part of workers? @ProgrammerIn-wonderland


ProgrammerIn-wonderland:
Mcp-connector is built on the same code that powers workers but is not a part of workers. It doesn't need the same change but we should keep it for consistencies sake

PR #3313, opened 2026-06-26
Add `forceRelay` option to Peer API and document raw `peerconnection` property

velzie:


velzie:
not sure if we need to make the reference to peerconnection more explicit? not sure if i want to provide a whole example for it though

Salazareo (approved):
lgtm

PR #3325, opened 2026-06-30
Add custom icon support for web links

Adcbda:
## Summary

- Add persistent custom icon support for `.weblink` files.
- Add a `Change Icon` context menu action that uses Puter's built-in file picker.
- Read selected icon files from Puter FS and store them as data URLs in the web link payload.
- Preserve legacy web links by falling back to the default link icon and upgrading plain-URL web links when edited.

## Details

Web link icons were previously hard-coded to `link.svg` in the item icon resolver. This change centralizes web link parsing and icon handling in `src/gui/src/helpers/weblink.js`, stores the selected icon in the `.weblink` JSON, and teaches `item_icon` to prefer that custom icon when rendering.

The icon picker accepts image files selected through Puter's file manager. Because files read from Puter FS may have a generic Blob MIME type, the helper normalizes data URLs using the selected file type, extension, and image byte signatures for PNG, JPEG, GIF, WebP, and SVG.

## Validation

- `npm run build --workspace src/gui`

The build passes with the existing webpack bundle size warnings.

### DIFF
--- a/src/gui/src/UI/UIItem.js
+++ b/src/gui/src/UI/UIItem.js
@@ -30,6 +30,7 @@ import truncate_filename from '../helpers/truncate_filename.js';
 import launch_app from '../helpers/launch_app.js';
 import open_item from '../helpers/open_item.js';
 import mime from '../lib/mime.js';
+import { changeWeblinkIcon } from '../helpers/weblink.js';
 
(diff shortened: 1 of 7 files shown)


Salazareo:
pushed some changes to prevent xss and down size some of the images for web links

PR #3421, opened 2026-07-21
initial pjs helper stuff

ProgrammerIn-wonderland:


src/puter-js/src/lib/PuterClient.js
```diff
@@ -0,0 +1,280 @@
...
+        });
+        xhr.addEventListener('abort', function () {
+            reject(signal?.reason ?? new DOMException('Aborted', 'AbortError'));
+        });
+
+        xhr.send(body);
+    });
+}
+
+export { fetchUrl };
```
Salazareo: nit: rename this file, doesn't really make sense for it to be PuterClient, makes it seem like a class

src/puter-js/src/lib/networkUtils.test.js
```diff
@@ -0,0 +1,222 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { fetchUrl } from './PuterClient.js';
+
+// -- Controllable fake XMLHttpRequest --
+// Drives fetchUrl's event handlers deterministically. Each instance plays the
+// `program` set on the class, which the test uses to script the response.
+function installFakeXHR (program) {
```
Salazareo, line 7: one thing I'd look out for here, I don't want to maintain a mock xhr.

like if there's some existing, tested mock out there, that might be better, but if not then we can keep this

src/puter-js/src/lib/utils.js
```diff
@@ -155,6 +155,64 @@ function replayXhrAfterReauth (response, success_cb, error_cb, resolve_func, rej
...
+ *   - `token_auth_failed` (legacy `APIError.create('token_auth_failed')`):
+ *     token no longer valid, prompt re-login (web env only).
+ *
+ * @param {Object} resp - The parsed response body.
+ * @returns {Promise<{action: 'replay'}|{action: 'reject', error: Object}|null>}
+ *   `replay` when the caller should re-issue the request once with the fresh
+ *   token, `reject` with the error to surface, or `null` when this is not a
+ *   reauth-recoverable 401 and the caller should handle it normally.
+ */
+async function resolveReauth (resp) {
```
Salazareo: might make sense to live in same place as fetch stuff

src/puter-js/src/lib/networkUtils.js
```diff
@@ -0,0 +1,280 @@
...
+
+    return new Promise((resolve, reject) => {
+        const xhr = new XMLHttpRequest();
+        xhr.open(method, url, true);
+        xhr.withCredentials = withCredentials;
+        // Text mode keeps NDJSON progress deltas available for stream(); callers
+        // that need binary/parsed bodies pass an explicit responseType.
+        xhr.responseType = responseType;
+
+        if ( includePuterAuth && globalThis.puter?.authToken ) {
```
Salazareo, line 230: what are the chances we can remove puter from globalThis? would that be a breaking change to things?

Salazareo (approved):
I think broadly good, some nit and questions there

src/puter-js/src/lib/networkUtils.js
ProgrammerIn-wonderland: hm, for this we could actually make PuterClient a class and then instantiate it and have a this.puter. What do you suggest here?

ProgrammerIn-wonderland: Internal: we decided this is an issue for a later time

PR #3425, opened 2026-07-22
Add loopback support to puter.peer

velzie:
Intentionally not documented (for now)


Salazareo (commented):
I know this is new, but can you get some type docs in here via jsdoc.

preferably in typescript type format, thought not too important as long as typed to best effort

```javascript
 async connect(\** @type {string} *\ invitecode, \** @type {{port:number} | undefined} *\ options = {}) {
```

PR #3433, opened 2026-07-23
feat: add Infron AI chat provider

meridah7:
## What

Adds [Infron](https://infron.ai) as a provider for the `puter-chat-completion` driver, following `doc/contributing-apis.md` and using the OpenRouter provider as the reference implementation (as suggested by the Puter team).

Infron is an AI inference platform with an OpenAI-compatible gateway (`https://llm.onerouter.pro/v1`) serving 400+ models across 100+ upstream inference providers.

## How it works

- **`InfronProvider`** mirrors `OpenRouterProvider`: OpenAI SDK pointed at the gateway, dynamic model catalog from `GET /v1/models` (kv-cached for 15 minutes). Two deliberate differences, both driven by the gateway's actual API shape:
  - The catalog endpoint requires authentication, so the fetch sends the configured key as a bearer token.
  - The catalog has no OpenRouter-style `pricing` object; prices are USD per million tokens (`min_prompt_price` / `min_completion_price`), converted to microcents per token (`× 100`). Non-chat catalog entries (image/video/embeddings/search) and display-only entries are filtered out via `category_type` / `is_display_only`.
- **Metering**: requests always send `usage: { include: true }`, which makes the gateway report an authoritative `cost` (USD) on the response. Like OpenRouter's `usage.cost` branch, that cost is billed as a single `billedUsage` line item with per-token costs zeroed. One wrinkle: Infron reports `cost` at the *top level* of the response rather than inside `usage` — non-streaming reads it off the spread completion, streaming merges it in via an `index_usage_from_stream_chunk` deviation. If `cost` is ever absent, metering falls back to catalog per-token pricing (cache reads billed at the full prompt rate there, so the fallback never under-bills).
- **Registration**: config key `infron` (`apiKey` + optional `apiBaseUrl`), added to the `AGGREGATORS` set so its aliases never shadow first-party providers in the model map — same treatment as `openrouter` and `together-ai`.
- Models are exposed as `infron:<vendor>/<model>` with the same alias scheme as OpenRouter. Default model: `infron:deepseek/deepseek-v4-flash`.

## Testing

- **Unit tests** (`InfronProvider.test.ts`, 13 tests): same harness as the OpenRouter suite — real `PuterServer` + live `MeteringService`, OpenAI SDK and axios mocked at the module boundary. Covers construction/base-URL override, authenticated catalog fetch, kv caching, category filtering, price-unit conversion, request shape (prefix stripping, `usage.include`, `stream_options`), both metering branches (top-level cost and per-token fallback, non-stream and stream), and moderation.
- **Integration test** (`InfronProvider.integration.test.ts`): env-gated on `PUTER_TEST_AI_INFRON_API_KEY`, exercises the real gateway; verified passing locally.
- Full `src/backend/drivers/ai-chat/` suite passes locally (428 tests).
- Response shapes (models catalog schema, top-level `cost` on stream + non-stream, usage field names) were all verified against the live gateway while writing this, not assumed from docs.

## Security pass

- The API key leaves the process only toward the configured gateway (SDK auth header + catalog bearer token); it is never logged.
- No new HTTP surface; provider is inert unless `providers.infron.apiKey` is configured.
- Errors from the gateway propagate through the existing driver classification path unchanged.

Happy to adjust anything to better match house conventions!

### DIFF
--- a/config.template.jsonc
+++ b/config.template.jsonc
@@ -287,6 +287,10 @@
             "apiKey": "",
             "apiBaseUrl": "https://openrouter.ai/api/v1"
         },
+        "infron": {
+            "apiKey": "",
+            "apiBaseUrl": "https://llm.onerouter.pro/v1"
+        },
         "together-ai": { "apiKey": "" },
         // Local Ollama. `enabled: false` skips the auto-probe at startup
         // (otherwise Puter logs ECONNREFUSED on every boot when no Ollama
--- a/src/backend/drivers/ai-chat/ChatCompletionDriver.ts
+++ b/src/backend/drivers/ai-chat/ChatCompletionDriver.ts
@@ -37,6 +37,7 @@ import { DeepSeekProvider } from './providers/deepseek/DeepSeekProvider.js';
 import { FakeChatProvider } from './providers/FakeChatProvider.js';
 import { GeminiChatProvider } from './providers/gemini/GeminiChatProvider.js';
 import { GroqAIProvider } from './providers/groq/GroqAIProvider.js';
+import { InfronProvider } from './providers/infron/InfronProvider.js';
 import { MiniMaxProvider } from './providers/minimax/MiniMaxProvider.js';
 import { MistralAIProvider } from './providers/mistral/MistralAiProvider.js';
 import { MoonshotProvider } from './providers/moonshot/MoonshotProvider.js';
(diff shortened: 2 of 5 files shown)


ProgrammerIn-wonderland:
.

Salazareo:
@meridah7 could you sign the CLA please

ProgrammerIn-wonderland (approved):

PR #3447, opened 2026-07-25
Rework permission requests: new dialog, working popup flow for websites

jelveh:
- Replace the UIWindow-based permission prompt with a standalone top-layer <dialog> (responsive, light/dark, app/site identity, input protection)
- Implement puter.ui.requestPermission for env=web: opens the GUI's /action/request-permission popup with pinned origin/source/msg_id, popup-closed detection, and a check-permissions polling fallback for crossOriginIsolated openers
- Move the GUI's request-permission action into postAuthActions so signed-out users sign in first; identify the app by opener origin, correlate responses with original_msg_id, close the popup after answering
- Always respond from the IPC handler so the SDK promise can't hang; normalize the result to a strict boolean (a failed grant no longer resolves truthy)
- Accept origin in /auth/grant-user-app and /auth/revoke-user-app, mirroring grant-dev-app (fixes puter.perms.grantOrigin/revokeOrigin)
- Add Playwright e2e coverage for both the desktop and popup flows; update e2e harness for the auth_token_v2 localStorage key
- Update types and docs (perms request methods now work on websites)


src/backend/drivers/ai-chat/ChatCompletionDriver.ts
```diff
@@ -216,10 +216,22 @@ export class ChatCompletionDriver extends PuterDriver {
...
 
     override onServerStart() {
         this.#registerProviders();
-        this.#buildModelMap();
+        // Not awaited — providers fetch model lists over the network and
+        // boot must not block on their latency. Shutdown awaits the
+        // tracked promise so provider I/O (and its error logging) never
+        // outlives the server; a rejection here must not become an
+        // unhandled one.
+        this.#modelMapReady = this.#buildModelMap().catch((e) => {
```
Salazareo: if this error then server should in fact not start I feel, otherwise we could silently not load any of our ai

Salazareo (changes requested):
just that one thing I don't think we should fail silently

src/backend/drivers/ai-chat/ChatCompletionDriver.ts
jelveh: reverted it

Salazareo (approved):

PR #3463, opened 2026-07-28
ci: attach provenance and SBOM attestations to the published image

kobihikri:
Hi, and thanks for Puter.

`.github/workflows/docker-image.yaml` publishes the image to `ghcr.io` on tag, but the pushed manifest carries no provenance or SBOM attestation. Someone pulling it has no way to check that it was built by this workflow, from this repository, at that tag.

For Puter that seemed worth raising because the image is the product as far as a self-hoster is concerned — people run it to host their own files and desktop, so the container is holding user data directly rather than sitting behind another distribution step that might vouch for it.

The change is two lines on the build step:

```yaml
                  push: true
                  provenance: mode=max
                  sbom: true
```

BuildKit attaches the attestation to the image manifest, so it travels with the image to GHCR without any extra plumbing — and importantly **no permissions change is needed**. Nothing has to gain `id-token`, and your existing `cache-from` / `cache-to` setup is untouched.

Consumers can then verify with:

```
docker buildx imagetools inspect ghcr.io/heyputer/puter:<tag> --format '{{ json .Provenance }}'
```

Two caveats worth knowing:

- `mode=max` records the full build including build args. Usually right for a public image; `provenance: true` gives a smaller record if any build arg has ever been sensitive.
- Attestations add an extra manifest to the index — GHCR handles this fine, but worth knowing if a mirror sits in front of it.

No SLSA level claimed; the attestation is what BuildKit produces.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow myself.

### DIFF
--- a/.github/workflows/docker-image.yaml
+++ b/.github/workflows/docker-image.yaml
@@ -87,6 +87,8 @@ jobs:
                   platforms: linux/amd64,linux/arm64
                   context: .
                   push: true
+                  provenance: mode=max
+                  sbom: true
                   tags: ${{ steps.meta.outputs.tags }}
                   labels: ${{ steps.meta.outputs.labels }}
                   cache-from: type=gha


Salazareo:
hi @kobihikri thanks for the contribution, could you sign the CLA please?

kobihikri:
**Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.

**Provenance is already there.** For *public* repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: *"Public repos: provenance attestations with `mode=max` are automatically added"*. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.

**The SBOM is genuinely new.** That part stands — the same page says *"SBOM attestations aren't automatically added to the image"*, and `sbom: true` is what enables them.

I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.

**So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.

Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.

Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

Salazareo:
> **Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**
> 
> I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.
> 
> **Provenance is already there.** For _public_ repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: _"Public repos: provenance attestations with `mode=max` are automatically added"_. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.
> 
> **The SBOM is genuinely new.** That part stands — the same page says _"SBOM attestations aren't automatically added to the image"_, and `sbom: true` is what enables them.
> 
> I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.
> 
> **So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.
> 
> Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.
> 
> Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

all good, we just need you to sign the CLA

kobihikri:
CLA is signed — the check is green now. Thanks for your patience with it, and for taking the correction without fuss.

One standing offer, since it's easy to lose in the thread: the PR title and description still describe this as adding provenance *and* SBOM, which — per my correction above — overstates it. The accurate framing is "adds an SBOM attestation; pins the provenance mode that's currently implicit." If you'd like, I'll rewrite the title and body to say that before you merge, so the claim doesn't land in the merge commit and changelog overstating what changed. Say the word and it's done in a minute; otherwise I'll leave it exactly as it is.

PR #3472, opened 2026-07-29
fix(ai-chat): preserve max_tokens 0 for Alibaba, DeepSeek, and Claude

Ayush7614:
## Summary
- Alibaba, DeepSeek, and Claude used `max_tokens || default`, so an explicit `max_tokens: 0` was replaced by the provider fallback (`1000` / model default).
- Switch to nullish coalescing (`??`) so zero is forwarded, matching MiniMax and the Completions providers fixed earlier.

## Why this is real
Callers that pass `max_tokens: 0` silently get a large default instead. No dedicated issue tracked this leftover for these three providers.

## Test plan
- [x] Regression tests fail before the fix (`expected 1000 to be 0`) and pass after for Alibaba, DeepSeek, and Claude
- [x] `npm run test:backend -- … -t "forwards max_tokens 0"` — 3 passed

### DIFF
--- a/src/backend/drivers/ai-chat/providers/alibaba/AlibabaProvider.test.ts
+++ b/src/backend/drivers/ai-chat/providers/alibaba/AlibabaProvider.test.ts
@@ -225,6 +225,21 @@ describe('AlibabaProvider.complete request shape', () => {
         expect(createMock.mock.calls[0]![0].max_tokens).toBe(256);
     });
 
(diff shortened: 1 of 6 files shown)


Salazareo:
thanks @Ayush7614 we'll review shortly and get it ready for merge if all is good

ProgrammerIn-wonderland (approved):
Good changes, Thank you!

PR #3473, opened 2026-07-29
fix(puter-js): ignore peer offers for unknown connections

Ayush7614:
## Summary
- `PuterPeerServer` guarded `setRemoteDescription` for missing connections, but still called `createAnswer()` on `undefined`.
- Guard the entire offer path; return the `#message` promise from `onmessage` so signalling errors are observable.

## Why this is real
Stale/unknown offer ids produced a `TypeError` and an unhandled rejection on the shared signalling websocket.

## Test plan
- [x] Unit test fails before the fix (`Cannot read properties of undefined (reading 'createAnswer')`) and passes after
- [x] `vitest run src/puter-js/src/modules/Peer.server.test.js` — 1 passed

### DIFF
--- a/src/puter-js/src/modules/Peer.js
+++ b/src/puter-js/src/modules/Peer.js
@@ -44,7 +44,7 @@ class PuterPeerConnectionErrorEvent extends Event {
     }
 }
 
-class PuterPeerServer extends EventTarget {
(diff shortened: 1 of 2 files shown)


Salazareo:
@velzie can you review plz

velzie (approved):

PR #3474, opened 2026-07-29
fix(puter-js): guard PWispHandler against unknown stream ids

Ayush7614:
## Summary
- `DATA` / `CONTINUE` / `CLOSE` frames for missing `streamMap` entries null-dereferenced and crashed the shared Wisp relay `onmessage` handler.
- Look up the stream first and ignore unknown ids (also guard a missing parse result).

## Why this is real
Late/duplicate frames after close, or hostile stream ids, could take down every socket sharing that Wisp connection.

## Test plan
- [x] New tests fail before the fix and pass after for unknown DATA/CONTINUE/CLOSE
- [x] `vitest run src/puter-js/src/modules/networking/PWispHandler.test.js` — 10 passed

### DIFF
--- a/src/puter-js/src/modules/networking/PWispHandler.js
+++ b/src/puter-js/src/modules/networking/PWispHandler.js
@@ -29,10 +29,18 @@ export class PWispHandler {
             };
             this._ws.onmessage = (event) => {
                 const parsed = parseIncomingPacket(new Uint8Array(event.data));
+                if ( ! parsed ) {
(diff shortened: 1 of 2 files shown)


ProgrammerIn-wonderland:
@r58Playz can you review

Salazareo:
@Ayush7614 ty for the PR, we're gonna move over to the epoxy client to handle network api in #3476  so will close this one

PR #3476, opened 2026-07-29
Revamp networking api with epoxy try 2

r58Playz:


r58Playz:
this needs puter.cafe to be fixed and a couple epoxy fixes

r58Playz:
make sure puter.cafe ~~has auth turned on and~~ is up to date before merging

PR #3499, opened 2026-08-03
feat: add Hoonify AI chat provider

jelveh:
Adds [Hoonify](https://hoonify.ai) — an OpenAI-compatible gateway for open-weights models — as a chat provider for the `puter-chat-completion` interface.

## What

- `HoonifyProvider`: OpenAI SDK against `https://api.hoonify.ai/v1` (overridable via `providers.hoonify.apiBaseUrl`), streaming with `include_usage`, tool calling, and Hoonify's `top_k` sampling extension via `custom.top_k`.
- Static model catalog (GLM-5.2, Gemma 4 31B Instruct, Qwen3.6-27B) priced from Hoonify's published catalog. Static rather than fetched because `GET /v1/models` carries no prices — a dynamic catalog would meter usage at zero cost.
- Registered as an aggregator so its aliases can never shadow first-party providers. Catalog ids are namespaced (`hoonify:<vendor>/<model>`); the lowercased ids map back to exact-case wire ids via a `wireId` field. `Qwen/Qwen3.6-27B` deliberately gets no vendor-scoped alias — it collides with an existing Alibaba alias, and a colliding alias drops the whole model from the routing map (a test pins this).
- Enabled by `providers.hoonify.apiKey`; disabled otherwise. No puter.js/type changes needed — model discovery is dynamic.

## Testing

- 23 offline unit tests (mocked OpenAI SDK, real in-memory PuterServer + live MeteringService): request shape, alias/wire-id resolution, non-stream + streaming output, metering costs, tool calls, error passthrough.
- Env-gated integration test (`PUTER_TEST_AI_HOONIFY_API_KEY`) — passes against the live API, as does a driver-level end-to-end completion. Live checks also confirmed the models endpoint matches the shipped catalog and that streaming emits a final usage chunk.
- Driver suites pass (44/44 incl. updated registration/cost-line assertions); full backend run shows no new failures vs a stashed-HEAD baseline.

No auth/permission surface changes; the provider only activates when a key is configured.

### DIFF
--- a/src/backend/drivers/ai-chat/providers/hoonify/models.ts
+++ b/src/backend/drivers/ai-chat/providers/hoonify/models.ts
@@ -1,0 +1,95 @@
+/*
+ * Copyright (C) 2024-present Puter Technologies Inc.
+ *
+ * This file is part of Puter.
+ *
+ * Puter is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import type { IChatModel } from '../../types.js';
+import { usdPerMToken } from '../../utils/pricing.js';
+
+const K = 1_000;
+
+/**
+ * IChatModel plus the exact-case model id Hoonify expects on the wire. The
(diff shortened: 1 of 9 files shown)


404oops (approved):
Seems like everything's in line.

src/backend/drivers/ai-chat/providers/hoonify/models.ts
```diff
@@ -0,0 +1,95 @@
...
+    // `qwen/qwen3.6-27b` is also an Alibaba alias for the same open model;
+    // sharing it puts this entry in Alibaba's bucket as a fallback route.
+    hoonifyModel(
+        'Qwen/Qwen3.6-27B',
+        'Qwen3.6 27B',
+        262_144,
+        65_536,
+        usdPerMToken(0.32, 3.2, 0.15),
+        ['qwen/qwen3.6-27b'],
+    ),
```
ProgrammerIn-wonderland, line 94: missing model thinkingmachines/Inkling-Small $0.50 in $1.20 out $0.10 cached input

ProgrammerIn-wonderland (commented):
Other than the missing model, code looks fine

PR #3501, opened 2026-08-04
Peer: allow anon token in place of authentication

velzie:


Salazareo (approved):
looks good to me on the surface, don't know if there's vuln testing we should do for this

PR #3541, opened 2026-08-11
fix(i18n): add Traditional Chinese app management strings

nrps9909:
## Summary

- translate the new app discovery and app request flows for Traditional Chinese users
- translate the desktop shortcut, AI availability, and app folder controls
- preserve both positional placeholders in the app-folder accessibility label

These 20 keys are actively used by the dashboard and item/taskbar UI but were absent from `zhtw.js`, so they fell back to English. This is a locale-only change.

## Validation

- `node --check src/gui/src/i18n/translations/zhtw.js`
- targeted English/zh-TW key and placeholder comparison for all 20 added keys
- `git diff --check`

`npm run check-translations` cannot currently start on the base revision because `package.json` points to an absent `tools/check-translations.js`; the targeted validator above covers the changed keys and placeholder parity.


Salazareo:
ty for the contribution @nrps9909

PR #3545, opened 2026-08-11
:wrench: PUR-1072: Flatten driver permissions to hardcoded values

jfcastro92:
## Summary

Driver permission checks resolved a constant by querying the database. Every `POST /drivers/call` ran two SQL queries — group membership, then the issuer lookup behind it — to answer a question whose answer is the same for every user in the system.

This deletes `hardcoded_user_group_permissions` and the entire group-keyed scan built on it, replacing it with a flat `default_user_permissions` floor resolved in memory. A user driver check now costs **zero** database round
trips. Every permission check in the system, `fs:*` included, also stops paying the wasted membership query.

## What was there before

`hardcoded_user_group_permissions` mapped issuer → group UID → permission, and `#scanHcUserGroupUser` resolved it by looking up the actor's group membership and then the issuer user for each entry. What it actually granted:

| grant | groups | production consumer |
|---|---|---|
| `service` | admin, user, temp | ✅ every `POST /drivers/call` (`DriverController.ts:272`) |
| `driver` | admin, user, temp | ✅ app permission dialog / implicit app map |
| `feature` | admin only | ❌ none — tests only |
| `local-terminal:access` | admin only | ❌ none — tests only |

Plus 19 descendant entries (`service:puter-kvstore:ii:puter-kvstore`, `driver:puter-kvstore`, …) that a parent already present in the same group subsumed, so no check could ever need them. They read as an allowlist while
being nothing of the kind.

The three groups are the only ones the map ever named, and the signup flow puts every account into one of them — so membership could not change the answer for `driver` or `service`. Two queries per check re-derived a constant.

## What this does

**Deletes the group-keyed map and everything that read it.** Gone: `hardcoded_user_group_permissions`, `#scanHcUserGroupUser`, `#composeGrantMap`, `#grantablePermissions`, `#universalUserGrants`,
`systemGrantsByGroupUid`, and both derived-set caches.

**Adds `default_user_permissions`** — a flat `{ driver, service }` floor in `hardcoded-permissions.js`, resolved directly in `scan()` for user actors. Roots only; a parent grant subsumes everything beneath it.

**Removes `registerSystemGrantForEveryone` / `registerSystemGrantForUsers`.** Both had zero production callers — they existed only to mutate the group-keyed map, and the users-only/everyone asymmetry has no meaning once grants aren't group-scoped.

**Drops the admin-only `feature` and `local-terminal:access` grants.** Nothing in `src/backend` reads either string. Anything that must depend on *who* the user is now belongs in the ACL as a `user_to_group_permissions` /
`user_to_user_permissions` row, which the scan's existing group and user-to-user paths resolve. The admin group's real `driver` ACL row (sqlite `0025_system-user.dbmig.js`, `mysql_mig_3`, `postgres_mig_1`) is untouched.

## ⚠️ Permissions widening — please review this section

**A user belonging to no group now holds `driver:*` and `service:*`.**

Users in any of the three groups *already* held every `driver:*` / `service:*` permission, including invented ones — all three grant the bare parents, and a parent grant subsumes everything beneath it. The check was never gating a
grouped user; it was re-deriving "yes" from the database on every call.

Group-less users are reachable in production: signup's group insert is best-effort and only `console.warn`s on failure (`controllers/auth/AuthController.ts:940`). Such an account is currently locked out of every driver call, permanently, with no recovery path. This repairs it.

Group removal is not used as a moderation lever — the only `removeUsers` callers are the temp→user promotion and the user-facing group API.

**Self-hosted deployments with no default group configured** also gain the floor. Previously the shortcut was derived from `default_user_group` / `default_temp_group`, so clearing both made every driver call fail closed; the floor is not group-derived, so it now applies regardless. Covered by a test.

**Derived actors do not inherit the floor.** An app-under-user is still gated by `default_implicit_user_app_permissions` — it holds a permission only if that map names it *and* its user holds it — and an access token still needs its own row. Both exclusions are load-bearing and both are covered by tests that fail
when the guard is removed (see Verification).

## Why the floor can't just be deleted

Asked and measured, not assumed. `service` is granted by **no** database row — no migration inserts one on any engine. With `default_user_permissions` emptied, 8 tests fail, and the breakage is not limited to user actors: apps lose driver access too, because `#scanUserAppImplied` requires the issuing user's reading to
have a terminal for the app's delegation to resolve. Relying on DB rows alone would mean new migrations granting `service` to all three groups — which puts the membership queries this PR removes straight back.

## Performance

Measured on sqlite with the scan cache forced cold and `read` / `pread` /`tryHardRead` instrumented. Each check run twice; steady-state (second) figures:

| check | before | after |
|---|---|---|
| plain user, driver permission | 2 queries, 1 membership | **0, 0** |
| app-under-user, driver permission | 3 queries, 1 membership | **1, 0** |
| plain user, `fs:*` | 6 queries, 1 membership | **5, 0** |

**The membership count is the number to trust.** It goes 1 → 0 on every path, including `fs:*`, which used to pay that query to search a map with no `fs` key. Totals are noisier: `SqliteDatabaseClient.tryHardRead` delegates to `read`, so
one logical query logs twice, and sequential checks share warm Redis caches, so totals shift with ordering.

No new query is issued on any path. On MySQL/Postgres the real saving is larger than these sqlite counts suggest: `tryHardRead` there races primary and replica, so a removed call is two round trips, not one.

## Verification

- **Full backend suite:** 5063 passed, 20 skipped, **0 failures**.
- **Typecheck:** `tsc -p tsconfig.build.json --noEmit` — 0 errors.
- **Lint:** eslint clean on all four changed files.
- **Mutation testing:** disabling the floor fails 8 tests; letting derived actors inherit it fails exactly the two guards that exist for it (`gives every app the default implicit driver permissions`, `a scoped token does not inherit the default user permissions`). This is how coverage of the new logic was established — v8's per-line attribution is unreliable under the project's esbuild transform.
- **Empty-floor A/B:** documents why the constant is required (above).
- **Semantic matrix:** 8 actor shapes × 8 permissions computed at HEAD and on the change. **7 of 64 rows differ**, and every one was predicted: 5 in the group-less shape, 2 in the admin shape (`feature`, `local-terminal:access`). Byte-identical for grouped users, temp users, both app shapes, and both token shapes. `fs:*` is false everywhere on both sides, so nothing leaked into the FS path.

## Not covered

- **Postgres has a pre-existing failure** in `PermissionService.test.ts` (`Hook timed out in 60000ms` on the `scan paths` suite). Confirmed present at HEAD with the change reverted — 42 passed / 28 skipped and the same timeout — so it is not introduced here. Postgres takes ~135s for that one file locally.
- MySQL not executed. No SQL is added and no query shape changed — the change only removes calls — but this is untested rather than inferred.
- No production load benchmark; the figures above are database round trips.
- `feature` remains in `default_implicit_user_app_permissions` and is now inert: an app can only hold what its user holds, and no user holds `feature`. Left in place rather than widen the diff into a second constant.


Salazareo (approved):
looks good, but some merge conflicts @jfcastro92 

can you also open a follow up ticket for us to grab after fs share to also get rid of the feature and terminal permissions, or if its easy enough just remove them here.

we don't really need those anymore, both fo those are deprecated concepts that no longer exist

PR #3546, opened 2026-08-11
feat: app user feedback system

jelveh:
## Summary

Adds `puter.ui.showFeedbackDialog()`, letting a user send feedback to an app's developer. In the **app** environment the Puter desktop renders the dialog; on a third-party **website** a puter.com popup hosts it (signing the user in first if needed). The message is stored in a new `app_feedback` table and emailed to the app owner's confirmed email — it never passes through the app's own code.

Feedback is **strictly opt-in per app** via a new `apps.feedback_enabled` column, which developers set through the existing update path:

```js
await puter.apps.update('my-app', { feedbackEnabled: true });
```

A real column (not an app-metadata key) is used deliberately: Dev Center saves the whole metadata blob, so a flag stored there would be silently erased on the next save.

## What's included

- **DB migrations** (sqlite `0065`, `mysql_mig_20`, `postgres_mig_9`): new `app_feedback` table + `apps.feedback_enabled` column. Parameter-free, idempotent DDL.
- **Backend** (layered): `AppFeedbackStore` (durable count queries) → `AppFeedbackService` (opt-in check, message normalization, abuse caps, best-effort owner email) → `AppFeedbackController` (`POST /app-feedback`, `GET /app-feedback/target`). New `app-user-feedback` email template using the escaping-safe `{{{nl2br}}}` triple-stash.
- **GUI**: `UIWindowAppFeedback` dialog (server-side opt-in pre-flight, canonical title + unique name shown to prevent impersonation, char counter, in-form errors, Escape / Cmd-Enter, always settles). Wired into `IPC.js` (app env) and a new `send-feedback` popup action in `initgui.js`.
- **SDK**: `puter.ui.showFeedbackDialog()` — IPC in app env, origin+source+msg_id-pinned popup in web env, resolve-only so it never hangs the caller. Types + docs updated.

## Security / abuse posture

- `requireUserActor` blocks app tokens, so feedback **cannot be submitted programmatically** on a user's behalf; `guiOriginOnly` keeps cross-origin browser pages out.
- App identity is derived only from the validated IPC sender (desktop) or the browser-attested opener origin (popup), **never from message contents**.
- The `send-feedback` popup action is in `NON_AUTH_POPUP_ACTIONS`, so hosting the dialog never delivers a token to the opener.
- Layered limits: route rate limits (per-user + per-IP), **durable DB-count caps** (per user/app/day and per user/day) that fail *closed* when the limiter backend is down, plus a per-app daily owner-email cap. Message capped at 4,000 chars after control-char stripping.
- Owner email is fully best-effort — an unconfigured transport, or an unconfirmed / unsubscribed / suspended owner, never fails the request or blocks storage. The sender's email is never disclosed (only their username, which the app already sees).

## Testing

- Backend feedback tests: **24 pass on both sqlite and Postgres**.
- GUI + root vitest suites and the type-check gate pass.
- New Playwright e2e spec (`showFeedbackDialog.spec.js`): **5/5 pass** — app submit/cancel/not-opted-in, gui-env returns false, and web popup reports false + never leaks a token; verified a real row lands in the DB. The adjacent `requestPermission` suite (shared popup/action code) still passes 33/33.

## Backward compatibility

Purely additive: a new SDK method, a new optional app attribute (default off), and new endpoints. No existing signature, response field, or error code changes.

## Entry points

- Apps call `puter.ui.showFeedbackDialog()` themselves.
- In **dashboard** mode, an opted-in app also gets a **Send Feedback** control in the app window's top-edge drawer (next to minimize/close), so a user can reach it without the app wiring up a button. It's rendered only for opted-in apps and opens the same dialog, targeting that app.


src/backend/services/feedback/AppFeedbackService.ts
```diff
@@ -0,0 +1,360 @@
...
+                ),
+                this.stores.appFeedback.countByUserSince(userId, since),
+            ]);
+            return (
+                userAppCount >=
+                    AppFeedbackService.PER_USER_APP_DAILY_LIMIT + slack ||
+                userCount >= AppFeedbackService.PER_USER_DAILY_LIMIT + slack
+            );
+        };
+        const tooManyError = () =>
```
Salazareo: nit: you can just prep an error and throw it as needed, method is unecessary:

const tooManyError = new HttpError(
                429,
                'You have sent a lot of feedback recently — please try again later',
                { legacyCode: 'too_many_requests' },
            );

Salazareo (commented):
generally looks good, one nit there 

is lacking some test coverage though; I think at least the service layer should get coverage since it has a lot of app logic, but ideally also the store

PR #3549, opened 2026-08-12
Remove redundant "(free)" suffix from OpenRouter free-tier models

reynaldichernando:


reynaldichernando:
oops sorry didn't mean to open it here

PR #3553, opened 2026-08-12
:sparkles: PUT-1412 File sharing backend api

jfcastro92:
Revives file sharing as a first-class feature: a user shares a file or folder with another account by email or username at `read` / `write` / `manage`, and the recipient finds everything shared with them under **Shared** in the file browser.

Puter shipped sharing once and removed the surface, leaving the machinery behind. `ACLService` and its enforcement were largely intact; **discovery, the write API, and the whole UI were missing**. Reviving them surfaced six correctness bugs in the surviving machinery, three of which are live on `main` today and independent of sharing — those are called out below.

Closes PUT-1412. Addresses PUT-1489, PUT-1490 (partly), PUT-1491, PUT-1492, PUT-1413, PUT-1414.

---

## ⚠️ Security callout

This PR is almost entirely permission logic. Reviewers should weight these:

**Fixes an existing privilege-retention bug.** `deleteUserUserPermByHolder` had **no issuer clause**, so a revoke deleted *every* issuer's grant on that permission — and, symmetrically, a `manage` delegate could delete a grant the **owner** issued. Now scoped: an owner may revoke any grant on their entry, a delegate only what they issued, and anyone may drop their own access. Regression test written first, fails without the fix (`ba48efd28`).

**Revocation now cascades.** Revoking a `manage` delegate also revokes what that delegate re-shared, recursively, with a cycle guard. Previously A→B→C left C with working access after B was removed (`06248b667`).

**Signed URLs were permanent.** `signFile` defaulted to `ttlSeconds = 9_999_999_999_999`, and `signEntryThumbnail` mints one per readdir entry — so a revoked recipient kept a working URL forever. Non-owner signatures now expire after **3600 s** (`NON_OWNER_SIGNATURE_TTL_SECONDS`). See *Accepted risks* — this is a bound, not revocation.

**Denials are not existence oracles.** Every share path reports through `getSafeAclError`, so a caller who cannot see an entry gets 404, not 403. Covered by tests for both `getShares` and self-revoke.

**Apps cannot exceed their user on a shared path.** An app-under-user actor is bounded by its user even with its own grant; grants are uuid-keyed, so rename/move cannot widen an app's reach; a shared AppData directory stays with the app it belongs to. Asserted rather than assumed (`ACLService.test.ts:837, 851, 883, 906`).

**Responses carry no internal identifiers.** Recipients are sanitized to `{ username }`; no user IDs, no e-mail addresses, no foreign paths.

---

## User-visible behaviour changes

1. **The filesystem root no longer aggregates issuer homes.** `rootListing` previously listed the home directory of anyone who had granted you something, which produced folders you could see but not open. The **Shared** folder replaces it (`e7c856dd3`).
2. **Non-owner signed URLs now expire** (see above). Already-issued signatures remain valid until their embedded expiry.
3. **Delete on an item shared with you is now "Remove from Shared."** Deleting used to attempt a move into your trash, which the backend refuses with a raw 403. It now drops your own access instead. True delete is hidden for non-owners.

---

## What's in it

**Data layer.** The dormant `share` table is extended into an index of active shares — `holder_user_id`, `fsentry_id`, `mode`, `applied_at`, with `KEY (holder_user_id, id)` for keyset pagination, `KEY (fsentry_id)` for the reverse lookup, and a unique key on `(holder_user_id, fsentry_id, issuer_user_id)`. `recipient_email` stays, so link sharing (PUT-1497) still fits. Paths are **not** denormalized — `fsentry_id` is stored and the display path resolved at read time, so shares survive rename and move.

**`ShareService`** owns the invariant that a share is a permission write *and* an index write together. The permission goes first; if the index write then fails because the entry died mid-flight, the grant is rolled back rather than left standing invisibly. Authorization reuses `PermissionService.canManagePermission` — no new authorization concept.

**`ShareController`** exposes `POST /share`, `POST /share/revoke`, `GET /share/shared-with-me` (keyset-paginated per `doc/pagination.md`), `GET /share/shares`.

**puter.js** gains `puter.fs.share()`, `unshare()`, `listShared()`, `getShares()`, with JSDoc `@overload` signatures and docs under `src/docs/src/FS/`.

**GUI** gains a Shared sidebar entry, a share dialog with a per-row mode picker and remove control, and context-menu entries. Access inherited from a parent folder is shown greyed with "via {folder}" and no controls, because it is managed on the folder, not the item.

**Anti-abuse.** 200 shares/user/day (`share_daily_limit`), 10 recipients and 50 items per request (`share_max_recipients`, `share_max_items`). Quota counts *creations*, so revoking and re-sharing cannot recycle a slot; changing an existing share's mode is not new reach and does not spend budget.

**Concurrency.** `setUserUser` was a non-atomic stat → compare → grant → revoke, so two concurrent calls on the same (holder, entry) could persist both modes or neither — a pre-existing race the share dialog makes easy to hit. Now serialized under a short Redis lock per (holder, node).

**Cross-region.** `bumpCacheGeneration` was a raw `redis.incr` that emitted nothing, and no `outer.cacheUpdate` applier existed, so permission invalidations never left their region. Adds the applier and the `outer.permission.*` events (`d8380f4a4`, `abce4949e`). Intra-cluster was already correct — the generation counter is in shared Redis and every scan-cache key folds it in.

## Migrations

Three dialects, additive only: `sqlite/0067_share_entries.sql` (schema version → 63), `mysql/mysql_mig_22.sql`, `postgres/postgres_mig_11.sql`. The MySQL one is written idempotently via `_puter_add_col` and `INFORMATION_SCHEMA`-guarded procedures, because the runner has no applied-state tracking and every statement must tolerate a re-run.

## Testing

- Full backend suite green: **5699 passed, 21 skipped**.
- `npm run typecheck` clean against the baseline; `npm run check:puterjs:types` clean.
- New: `ShareService.test.ts` (25 cases), `ShareController.http.test.ts`, `CacheReplicationService.test.ts`, `sharing.suite.ts` for the SDK across runners.
- Each of the six bug fixes has a regression test that fails without the fix.
- **Run end-to-end with two local accounts**: share a file, confirm it appears under Shared for the recipient, open it, confirm `write` can write and `read` cannot, revoke, confirm it disappears. Chained revocation (A→B→C) verified manually — `revoked: 3`, both recipients 404 afterwards.

**Not covered, deliberately:**
- No automated GUI coverage — the dialog and sidebar are manual-only, consistent with the rest of the GUI tree.
- The end-to-end run predates the merge of ~28 commits from `main`; `TabFiles.js`, `UIItem.js`, `UIWindow.js` and `style.css` auto-merged into the sharing work without conflict, and auto-merged is not the same as visually correct. Worth a second manual pass before merge.
- `inheritedFrom` has unit coverage but no end-to-end assertion through the SDK.

## Accepted risks / follow-ups

1. **Signed-URL exposure is 3600 s, not immediate.** `verifySignature` takes no actor and never consults the ACL, so a URL fetched before revocation keeps working until it expires. A per-entry signature epoch is the durable fix and is not in this PR. Reviewers are agreeing to a number here, not to "short".
2. **Cross-region relies on DynamoDB last-writer-wins.** `store-kv-v1` **is** a Global Table (confirmed with @dsalazar), so a revoke does replicate — the "indefinite access" concern is resolved. What remains is LWW conflict resolution, and the sharp case is **read-vs-revoke**, not grant-vs-revoke: a `DELETE` in one region loses to any later-stamped `PUT`, and the cache-warm write at `PermissionService.ts:708` is exactly such a PUT, fired by an ordinary read. A scan in the peer region against its own not-yet-converged SQL view can re-create a revoked key, and because a flat hit is terminal (`#flatValidateUserPerms` never corroborates against SQL) the resurrected key wins. The exposure is **bounded, not permanent**: warm writes carry a 60 s TTL (`FLAT_PERM_WARM_TTL_SECONDS`) precisely for this reason, so a resurrected key ages out and the next scan re-derives from SQL. Clock skew picks the winner, and the asymmetry runs the wrong way — one direction drops a grant (safe), the other resurrects a revocation.

   Mitigations, none of which are in this PR: tombstone-on-revoke instead of delete (`FlatPermValue.deleted` already exists and the read path already honours it), a TTL on grant keys (`setFlatUserPerm` already accepts `expireAt`; the grant path at `:857` just doesn't pass one), and a cross-region reconcile job. Single-region behaviour in this PR is correct and tested.

3. **The `setUserUser` lock fails open** when Redis errors, so a Redis outage restores the pre-existing race rather than blocking writes. Deliberate — availability over strict serialization on a path that was previously unserialized entirely.
4. **The SQL fallback in the permission scan is still present.** PUT-1490 asks for its removal or a concrete removal plan; the plan exists but is not in this PR, and `#linkedValidateUserPerms` remains in the read path.
5. `fsentry.is_shared` is still read by the GUI (`helpers.js:1987,2020`) and still never populated by the backend. Unchanged by this PR; needs to be populated or the reads deleted.

## Cross-store consistency evidence

`src/backend/services/share/ShareConsistency.test.ts` snapshots all four layers — flat KV, `user_to_user_permissions`, the `share` index, and the Redis cache generation — before and after each of nine scenarios with three users, and asserts the effective `acl.check` alongside the rows. `FILE-SHARING-CONSISTENCY-EVIDENCE.md` is its generated output.

It found a real bug, now fixed in this PR: **deleting a shared entry retired the rows in both stores but never bumped the holders' cache generation**, so recipients kept a cached "allowed" answer after the file was gone. `unshare` bumps; the delete path did not. Both stores read clean while `acl.check` still returned `true` — which is why the plan's invariants now assert the effective answer, not just the rows.

## Note on scope

Two commits at the base of this branch — `3eb9d64bb` (drop hardcoded group permission map) and `26841a9bd` (driver credential-gate test) — belong to **PUT-1072**, not to this ticket. They are here because this branch was cut from the PUT-1072 branch. If that work merges first they will drop out on rebase; otherwise they need review here.


Salazareo (approved):
fixed some last few things about the path obfuscation, and some gui issues with claude, but think this looks ready to go

PR #3563, opened 2026-08-13
feat: let the Contact Us form carry screenshots and recordings

jelveh:
## What

The Contact Us form now takes up to 5 images or videos alongside the message. They're delivered as attachments on the support email.

## Why

From a user report:

> I think the "Contact Us" form should allow attachments, especially since it was also created for reporting bugs. This would be useful for including images and videos for bugs that aren't so obvious and that require multiple steps to reproduce.

That's the case exactly: the form doubles as our bug-report channel, and a visual glitch or a five-step repro is far cheaper to show than to describe.

## Security

Everything below is enforced server-side, on the decoded bytes. The client-side checks in `helpers/contact_attachments.js` exist only to fail fast before a 40 MB upload; they're not a control.

**Type.** Sniffed from magic numbers and matched against an allow-list: PNG, JPEG, GIF, WebP, MP4, QuickTime, WebM. A caller-declared MIME is never read, so it can't be used to smuggle anything past the list. Two deliberate exclusions:

- **SVG** — script-capable, and support tooling renders what it's handed.
- **Non-video ISO-container brands** that share MP4's `ftyp` box: HEIC, `M4A `, JPEG 2000. The brand decides, not the box.

**File name.** Reduced to a display label; the extension is re-derived from the sniffed type, so PNG bytes named `payload.html` arrive as `payload.png` and nothing can ever come out `.exe`. Stripped: path components, C0/C1 controls (a CR/LF in a name would break out of `Content-Disposition`), bidi overrides (`report<RLO>gnp.exe` → `reportgnp.png`), quoting characters, leading dots.

**Size.** 10 MB per file, 15 MB per submission, 5 files. Counted on decoded bytes — the encoded length is capped *before* decoding, so an oversized payload costs a length check rather than a 10 MB allocation. The total sits well under the 25 MB ceiling most mail providers enforce, since the outgoing mail base64s these again.

**Encoding.** Strict base64, reusing the round-tripping decoder that already guards app icons. `Buffer.from(s, 'base64')` silently drops characters it doesn't recognise, so `<png>" onerror=alert(1)` decodes without complaint; the round-trip is what rejects it.

**Rate.** One request is now worth megabytes of parsing and outbound mail. The existing 10/15min per-user limit gains a 40/24h per-IP backstop (the per-user counter can't see a single machine cycling fresh accounts), `concurrent: 2` per user, and a `Content-Length` gate that 413s an impossible body before anything decodes.

**Storage.** Payloads are not stored. They ride the email with `Content-Disposition: attachment` (no inline rendering). A new `feedback.attachments` column records `{name, type, size}` only — enough to keep an abusive submission attributable once the mail has been dealt with, without parking megabytes in the database.

Two things worth a maintainer's call:

- `allowFullAccessToken: true` stays on this route, so an app holding a full-access token can submit attachments programmatically. That's pre-existing for the message field; I left it alone rather than risk breaking a first-party caller, and the allow-list plus caps bound what it can do. Happy to narrow it if you'd rather.
- HEIC is rejected, which means iPhone photos are refused if Safari ever stops transcoding them to JPEG on upload. Worth revisiting if reports come in.

## Notes on the diff

- The strict base64 decoder and the image sniffer move out of `util/appIcon.ts` into a new `util/mediaSniff.ts`, which gains the video counterpart. `appIcon.ts` re-exports `sniffImageMime` so nothing downstream changes.
- Migrations add a nullable `attachments` text column across all three engines (sqlite 0067, mysql 22, postgres 11); the sqlite schema-version constant in `SqliteDatabaseClient.test.ts` moves with it.

## Testing

51 new unit tests (42 for the validator, 9 for the GUI helper) plus 8 controller cases covering the allow-list, the size and count caps, malformed entries, and each file-name attack.

Also driven end to end against a local dev backend, through the real desktop dialog:

- Form renders; picker staging works; a dropped SVG and PDF are rejected with the right message while the usable files in the same drop are kept.
- Count cap disables the Attach button; removing a chip re-enables it.
- Drag-over highlights and prevents the browser navigating to the file.
- Submit → 200 → correct `feedback` row.
- Composed message captured off an SMTP sink: two `image/png` parts, `Content-Disposition: attachment`, sanitized filenames, and a body manifest so the recipient can tell a stripped attachment from one that never existed. A file name carrying `\r\nBcc: victim@example.com"; x="y` produced no `Bcc` header.

Full backend suite is green apart from 30 pre-existing failures that reproduce on a clean tree (node can't resolve `puter.localhost` on this machine).

## Also fixed in passing

- The form used to POST an empty message when Send was pressed with nothing typed.
- Submit failures left the button disabled with no explanation; they now surface an error, which matters much more once a request can fail for being too large.


Salazareo (changes requested):
I'm concerned about buffering bytedata here like this. could quickly fill up memory.

@ProgrammerIn-wonderland the work youre doing with streaming some files, how generic is it? could we use that here

PR #3595, opened 2026-08-17
:sparkles: PUT-1497: Share file link sharing and notifications

jfcastro92:
Builds on the sharing API from #3553: telling people what was shared with them, reaching someone who has no account yet, and keeping either from becoming a way to spam.

## What this adds

**Invites.** Sharing with an email address that has no confirmed account no longer fails. The invite is recorded (`share` row with a null `holder_user_id`) and emailed; it becomes a real grant when the address is confirmed, re-authorized at that moment because an invite can sit for weeks and the issuer may have lost the right to share it since. Pending invites are visible in `getShares()`, cancellable via `unshare()`, and shown in the share dialog.

**Email digests.** Sends to one recipient are held for a window (90s, configurable) and merged into a single message — five people sharing within a minute is one email that names them, not five. The queue is durable and cross-region safe: each pending send is one KV/DynamoDB entry, the flush takes a region-local Redis lock against same-region stampedes and claims each entry with a `DeleteItem`-returning-`ALL_OLD` (`SystemKVStore.take`) so racing regions can't double-send, and a failed send re-queues what it claimed. `onServerPrepareShutdown` drains what is pending while the transport is still up, and a recovery sweep picks up entries whose timer died with the node that armed it — a restart or a rolling deploy would otherwise leave them until their TTL with nobody to send them.

The budget is spent **opening** a digest, not per share, since a digest is one email: anything arriving while one is already collecting joins it for free. The wording names up to three files per sender and counts the rest — one file reads "alice shared report.txt with you", four reads "4 items — a, b, c, +1 more". Templates were reworked to suit a digest with several senders, and every gate that can silence mail (notifications off, no transport, unconfirmed or unsubscribed recipient, address refused, budget spent) now logs why, because in a region a silent early return is indistinguishable from a lost email.

**Announcement budgets.** Announcements are bounded on two axes through the existing sliding-window limiter: 1 per 15 minutes and 20 per day from one sender to one recipient, and 10 per hour / 50 per day to one recipient from anyone. The recipient axis is the one that matters — a per-pair limit alone lets twenty accounts each spend a full budget on the same person. Over budget, **the share still succeeds**; the notification push is dropped, and email is only prevented from opening a *new* digest — a suppressed share still joins one already collecting, so the mail never undercounts what was shared. Invite email is budgeted the same way, keyed on a hash of the canonical address.

**Grouping.** A new share rewrites the notification the recipient hasn't dismissed instead of stacking another, so five people sharing with you reads "alice and bob shared 5 items with you". The row is updated even when the interruption is suppressed, so the count is right whenever they next look; the GUI refreshes a toast in place when a notification arrives under a uid it is already showing.

**Blocking.** A recipient can refuse a specific sender. A blocked sender's share is refused, spends no daily quota, writes no row, and their unclaimed invite is dropped when the address is confirmed. Access the sender already granted is untouched — that is what `unshare` is for. Managed from a Blocked people card in the dashboard's Security tab.

**Two share-dialog fixes.** The revoke confirmation rendered *underneath* the dialog that opened it in fullpage/dashboard mode (UIWindow promotes a parentless window to the stay-on-top band; UIAlert opts out explicitly), leaving no way to confirm without closing the dialog. And `i18n()` already HTML-encodes, so wrapping it in `html_encode` again displayed "Can edit &amp; share".

## Wire changes

All additive — no signature, field, or error code is renamed or removed, and existing callers are unaffected.

| Change | Detail |
| --- | --- |
| `POST /share` response | Now carries **`name`** (the item's own name) on each result. The path a recipient sees is masked, so the name is what labels the share — and the same value is what the share email needs to say which file it is, instead of the literal "an item". |
| `POST /share` response | An invite carries **`pending: true`** and **`recipient_email`**, with a `null` `holder`. |
| `POST /share` error codes | New **`recipient_not_accepting_shares`** (403) when the recipient has blocked the sender. Reported plainly rather than disguised as a missing recipient so the sender stops re-sending; it runs *after* the existing manage check, so only someone entitled to share can observe it. |
| `POST /share` error codes | New **`email_not_allowed`** (400) for an address that could never receive its invite. The one behavior change here rather than an addition: `share(path, 'a@b')` now fails where it used to resolve with a `pending` share, because an unusable address became a permanent invite row that spent daily quota and nobody could ever claim. |
| Docs | `FS/share.md` gains an **Errors** section — the page previously documented none of the share codes, new or existing. |
| New endpoints | **`GET/POST/DELETE /share/blocks`** — list, add, remove. Same gates as the other share routes (`api` subdomain, verified account, share rate limits). Usernames only in responses; no internal ids. |
| puter.js | `Share.name` is now set by `share()`, not only `listShared()` — the typedef said otherwise and has been corrected. `toShare()` already mapped the field, so no SDK code change was needed. No new methods; the blocklist has no SDK surface yet (deliberately deferred — see below). |

**Internal, not public API:** `EmailClient.isConfigured` is gone — callers ask `config.email`, which is what the getter was reporting on. `SystemKVStore.take()` is new (a delete that returns the value to exactly one caller, via `DeleteItem` `ALL_OLD` plumbed through `DDBClient.del`'s optional `returnOld`); it is not exposed on the KV driver, so `puter.kv.*` is unchanged.

## Config

New keys, all optional, documented in `config.template.jsonc` and `doc/self-hosting.md`:

- `share_email_notifications` — email a recipient who already has an account. **On unless set to `false`.** Recipients decline with the unsubscribe link the mail carries (honoring `user.unsubscribed`, which the existing `/unsubscribe` page writes and app feedback already respected but share email did not), or per-sender by blocking them. An address with no account is emailed either way; there is nothing else to reach them with.
- `share_notify_limits` — `{ pairWindowSeconds, pairDaily, recipientHourly, recipientDaily, emailBatchSeconds }`. Omit any field for the built-in default; a non-positive value removes that bound.

## Migrations

Additive, no backfill, one per dialect: `sqlite/0068_user-block.sql`, `postgres_mig_12.sql`, `mysql_mig_23.sql`. New `user_block` table (`blocker_user_id`, `blocked_user_id`, `created_at`) with a unique index on the pair, which serves both the share-path check (point lookup) and the listing (range scan on the leading column). Named generically rather than `share_block` so anything else one person can push at another can reuse it.

`src/docs/src/rate-limits-and-quotas.md` gains a Sharing section — it had none, so the limits already shipped in #3553 (`share:mutate` 60/min, `share:mutate-daily` 500/day, `share_daily_limit` 200/day) were undocumented too. They are published here alongside the new announcement budgets.

## Security notes

- **Recipient existence is not probeable.** Authorization runs before recipient resolution, so "recipient does not exist" — and the new "not accepting shares" — can only be observed by someone already entitled to share the item.
- **Blocking is enforced at both moments that matter**: when the share is issued, and when a pending invite is claimed.
- **Outbound mail to third parties is bounded.** Invite email reaches someone who never asked to hear from us, so it is rate-limited per (sender, address) and per address, with the address hashed rather than stored in cache keys.
- **Responses stay curated.** No internal ids in share or blocklist responses; asserted in the HTTP tests.
- **The digest holds addresses at rest, briefly.** A pending entry carries the recipient's address, username and uuid in KV until it is sent, with a one-day TTL as the ceiling. Budget keys never carry an address — those are hashed. Nothing is written for a recipient who is over budget with no digest open.
- **Recipients can decline.** Share email honors `user.unsubscribed` (which app feedback already respected and share email did not) and carries the unsubscribe link, so turning it on by default doesn't leave anyone without an exit.

## Testing

`npm run test:backend` green (236 files, 6,078 tests), `npm run typecheck` clean, `npm run check:puterjs:types` clean, puter.js node suite green (427).

New coverage: notification budgets and grouping (including that a suppressed announcement still updates the row), `shareNotifyTitle` for one/two/many senders, blocked-share enforcement and claim-time drop, `/share/blocks` over HTTP, and `UserBlockStore` against both sqlite and Postgres (`PUTER_TEST_DB_ENGINE=postgres`) so both migrations are exercised.

`shareEmail.test.ts` covers what the mail actually says by spying on `EmailClient.sendRaw`, which receives the fully rendered subject and body: every template renders for real, and the signup confirmation code is read out of the captured message, so the invite → confirm → claim journey runs the same path a person would. It caught the "an item" bug fixed in b75b0f9. (An in-process SMTP sink was tried first and dropped — 829 lines and 11.5s versus 283 and 3.4s, to prove only that nodemailer accepts a payload. MailHog stays a manual-testing tool, documented in `doc/self-hosting.md`.)

### Found by deploying it

Two bugs surfaced in a region that no test had caught, both now fixed with regression tests:

- **A four-file share emailed one filename.** Each file arrived as its own call; the first spent the pair budget and the other three were refused *before* being queued, so the digest honestly reported the single entry it held. Fixed by spending the budget on opening a digest rather than per share (above). Separately, the per-recipient name map overwrote instead of accumulating, so even a single multi-file share carried only the last name — the invite path had the same bug.
- **A digest could be stranded by a deploy.** The window's timer lived only in the process that armed it, so a restart left entries in KV that nothing would flush. Hence the recovery sweep.

The counting fix is mutation-checked: reverting the join rule makes the new test fail with the exact symptom reported (`expected 'testuser shared one-01681257.txt with you' to be 'testuser shared 4 items with you'`).

### Verified by hand against a local instance

Shared one file from `jfcastro` to `jfcastro9208` (both real accounts with confirmed addresses) through the share dialog, with `share_email_notifications: true` and the SMTP transport pointed at a local MailHog. What arrived:

```
From:    "Puter (local)" <no-reply@puter.localhost>
To:      <recipient>
Subject: jfcastro shared Screenshot 2026-08-17 at 10.30.17 AM.png with you
```

```html
<p>Hi there jfcastro9208,</p>
<p>jfcastro shared <strong>Screenshot 2026-08-17 at 10.30.17 AM.png</strong> with you on Puter.</p>
<p><a href="http://puter.localhost">Open it on Puter</a></p>
<p>Sincerely,</p>
<p>Puter</p>
```

The filename in that subject is the fix in b75b0f9 — before it, the same share read "jfcastro shared **an item** with you". The subject arrives RFC 2047 encoded (`=?UTF-8?Q?…`) because macOS screenshot names carry a narrow no-break space before "AM"; nodemailer encodes non-ASCII headers per spec and clients decode them, so this is the wire form, not what the recipient sees.

The in-app notification landed in the same action, delivered to the open session, carrying the grouping state a second sender would fold into:

```json
{
  "title": "jfcastro shared an item with you",
  "fields": {
    "username": "jfcastro",
    "count": 1,
    "senders": [{ "username": "jfcastro", "count": 1 }],
    "groupUntil": 1786995683787
  }
}
```

### UI Bugs fixed

- Encoding on the permissions list picker ( & instead of &amp)

image

- When removing a permission from an existing record:

image

## Deliberately not in this PR

- **puter.js surface for the blocklist.** The dashboard talks to `/share/blocks` directly, the way session management already does. An SDK method carries docs pages, JSDoc types and a suite on node/browser/workerd; worth doing when an app needs it.
- **Per-feature email preferences.** A recipient can turn off *all* Puter email (the account-wide unsubscribe) or block an individual sender, but not "no share email, keep the rest". That needs a preferences surface this PR doesn't add.
- **The review's lower-severity findings.** `share.recipient_email` has no index on sqlite where mysql and postgres do (the three new query sites full-scan a small table there); `unshare` by an email that resolves to no confirmed account now answers `200` with `revoked: 0` where it used to `404`; and two `i18n()` call sites in the share dialog still double-encode their replacements, the same pattern fixed two lines above them.


Salazareo (approved):
pushed few fixes for gui + the block all functionality, but looks good now, feel free to merge and deploy after fixing merge conflicts

PR #3598, opened 2026-08-18
feat(dashboard): rewrite Files-tab sharing as a from-scratch modal

jelveh:
## What

Replaces the desktop `UIWindowShare` window with a from-scratch modal (`UIShareModal`) when sharing files from the Dashboard's Files tab. Built on the same overlay pattern as the item-properties modal: a centered card on desktop, a bottom sheet on mobile (≤600px), styled entirely with the dashboard's design tokens so dark mode works out of the box.

## How it's wired

Follows the `onShowProperties` precedent: `generate_file_context_menu` gains an optional `onShare` hook that TabFiles supplies. Every other caller falls back to the desktop share window — verified on `/desktop` that the original dialog still opens there, unchanged.

## Behavior

Feature parity with `UIWindowShare` — grant by email/username with an access level, list who has access (owner, direct grants, and grants inherited from an ancestor folder), change a grant's mode, revoke — plus:

- Revoke confirmation is inlined into the row (Cancel / Remove) instead of stacking a desktop `UIAlert` over the dashboard.
- Real form semantics: Enter submits, the Share button is disabled until there's input and shows a spinner in flight; API errors keep the typed name in place for correction.
- Dialog a11y: `role=dialog` + `aria-modal`, focus moves to the input on open and returns on close, Tab is trapped inside, Escape closes (the first Escape cancels an open revoke confirmation instead), and status messages announce via an `aria-live` region.
- Backdrop close is keyed on where the press *started*, so a text-selection drag out of the input can't dismiss the dialog and eat typed input.
- Name-derived avatar colors, a loading state for the first fetch, and 16px control fonts on phones to avoid iOS focus-zoom.

## Also fixed

A double-encoding bug shared with `UIWindowShare`: mode labels went through `i18n()` (which html-encodes) and then `html_encode()` again, rendering the manage option as "Can edit &amp;amp; share". `mode_label` now returns raw text and encoding happens exactly once at render time, in both files.

Adds one i18n key (`share_access_level`, used for the selects' aria-labels); other locales fall back to English.

## Testing

- GUI vitest suite: 296/296 passing.
- Live browser walkthrough against a local backend: grant, mode change (persists through server refresh), revoke cancel + confirm, error path ("Recipient does not exist"), Enter-to-submit, all three dismissal routes (X / Escape / backdrop), mobile bottom-sheet layout at narrow widths, and the `/desktop` route to confirm the desktop share window is unaffected.


Salazareo:
probably best to wait till the last PR is in, otherwise might have conflicts

PR #3607, opened 2026-08-19
🛠️  PUT-1521: Cleanup puter js permissions api + backend routes

jfcastro92:
Two piles of unused surface had accumulated around permissions, and the `request*` family had grown one method per task. This removes the former, and replaces the latter with a single front door rather than a shorter list of the same shape.

**1. User↔user and user↔group permission plumbing that nothing called.** Filesystem sharing moved to `puter.fs.share()`, which records a share row so the owner can see and revoke it. The older direct-grant paths were left behind: `/auth/grant-user-user` was already a 501 stub, and `grant/revoke-user-group` plus the five `/group/*` CRUD routes had no caller anywhere — not the GUI, not a doc, not an app. `puter.perms` layered 8 undocumented wrappers on top of them.

**2. One `request*` method per task.** 15 methods differed only by a folder name or an access level (`requestReadDesktop`, `requestWriteDesktop`, … `requestManageSubdomains`). Every new resource meant another method, another doc page, another sidebar entry.

**Supported `puter.perms` surface: 33 → 8.** Two permission methods (`request`, `check`) plus the six app/origin grant methods. Eight backend routes gone, along with the service and store methods they alone fed.

## puter.js: one `request()`, one `check()`

The resource is the first argument and it decides which payload fields are accepted — the shape `emit(name, data)` uses, so a caller writes one call regardless of what they are asking for:

```js
await puter.perms.request('email');                                         // → address
await puter.perms.request('folder', { name: 'Documents', access: 'write' }); // → path
await puter.perms.request('apps', { access: 'read' });                       // → boolean
await puter.perms.request('subdomains', { access: 'write' });                // → boolean
await puter.perms.request('appData', { app: 'contacts', scopes: 'read' });    // → boolean
await puter.perms.request('appRootDir', { app: uid });                        // → fs item
await puter.perms.request('permission', { permission: 'fs:<uid>:read' });     // → boolean
```

Returns stay per-resource, so a call gives back what the caller actually needed rather than a boolean plus a second call to go get it. Everything denied is falsy, so one `if` covers both outcomes either way.

`PermsBatchEntry` is a discriminated union and each resource gets its own `@overload`, so `request('apps', { name: 'Desktop' })` is a type error rather than a silently ignored field. The generated `.d.ts` carries all sixteen overloads.

### `check()` — the same question, without a prompt

```js
if ( ! await puter.perms.check('folder', { name: 'Documents', access: 'write' }) ) {
    showEnableBackupsButton();
}
```

Takes the same resource and details as `request`, and never opens a dialog. This is what lets an app offer an opt-in only where one is needed instead of prompting on load. Built on `/auth/check-permissions`, which is already live and already used internally by `UI.js`, so no new backend surface.

Two deliberate choices:

- **A failed check throws; it does not answer `false`.** A caller that cannot tell "denied" from "the check never ran" would prompt someone who had already granted it.
- **A set is held only when all of it is.** `'appData'` with several scopes, or `'permission'` with a list, answers `false` when partly granted — the prompt is still needed.

`'appRootDir'` is the one resource that asks the server rather than checking a permission string: `app-root-dir:…` is a pseudo-permission that only resolves while a grant is being written, so a permission check on it always answers false.

### Batching

Either method takes an array, each entry naming its own resource:

```js
const [documents, apps] = await puter.perms.request([
    { resource: 'folder', name: 'Documents', access: 'write' },
    { resource: 'apps' },
]);
```

Everything already held is settled **before** anything is asked, so the dialog lists only what is actually missing — and never appears at all when the whole batch is already granted. The set goes up as one request, which is the point: the user answers once and the answer covers all of it. A denial denies every entry that needed the prompt; entries that were already held keep their value, since nothing was asked about them.

`check([...])` answers per entry in the order asked, so a caller can tell *which* parts are missing rather than only that some part is.

### How the resources are wired

Each resource declares four things in one registry entry: how to ask for it alone, whether it is held, the strings a batch pools into a prompt, and the value once held. The strings themselves are defined once in `lib/permissionStrings.js`, so a request and its matching check cannot name the same permission differently. Single-resource `request` delegates to the method that has always served that resource, so those paths behave exactly as before.

### Backward compatibility

**All 22 older methods stay callable, typed, and unchanged** — same arguments, return values and error codes. They carry `@deprecated` naming the call that replaces them:

| Deprecated | Replacement |
|---|---|
| `requestEmail()` | `request('email')` |
| `requestFolder(name, access)` | `request('folder', { name, access })` |
| `requestApps(access)` / `requestSubdomains(access)` | `request('apps' \| 'subdomains', { access })` |
| `requestAppData(app, scopes)` | `request('appData', { app, scopes })` |
| `requestAppRootDir(app, access)` | `request('appRootDir', { app, access })` |
| the 16 one-per-task names | the calls above |

They stay in the generated `.d.ts` rather than being hidden — `stripInternal` has no effect on declarations emitted from JavaScript, and dropping them by hand would break TypeScript callers the runtime still serves. Editors show them struck through.

`request('fs:/path:read')` still works: a lone string routes to the raw-permission path. No resource name contains a `:` and every permission string does, so the two forms cannot be confused.

`grantApp`, `revokeApp`, `grantOrigin`, `revokeOrigin`, `grantAppAnyUser` and `revokeAppAnyUser` are untouched and not deprecated — the consent dialog and the dashboard's uninstall path run through them.

**Eight methods are removed outright** and this is a break, signed off after review: `grantUser`, `revokeUser`, `grantGroup`, `revokeGroup`, `createGroup`, `addUsersToGroup`, `removeUsersFromGroup`, `listGroups`, plus the `req_` shim. All were undocumented, absent from the sidebar, and `grantUser` already returned 501 from the server. `puter.fs.unshare()` replaces `revokeUser` and covers the legacy case, since it falls back to live grants when no share row exists.

New exported types: `PermsResource`, `PermsBatchEntry`, `PermsAccessRequest`, `PermsFolderRequest`, `PermsAppDataRequest`, `PermsAppRootDirRequest`, `PermsPermissionRequest`, `PermsRequestDetails`. Validation is client-side and rejects with `{ message, code: 'invalid_argument' }` before any network call.

## Backend

Routes deleted: `grant-user-user`, `grant-user-group`, `revoke-user-group`, `/group/create`, `/group/add-users`, `/group/remove-users`, `/group/list`, `/group/public-groups`.

That orphaned, and so also removes:

- `PermissionService.grant/revokeUserGroupPermission` and its group-members cache bump
- the three `PermissionStore` group writers
- `GroupStore` down to `addUsers`/`removeUsers` — `getByUid` had no production caller once the routes went, which took the row decoder and `GroupRow` with it

### `revoke-user-user` is kept, deprecated

Deleting this one went too far and it has been restored. The grant side is retired and stays retired — `puter.fs.share()` is the only way in — but access those grants left behind has to remain withdrawable, and a caller reaching the endpoint over HTTP directly had no replacement. Revoking can only ever narrow what someone can reach, so keeping it carries no risk.

`PermissionService.revokeUserUserPermission` never left (it is load-bearing for `puter.fs.share()`), so this only re-wires the handler to it with the gates it always had. Nothing in this repo calls the route, which makes it exactly what a later cleanup reads as dead — so a test pins the registration and its gate alongside the restored 400 and grant/revoke round-trip cases.

The SDK's `revokeUser` wrapper is **not** restored: `puter.fs.unshare()` covers it, and re-adding the wrapper would re-grow the surface this PR exists to shrink.

**Direction this sets, per review:** user-to-user access should be built as feature-specific APIs backed by the permission service — the way sharing was, where the grant and a listable, revocable share row are written together — rather than a catch-all string give/revoke API. That is the reasoning for deleting the rest rather than keeping them around.

## Migration: three groups no code reads

`freeai`, `experimental` and `dangerous` exist in prod but in no migration — added by hand when hardcoded permissions were keyed by group name. That map is now a flat per-user floor (`default_user_permissions` in `data/hardcoded-permissions.js`), so a group nothing looks up grants nothing. Dropped for sqlite, mysql and postgres.

**Guarded, not unconditional.** `user_to_group_permissions.group_id` and `jct_user_group.group_id` are both `ON DELETE CASCADE`, so removing a group that still carries permissions or members would silently revoke them from every member. Only a group with neither is dropped, and four tests cover both branches. `system`, `admin`, `user` and `temp` are untouched — config names two, code names the others.

Consequence worth knowing at deploy time: **if a group survives the migration, the cleanup silently did not happen** for it. Check which way it went:

```sql
SELECT g.id, g.extra ->> '$.name' AS name,
       (SELECT COUNT(*) FROM user_to_group_permissions p WHERE p.group_id = g.id) AS perms,
       (SELECT COUNT(*) FROM jct_user_group j WHERE j.group_id = g.id) AS members
FROM `group` g
WHERE g.extra ->> '$.name' IN ('freeai','experimental','dangerous');
```

Non-zero `perms` or `members` means those grants need a deliberate decision before the rows can go.

## ⚠️ Permissions callout

Per AGENTS.md, this is a permission-related change. **No permission *check* logic was modified.** What deliberately stayed:

- **`PermissionService.grant/revokeUserUserPermission`** — `ACLService` and `ShareService` power `puter.fs.share()` through them. The grant route is gone and the revoke route is deprecated-but-live; the service methods are load-bearing.
- **The group permission read path** (`#scanUserGroup`, `readUserGroupPerms`, `FSService`) — kept because prod may hold group permission rows this repo can't see. Note the migration-seeded `system → admin group → driver` row is *not* the justification: `default_user_permissions` already gives every user actor bare `driver`, so that row is redundant.
- **`GroupStore.addUsers/removeUsers`** — signup, `save_account`, OIDC and the self-hosted default user all assign group membership.
- **`grant/revoke-user-app` and `grant/revoke-dev-app`** — the consent-dialog path (`UIPermissionDialog`, `UIDesktop`), and the dashboard calls `puter.perms.revokeApp(appUid, '*')` to clear grants on app uninstall.
- **`/auth/check-permissions`** — unchanged; `check()` is a new consumer of it, not a new endpoint.
- **No table is dropped and no schema is altered.** `user_to_user_permissions`, `user_to_group_permissions` and both audit tables keep their rows, so no historical grant or audit record is lost.

`check()` widens no access: it reports what the caller already holds, evaluated by the same `checkMany` path the route already used.

## Docs

The review asked that the docs wait for a nicer request surface, so they now land on it rather than on the collapsed one-per-task names:

- **`/Perms/request`** — the resource table, the details each takes, batching, and the raw-string escape hatch
- **`/Perms/check`** — new
- **`/Perms/appData`** — `requestAppData.md` re-homed rather than deleted: its scope table, private-entry guidance (`disableSharing`) and grant-lifetime notes are not signature documentation and have nowhere else to live. Inbound links from `KV/set.md` and `Objects/app.md` follow it.
- Deleted: the four thin per-method pages (`requestEmail`, `requestFolder`, `requestApps`, `requestSubdomains`) — signature plus one example, all of it now on the `request()` page
- The `Perms.md` overview is rewritten around the two methods, and the playground examples move to the new call form

`requestAppRootDir` stays undocumented, matching its status today — this PR shrinks published surface rather than growing it.

## Testing

- `npm run test:backend` — **6104 passed**, 24 skipped
- `npm run build:workerLib && npm run test:puterjs` — **1813 passed** across node, browser and workerd (was 1754 on this branch before the new API)
- `npm run check:puterjs:types` — declarations generate and type-check without `skipLibCheck`
- `npm run typecheck` — no new errors
- **30 new unit tests** in `request.test.js` (perms directory now 60) covering dispatch, per-resource returns and permission strings, batch pooling into one prompt, batch validation before any prompt is raised, legacy string routing, and every `invalid_argument` path
- **`perms.suite.ts`** exercises `request` and `check` in both forms on node/workerd, plus the batch that resolves with no prompt at all (both folders already readable, so it runs on every platform)
- `/auth/revoke-user-user`'s registration and its `requireUserActor` gate are asserted from the route decorator metadata, so a future cleanup deleting it fails loudly
- Regression-checked the sharing and app-data work (#3553, #3595, #3603, #3516, #3545) and the retained group read path — a test inserts a `user_to_group_permissions` row, asserts `check()` is true, deletes it, asserts false

**Not re-run:** the Playwright `requestPermission.spec.js` suite (33 passed earlier on this branch). It drives the GUI consent dialog through `puter.ui.requestPermission` and the `grant/revoke-user-app` routes; it does not reference `puter.perms`, and none of that changed here.

## Notes for reviewers

- **This reverses one of the eight route deletions you approved.** `revoke-user-user` came back deliberately, for the backward-compat reason above; the other seven and the 501 stub stand.
- **One test from #3545 was deleted.** `never queries group membership to resolve a user permission` spied on `GroupStore.listGroupsWithMember`, which this PR removes, so the test couldn't survive. The invariant now holds by construction: the only `jct_user_group` reference left in production code is the inline `JOIN` inside `readUserGroupPerms` — one query, no membership enumeration — and the method it guarded no longer exists to call. Happy to add a replacement guard if reviewers want belt-and-braces.
- **`@internal` does not do what AGENTS.md line 118 says.** It claims tagged members are stripped from the generated declarations. `stripInternal` has no effect on declarations emitted from JavaScript — verified with an isolated repro, and a pre-existing `@internal` method in `src/puter-js/src/index.js` also survives into `types/`. That's why the aliases here are marked `@deprecated` only. Worth a separate docs fix.
- **A batch can skip a prompt a single call would raise.** `check` is the authority on what is held, and the batch consults it before prompting; single-resource `request` keeps its existing flow unchanged (`requestFolder`, for instance, always prompts for `write`). For an app-under-user actor — the audience for these calls — the two agree, since the app holds nothing implicitly. Happy to align the single path too if you'd rather, but that changes a shipped method's flow and felt out of scope here.


Salazareo (changes requested):
backend changes look good,

but I think the docs should wait until we expose a nicer way to request the permissions:

`puter.perms.request(Enum, detailsObj)`

Salazareo (approved):
made some changes, but should be good now

PR #3615, opened 2026-08-20
docs: use absolute links to router docs in workers create page

ZHOU7821:
Replaces two relative `../router` links with absolute `/Workers/router/` links in the workers create docs. The relative form resolves to a non-existent path when the file is viewed on GitHub and diverges from the absolute-link convention used elsewhere in the docs. Confirmed `/Workers/router/` returns 200 and matches the style of neighboring links.


reynaldichernando:
thank you @ZHOU7821

PR #3616, opened 2026-08-20
feat: add MetaProvider for Muse Spark models and integrate with ChatCompletionDriver

404oops:
### What it does

Adds Meta's Model API as a **direct** chat provider (PUT-1480), serving the Muse Spark family over its OpenAI-compatible Chat Completions endpoint at `api.meta.ai/v1`. [MetaProvider.ts](src/backend/drivers/ai-chat/providers/meta/MetaProvider.ts) wraps the `openai` SDK and is registered in [ChatCompletionDriver.ts:1137](src/backend/drivers/ai-chat/ChatCompletionDriver.ts:1137) whenever `providers.meta.apiKey` is set, slotted after `alibaba` and ahead of the aggregators.

The consequential part is a **routing change**, not just a new option. Because `compareModelPreference` ranks direct vendors above resellers within a shared model bucket, registering Meta moves `meta/muse-spark-1.x` traffic off OpenRouter and onto Meta, with OpenRouter demoted to fallback. OpenRouter was previously the only route, and its account-level 18+ gate was 403ing both models for every caller — the failure in PUT-1481.

### Models

Two entries in [models.ts](src/backend/drivers/ai-chat/providers/meta/models.ts), both 1,048,576 context at $1.25/M input, $0.15/M cached, $4.25/M output, both tool-calling:

| Model | Max output | Input modalities |
|---|---|---|
| `muse-spark-1.2` | 131,072 | text, image, video, audio, pdf |
| `muse-spark-1.1` | 32,000 | text, image, video, pdf |

`muse-spark-1.2-contributor` is deliberately excluded — the same checkpoint at ~1/12th the price in exchange for Meta training on prompts and completions. It needs separate enrollment (a standard key gets `model_not_found`), and its rate would make it the cheapest route in the bucket, so listing it would quietly route callers' prompts into Meta's training set.

### Provider translation details

`max_tokens` → `max_completion_tokens`; `reasoning_effort` accepted from either `reasoning_effort` or `reasoning.effort`, with `custom.reasoning_effort` reaching the `minimal`/`xhigh` tiers the driver's own field can't express, and `none` dropped (Muse Spark always reasons and 400s on it). Two translations are load-bearing rather than cosmetic: `prompt_cache_retention` must be Meta's `in_memory`, not the driver's `in-memory`, and cached reads are subtracted from the prompt count so they aren't billed at both the input and cached rate. Params Muse Spark rejects (`stop`, `n > 1`, `logprobs`, `logit_bias`, `verbosity`) are never forwarded.

### Testing

30 unit tests plus 2 routing regression tests, and one live integration test behind a real key. The PR reports a negative control worth trusting the routing claim on: with the provider unregistered, both routing tests fail and `meta/muse-spark-1.2` resolves to `openrouter:meta/muse-spark-1.2`. Claims a full backend suite of 6,160 passed / 0 failed / 27 skipped, clean typecheck and eslint.


config.template.jsonc
```diff
@@ -346,6 +346,10 @@
             "apiURL": ""
         },
         "gemini": { "apiKey": "" },
+        // Meta Model API (Muse Spark). `muse-spark-1.2-contributor`
```
ProgrammerIn-wonderland: Why this comment?

ProgrammerIn-wonderland:
Provider looks good to merge

PR #3624, opened 2026-08-21
:wrench: PUT-1560: Deep-link the shared item from email and notifications

jfcastro92:
A share email named the file but had nowhere to go: the only link was "Open Puter", and finding what someone shared meant hunting for it under Shared. Each named file in the digest now links to itself, and a notification covering a single item points at that item.

The link carries one parameter, the masked path a recipient is already given:

    https://puter.com/?shared=%2Falice%2F<uuid>%2Freport.txt

Its second segment is the uuid, so there is no second copy of it to disagree with the path, and the GUI can still find the entry when a rename has left the name segment stale - it stats the path, then falls back to the uuid.

Built from the owner, uuid and name rather than from `ResolvedShare.path`. That path is masked for whoever made the request, and the issuer owns the entry, so it comes back as the owner's *real* path - mailing it would tell the recipient which folders the owner keeps things in, which is the one thing masking exists to prevent. A test asserts the real path never reaches the mail.

`digestLines` now returns `lead`/`items`/`trail` beside `what`, so the template can put an anchor around each name while Handlebars keeps escaping the names themselves; the URL is machine-built from the configured origin and one encoded path, so it stays literal. Concatenating the parts reproduces `what` exactly, which a test pins - the linked and sentence forms must not describe different shares.

Notifications carry the masked path rather than a URL: the recipient is already in the GUI, which opens the item in place instead of reloading. Only a single-item notification gets a target; folding into a group drops it rather than picking one of five.

In the GUI, `?shared=` joins `?download=` and `?app=` as a param that keeps the desktop booting at `/`, and the handler reuses the `/@user` public-folder flow - extracted to `open_path_target`, which carried a TODO asking for exactly this - so a file opens in its associated app and a folder in an explorer window. The param is cleared once it has been acted on, so a reload lands on the desktop rather than opening the item twice.

Invites are deliberately not linked: there is no account to route to yet, and the invite's own call to action is to create one.

Rolling-deploy safe: a digest entry queued before this has `names` and no `items`, and still flushes - without links. New entries write both, so a node on the previous build can flush them too.

## Signed out, which is the common case

A link that arrives by email is usually opened by someone with no session, and
Puter answers that by signing them into a **temporary** account. The link then
resolved against an account that cannot see the item, failed with "User or path
not found", and was cleared — so signing in afterwards did not help. The link was
spent.

A temporary session is never a share recipient, so it now prompts for login
*before* touching the link, and leaves the link in the address bar while it does.
Login reloads on success, the reload brings the link back, and the item opens as
the account that can actually see it. Dismissing the prompt clears the link
rather than looping on it.

Found by running it; see M6 below.

## Test results

Automated, against `202c43f46`:

| | |
|---|---|
| `npm run test:backend` | **6186 passed**, 25 skipped |
| GUI (`vitest run src/gui`) | **319 passed** |
| `npm run typecheck` | clean, no new errors |

New coverage: `shareDeepLink.test.ts` (8) for link building and encoding,
`resolve_shared_item.test.js` (4) for path-then-uuid resolution,
`parse_shared_path.test.js` (7) for the link shape, plus deeplink assertions in
`shareEmail.test.ts`, `EmailClient.test.ts` and `ShareNotificationService.test.ts`.

### End to end on a local instance

MailHog as the transport (`docker run -d -p 1025:1025 -p 8025:8025
mailhog/mailhog`), two accounts signed up and confirmed through their own mail,
then real shares. Captured from the actual messages:

| Case | Result |
|---|---|
| One item shared | 1 link, correct uuid, origin **with** its port |
| Three items in one call | 3 distinct links, one per uuid; subject "shared 3 items with you" |
| HTML **and** text parts | both carry the URL |
| Invite (no account) | **no links** — as intended |
| Owner's real path | absent from the message |

A representative link, exactly as it arrived:

```
http://puter.localhost:4100/?shared=%2Fdlalice%2F8b757eec-31bb-40f0-81b6-dd45cb122700%2Fdeeplink-one
```

### In a real browser (Playwright, chromium)

| | Result |
|---|---|
| **M1** link opens the item | **PASS** — explorer window titled `deeplink-one`, breadcrumb `Shared › deeplink-one`, no alert |
| **M1** param cleared afterwards | **PASS** |
| **M2** reload does not reopen | **PASS** |
| **M6** signed out → login → opens | **PASS** *(only after the fix above; it failed before)* |

Screenshots taken: desktop before, item opened, after reload, the login prompt
with the link still in the address bar, and the item open after signing in.

### Not covered

- A **file** rather than a folder, so the associated-app branch of
  `open_path_target` is unexercised. The fixtures are folders because creating a
  file needs the upload protocol.
- **Rename after sending** — the uuid fallback is unit-tested but not driven
  through a browser.
- Rendering in a **real mail client**; MailHog shows the HTML but is not one.
- The digest **rolling-deploy** fallback (`names` with no `items`) is covered by
  shape, not by an actual mixed-version deploy.

### DIFF
--- a/src/backend/clients/email/EmailClient.test.ts
+++ b/src/backend/clients/email/EmailClient.test.ts
@@ -19,6 +19,7 @@
 
 import { describe, expect, it, vi } from 'vitest';
 import type { IConfig } from '../../types';
+import { digestLines } from '../../services/share/shareNotifyTitle';
 import { EmailClient } from './EmailClient';
 
 const FROM = '"Puter" <no-reply@puter.test>';
@@ -262,13 +263,23 @@ describe('EmailClient — share notification templates', () => {
         return captured;
     };
 
+    // Built by the producer rather than hand-shaped, so a change to the digest
+    // wording can't leave these fixtures describing a shape it no longer emits.
     const HOLDER = {
         recipient: 'alice',
         subject_line: 'bob shared notes.md with you',
-        shares: [
-            { sender: 'bob', what: 'notes.md' },
-            { sender: 'carol', what: '3 items — a.txt, b.txt, +1 more' },
-        ],
+        shares: digestLines([
+            {
+                username: 'bob',
+                count: 1,
+                items: [{ name: 'notes.md', link: 'https://puter.test/?shared=%2Fbob%2Fu1%2Fnotes.md' }],
+            },
+            {
+                username: 'carol',
+                count: 3,
+                items: [{ name: 'a.txt' }, { name: 'b.txt' }],
+            },
+        ]),
         link: 'https://puter.test',
         unsubscribe_uuid: null,
     };
@@ -293,11 +304,39 @@ describe('EmailClient — share notification templates', () => {
     it('escapes item names in the html and leaves them raw in the text', async () => {
         const { html, text } = await renderShare('file_shared_with_you', {
             ...HOLDER,
-            shares: [{ sender: 'bob', what: 'r&d "notes".md' }],
+            shares: digestLines([
+                {
+                    username: 'bob',
+                    count: 1,
+                    items: [{ name: 'r&d "notes".md' }],
+                },
+            ]),
         });
 
         expect(html).toContain('r&amp;d &quot;notes&quot;.md');
         expect(text).toContain('r&d "notes".md');
+    });
+
+    // The name links to the item. The URL is ours, built from the origin and one
+    // encoded path, so it stays literal — an entity-escaped `=` would still
+    // resolve, but the text part has no parser to undo it.
+    it('links a named item to itself in both parts', async () => {
+        const link = 'https://puter.test/?shared=%2Fbob%2Fu1%2Fnotes.md';
+        const { html, text } = await renderShare('file_shared_with_you', HOLDER);
+
+        expect(html).toContain(`<a href="${link}"`);
+        expect(html).toContain(`>notes.md</a>`);
+        expect(text).toContain(link);
+        expect(html).not.toContain('&#x3D;');
+    });
+
+    // An item with no link — an invite, with no account to route to yet —
+    // renders as the plain name it always did.
+    it('leaves an unlinked item as plain text', async () => {
+        const { html } = await renderShare('file_shared_with_you', HOLDER);
+
+        expect(html).toContain('a.txt');
+        expect(html).not.toContain('>a.txt</a>');
     });
 
     it('renders as a responsive single column with no remote assets', async () => {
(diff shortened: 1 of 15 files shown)


jfcastro92:
## Browser evidence

Playwright/chromium against `202c43f46` on a local instance, with MailHog as the
mail transport. Two accounts signed up and confirmed through their own mail;
every link below came out of a real captured message.

| Case | Result |
|---|---|
| **M1** link opens the shared item | ✅ PASS |
| **M1** `?shared=` cleared afterwards | ✅ PASS |
| **M2** reload does not reopen it | ✅ PASS |
| **M6** signed out → login → item opens | ✅ PASS *(failed before the fix in this PR)* |

---

### Signed in as the recipient

**1. Baseline** — desktop before the link is opened. No item windows.

1-desktop-before

**2. M1 — the link opens the item.** Window titled `deeplink-one-mt3d128s`, and
the breadcrumb reads **`Shared › deeplink-one-mt3d128s`**: the masked path
resolved and the item is presented under Shared, not under the owner's real
path. No alert, and `?shared=` is gone from the address bar.

2-deeplink-opened

**3. M2 — a reload does not reopen it.** The link was consumed once; refreshing
lands on a plain desktop.

3-after-reload

---

### Signed out — the bug this found, and the fix

**4. Before the fix.** A visitor with no session is signed into a *temporary*
account, so the link resolved against an account that cannot see the item:
**"User or path not found."** The param was then cleared, so the link was spent —
signing in afterwards did not help. This is the common case for a link that
arrives by email.

6-signedout-landing

**5. After the fix — login first.** A temporary session is never a share
recipient, so it prompts for login before touching the link, and **leaves the
link in the address bar** while it does.

4-signedout-login

**6. M6 — the item opens after signing in.** Login reloads on success, the
reload brings the link back, and it opens as the account that can actually see
it. Same `Shared ›` breadcrumb.

5-signedout-after-login

---

### Not shown here

- A **file** rather than a folder, so the associated-app branch of
  `open_path_target` is unexercised — the fixtures are folders because creating a
  file needs the upload protocol.
- **Rename after sending**: the path-then-uuid fallback is unit-tested
  (`resolve_shared_item.test.js`) but not driven through a browser.
- Rendering in a **real mail client** — MailHog shows the HTML but is not one.

Salazareo (approved):
nit: rename the gui files to be camelCase, we need to start standardizing these

jfcastro92:
> nit: rename the gui files to be camelCase, we need to start standardizing these

Created a follow up ticket for all files inside the GUI helpers. Renaming all of them have a blast radius on all dependency imports: https://linear.app/puter/issue/PUT-1568/rename-srcguisrchelpers-files-to-camelcase

PR #3630, opened 2026-08-24
refactor: dedupe model catalog aliases and stop mutating shared catalogs

404oops:
Model catalogs are module-level constants that providers hand back by reference, but the chat/image/video drivers' `#buildModelMap` wrote normalized ids and appended `puterId` straight into them — so aliases accumulated across driver instantiations and every rebuild worked on already-mangled data. On top of that, several catalogs carried duplicate identifiers that made `.find()`-based resolution silently serve whichever entry came first. 

This PR fixes the mutation, dedupes the catalogs, and adds invariant tests so the duplicate class of bug can't come back quietly.

### Changes

- Driver non-mutation: `ChatCompletionDriver`, `ImageGenerationDriver`, and `VideoGenerationDriver` `#buildModelMap` now work on copies; the shared catalog constants are read-only. The chat driver's copy preserves the `models()` wire shape (no `aliases: []` sprouting on entries that declared none).
- `modelLookupNames()` (in `utils/modelRouting.ts`): one deduplicating id+alias flattener replacing the hand-rolled `list()` loops in ~17 providers, so a repeated or self-referential alias can no longer double-list a model.
- Catalog cleanups: self-aliases (entries listing their own id) and exact duplicate aliases removed from the claude, deepseek, groq, mistral, and openai catalogs.
- ClaudeProvider `max_tokens`: the default ceiling now comes from the already-resolved `modelUsed` entry instead of a second name-based lookup that missed dated ids (e.g. `claude-opus-4-5-20251101`) and silently capped them at 4096.
- New invariant tests (src/backend/drivers/ai-chat/providers/modelCatalogs.test.ts): every static catalog is checked for cross-entry id/puterId/alias collisions, repeated aliases, and self-aliases, with a disk-scan guard that fails if a new catalog isn't registered.

### One more thing
`gemini-3.7-flash` existed twice in GEMINI_MODELS with different values, and the first (stale) entry was the one resolution actually served. Folding them into one entry changes the live cached-token price from 8 to 7.5 (usd-cents per 1M) and the knowledge date from 2025-01 to 2026-03. The surviving values are the correct ones. This is a billing change.

### Testing

All backend driver suites pass: 1074 tests across `ai-chat`, `ai-image`, and `ai-video`.
Verified end-to-end against a locally built docker-compose deployment: all 587 registered chat models called through the live REST API with zero "model not found" routing failures, including alias and case-variant spellings.


Salazareo (approved):
This looks sane to me, just heads up to double check make sure no model name mappings go missing

404oops:
Had an agent crosscheck, nothing was lost. Think we're good.

image

PR #3641, opened 2026-08-25
🧹 PUT-1568: Rename GUI helper files to camelCase

jfcastro92:
src/gui/src/helpers was mostly snake_case, disagreeing with itself and with the camelCase file convention in AGENTS.md. Rename the 45 offenders with git mv and repoint every import specifier at the new paths.

Files only — exported symbols are unchanged. applyItemAddedToContainers.js still exports apply_item_added_to_containers and no call site is edited beyond its import line; those identifiers cross module boundaries and some hang off window, so renaming them is a separate, riskier change.


Salazareo (approved):
looks good, but just care that no dynamic import or anything like that breaks, maybe test on zenpacket first

jfcastro92:
Chased the dynamic-import concern. Results:

| Check | How | Result |
|---|---|---|
| Dynamic imports of helper paths | `git grep -E "import\([^)]*helpers/"` over `src/` | none exist |
| Computed/interpolated module paths | grep for `` import(`…${…}`) `` in `src/gui/src` | none |
| Stale path references | `git grep "<old_name>.js"` for all 45 old basenames | 0 hits |
| References outside JS | grep `*.html`, `*.json`, `src/gui/build.js`, `vi.mock()` | 0 (one prose mention of `launch_app()` in TESTING.md) |
| Case-only renames | reviewed all 45 | none — all snake→camel, so a stale import fails locally instead of only on Linux |
| Static imports resolve | `node src/gui/build.js` | builds; webpack hard-errors on unresolved imports, incl. the split chunk |
| Runtime boot, branch vs `main` | signed-in desktop in Playwright, capturing console + page errors + failed requests | **byte-identical output** on both |
| GUI unit tests | `vitest run src/gui/src` | 31 files / 346 tests pass — same count as `main`, so no renamed test dropped out of discovery |

Zenpacket tests: 
Screenshot 2026-08-26 at 10 00 00 AM
Screenshot 2026-08-26 at 10 00 08 AM

PR #3644, opened 2026-08-25
:wrench: PUT-1585: Share flag in readdir shared users in stat

jfcastro92:
## What this adds

A file's listing entry now says whether it has been shared, and `stat()` can name who it was shared with.

- **`isShared`** on `/fs/stat` and every branch of `/fs/readdir`; **`is_shared`** on their legacy counterparts and on the v1 entry shape the SDK publishes.
- **`return_shares`** on `stat` fills a `shares` array with the same objects `getShares()` returns — inherited access and unclaimed invitations included. The legacy `/stat` has accepted this flag and answered `shares: []` for as long as it has existed; it now answers for real.
- **puter.js**: `is_shared` on stat/readdir results, `returnShares` on `stat()`, both typed.

## The performance compromise

The ticket asks whether children of a shared folder should be flagged, and whether that can be fetched without hurting readdir. The answer here is **direct shares only**:

- One batched query per listing — `SELECT DISTINCT fsentry_id … IN (…)` on `idx_share_fsentry` — never a query per entry, and it runs alongside the suggested-apps and thumbnail work the listing already does rather than after it.
- Ids are chunked at 1000, because a readdir page can be 10,000 entries.
- Entries the caller doesn't own are filtered out **before** the query, so listing someone else's directory costs nothing at all.
- No ancestor walk on the read path. A file inside a shared folder is reachable without being shared itself, so it reports `false`, and `getShares()` / `stat({ returnShares: true })` remain the place inherited access shows up, via `inheritedFrom`.

## Permissions callout

- `isShared` is owner-scoped. For an entry the caller doesn't own it is `null`, not `false` — a share recipient browsing an owner's folder is never told whether other people can also reach those files.
- `shares` is gated on `manage`, the same authority `GET /share/shares` requires. Without it the array is empty rather than an error, so asking for it can't turn a `stat()` you were allowed to make into a failure.
- The flag counts shares granted by anyone holding `manage`, not only the owner's own — the same set `getShares()` reports. That is deliberate, and the docs say so.

## Reading the index required making the index honest

`listSharesOf` filters its rows against live grants, because a `share` row can outlive the grant it records. A listing can't afford that filter — it is exactly the per-entry work this design avoids — so the row itself has to be truthful.

One route left it untruthful: the deprecated `POST /auth/revoke-user-user` withdraws a grant without touching the index. Reproduced directly — `isShared: true` alongside `listSharesOf: []`, i.e. a file reporting itself as shared to nobody, permanently. The route now retires the index row with the grant (`ShareService.onGrantRevoked`), with a regression test. Two other paths I suspected turned out to be clean already: revoking a delegate, and a delegate leaving, both delete their rows today, which I verified rather than assumed.

## Why both API versions changed

`puter.fs.stat()` calls the legacy `POST /stat`, not `/fs/stat`. Only touching the v2 controller would have shipped a flag no SDK caller could see, so `LegacyFSController.stat`/`readdir` and `toLegacyEntry` carry it too. Where the two disagreed I matched v1: `shares` is always an array when requested, since that is the shape the legacy route has been returning.

`ShareController#toClientShare` moved to `controllers/share/clientShare.ts` so the FS controllers publish the identical share shape instead of growing a second copy that could drift.

## One SDK bug this surfaced

Nothing invalidated the puter.js entry cache on `share()`/`unshare()` — the socket handlers only fire on item mutations. `is_shared` rides in that cached entry, so every `consistency: 'eventual'` read kept serving the pre-share value, and the GUI's own listing refresh defaults to eventual. The badge this feature exists to drive would not have appeared until an unrelated write flushed the cache. Both operations now drop the affected `item:`/`readdir:` keys; the tests fail without the fix.

## Verification

- Backend: full suite, 6423 passing. New coverage at three levels — store (chunking, pending invites count, revocation clears the flag), service (owner-only; a file in a shared folder is *not* flagged while `canRead` confirms the recipient can reach it; the index/grant agreement above), and real HTTP for both API versions.
- puter.js: 1845 passing across node, browser and workerd, plus unit tests for the v2→v1 mapping, the stat cache rule, and share/unshare invalidation.
- `check:puterjs:types` clean; docs build clean.

## Notes for reviewers

- **`Give the puter.js API suites rate-limit headroom` is separable.** ~500 API-suite tests run through one seeded free-tier account, and adding a single `readdir` was enough to trip `fs:readdir-burst` on the browser runner. It resolves the seeded users to the unlimited policy, next to the subscription gate the harness already disables for the same reason. Test config only — no published limit moved, so `rate-limits-and-quotas.md` is untouched. The cost is that no api suite exercises free-tier windows any more. Drop that commit and the suite's readdir assertion if you'd rather keep them; the backend HTTP tests and the mapper unit test still cover the flag.
- **Known, not fixed:** `return_shares` runs the full `listSharesOf` under `FS_STAT_LIMIT` (1200/min) rather than `SHARE_LIST_LIMIT` (600/min), and the two counters are separate, so a caller can do that work faster through `stat` than through the route built for it. Consuming a second scope conditionally isn't expressible in the current gate — worth a follow-up rather than something improvised here.


src/backend/controllers/fs/FSController.ts
```diff
@@ -972,17 +973,28 @@ export class FSController extends PuterController {
...
         const wantsSize = this.#toBoolean(body.return_size);
-        const subtreeSize =
-            entry.isDir && wantsSize
-                ? await this.services.fs.getSubtreeSize(userId, entry.path)
-                : undefined;
-
-        entry.suggestedApps =
-            await this.services.suggestedApps.getSuggestedApps(entry);
+        const [subtreeSize, suggestedApps, shareFlags, shares] =
+            await Promise.all([
```
Salazareo, line 977: nice yeah, this is good to paralelize

Salazareo (approved):
looks good to me

PR #3645, opened 2026-08-25
:bug: PUT-1589: Shared file fs events seem to be lost

jfcastro92:
## The bug

A share recipient's client keeps its cache fresh from fs events pushed over their socket. `ShareService` already fans those out to holders — but only for three of them: `fs.write.file`, `fs.move.node`, `fs.remove.node`.

A new entry does not emit `fs.write.file`. It emits `fs.create.file` / `.directory` / `.shortcut` / `.symlink` — `fs.write.file` is the *overwrite* case. An in-place rename emits `fs.rename`. Neither had a listener, so a recipient watching a shared folder never learned that a file appeared in it or was renamed, while deletes, moves and overwrites worked. That asymmetry is why this reads as "events seem to be lost" rather than "sharing doesn't update".

Part of why it was never wired: `fs.create.*`, `fs.rename` and `outer.gui.item.renamed` were absent from the typed event map, so `tsc` rejects a listener for them.

## The disappearance half of the bug

Delivering the events that *were* wired turned out not to be enough either. Both mechanisms that make a file vanish were silent for anyone holding the folder above it:

- **Delete.** The audience came from the permission rows the delete removed (`#flushRetire`). A file inside a shared folder has no grant of its own — only the folder does — so the audience was empty. Deleting the *folder*, or a directly shared file, worked, which is what made this look like an unrelated quirk.
- **Move out.** The audience was resolved from the entry's **new** path. The GUI's Delete is a move to the owner's Trash, where no recipient has a share, so the lookup found nobody. `item.moved` cannot express a disappearance anyway.

Both now resolve the audience from **where the entry was**, not from the grants that went with it:

- A delete fans `item.removed` out to holders reaching the entry through an ancestor, on top of the roots `#flushRetire` already reports. A holder covered by both is told once.
- A move resolves **both ends**. Reaching both is `item.moved`, only the destination is `item.added`, only the origin is `item.removed`. Each is masked through the share that holder saw that end through, which need not be the same share at both ends.

Recipients are named by the path they knew — `/owner/<sharedRootUuid>/Docs/file.txt` — never the owner's real path or a Trash path they cannot see. That is not cosmetic: the desktop's `item.removed` handler finds the row by `data-path`, so a Trash path matches nothing and leaves the item on screen.

### The top-level share was worse

For a share on the entry *itself*, the grant is keyed on uuid, so it follows the entry into Trash. Both ends of the move resolved, and the owner's Delete reached the recipient as `item.moved` — **renaming their shared item to the GUID Trash had given it**. `shared-with-me` has always omitted trashed entries, so the listing and the event disagreed; only the event was wrong. Trashing a top-level share is now `item.removed`, and restoring one is `item.added`.

The mirror case: a move that leaves the recipient's masked address unchanged is now silent. A share masks its own root, so the owner shuffling it around their own tree is invisible to the recipient, and the event carried a `from_path` identical to `path`.

## Revoked recipients kept receiving pushes (PUT-1611)

Rolled in here because it lives in the same function this PR rewrites, and this PR widens the fan-out — deletes and moves now reach holders they previously missed, which makes the leak worse.

The audience came straight from the `share` index with no live-grant check. `/auth/revoke-user-user` deletes the permission and leaves the index row, so a revoked recipient's socket kept receiving name, size, masked path and mtime for every write and move under that folder, indefinitely.

The service already solves this elsewhere — `#reachingHolders` returns exactly the holder/entry pairs whose grant still stands, and `listSharedWithMe` was moved onto it during PUT-1412 for this reason. The realtime path never got the same treatment; it does now.

**Cost: none on the unshared path.** With no share rows reaching the entry there are no holders to check, so the write path every user takes is unchanged — pinned by a test that counts permission reads and asserts zero.

## The path half of the bug

The payload masked paths against the entry *itself* (`/owner/<file-uuid>/<name>`), and:

- puter-js handles `item.added` with a targeted eviction — `puter._cache.del('readdir:' + dirname(item.path))` (`FileSystem/index.js:149`). Against a self-masked path the parent is `/owner/<file-uuid>`, which is not a key anything cached.
- The desktop finds the window to render into by `dirpath` (`apply_item_added_to_containers.js:40`) — a field the payload didn't carry at all, so the helper returned early.

Both would have produced an event that arrives and changes nothing. Paths are now masked at the share the holder reached the entry through, and the payload carries `dirpath`, plus the `type` and `immutable` the item renderer reads.

## Behavior changes worth calling out

- Moving an item **into** a shared folder used to send that folder's holders `item.moved`. They had never seen the item, so there was nothing for the GUI to move, and `from_path` was dropped as outside their share — the event arrived and did nothing. It is now `item.added`.
- A move that does not change the recipient's masked address is now silent instead of sending `item.moved` with `from_path === path`.
- Trashing a top-level share is now `item.removed` rather than `item.moved`.

## Blast radius

**Creates.** `fs.create.*` fires on every `touch`, `mkdir`, `mkshortcut`, completed upload and batch-write item, for every user, shared or not. Measured: **+1 indexed `share` query per create** (25 creates → 25 extra reads out of 102 total; ancestor lookups are served from the entry cache).

Creates are therefore coalesced by parent folder, the same `setImmediate` buffering the delete path has used for subtree deletes. Siblings share every ancestor, so one lookup serves the whole burst. Measured on 25 concurrent creates into one folder: **25 lookups before, 1 after**.

**Deletes.** The ancestor pass is new work on every delete. Coalesced by parent the same way and reusing the existing retire batch, so a subtree costs one lookup per folder rather than one per entry — pinned by a test: 8 concurrent deletes from one folder settle in **≤2 lookups**, not 8.

**Moves.** A move now resolves the audience twice instead of once — **+1 indexed `share` query per move** where the path actually changed. A rename in place is unaffected; it goes through `fs.rename`, which still resolves once.

**Live-grant check.** Two batched, holder-keyed reads per distinct holder of a reaching share. Zero when nothing is shared, which is the overwhelming majority of writes.

## Also in here

- `from_path` on a move and `old_path` on a rename are masked the same way, and dropped when the move started outside the share. A directly shared entry masks its former path by uuid, which still resolves after the entry has moved.
- A holder with a share on both a folder and something inside it used to get one event per share; the nearest share wins now.

## Verification

Twenty-two cases in `keeping recipients in sync`, driving the real `fs.touch` / `fs.rename` / `fs.remove` paths rather than hand-emitted events, so they cover whichever event those paths actually emit. Every behavioral fix has a test that fails without it — confirmed by reverting the change and re-running each time. They pin the audience, the exact masked path, `dirpath`, `from_path`/`old_path`, the one-lookup-per-folder property on both creates and deletes, that a move inside a share is still a move, that a holder of both a file and its folder is told once, that a trashed top-level share is named as the recipient knew it, that a revoked recipient hears nothing, and that the unshared write path does zero permission reads.

Full backend suite: 6428 passing. Typecheck and lint clean.

### Against a running local Puter

Also run end to end against `npm start`, since socket delivery to a second user was the one thing the unit tests can't reach.

**Event matrix, two real socket.io clients** — one socket per user, owner mutates through the API, the recipient's socket recorded:

| Owner does | Recipient gets | |
|---|---|---|
| create a file in the shared folder | `item.added` | ✅ |
| mkdir inside the shared folder | `item.added` | ✅ |
| create a file one level deeper | `item.added`, dirpath = the subfolder | ✅ |
| rename a file | `item.renamed` + `old_path` | ✅ |
| overwrite a file (`/fs/write`) | `item.updated` | ✅ |
| move a file within the share | `item.moved` + `from_path` | ✅ |
| create **outside** the share | nothing | ✅ silent |
| 10 concurrent creates | 10 × `item.added` | ✅ all 10, one share query |
| delete the shared folder itself | `item.removed` | ✅ |
| delete a directly shared file | `item.removed` | ✅ |
| delete a file **inside** a shared folder | `item.removed` | ✅ |

Every delivered path is masked at the share and never carries the owner's real path.

**Three users on one shared folder** — A owns `/A/shared-x`, B holds it `write`, C holds it `read`, all three on live sockets. B (not the owner) deletes `doomed.txt` inside it. Run on this branch and on the same build with only `ShareService` reverted:

| | before | after |
|---|---|---|
| **B permanently deletes** | | |
| A (owner) | `item.removed` | `item.removed` |
| B (actor) | **silent** | `item.removed` |
| C (other recipient) | **silent** | `item.removed` |
| **B deletes via the GUI (move to A's Trash)** | | |
| A (owner) | `item.moved` → their Trash | `item.moved` → their Trash |
| B (actor) | **silent** | `item.removed` |
| C (other recipient) | **silent** | `item.removed` |

The owner keeps getting `item.moved` for the trash case, which is correct — the file really did move, into a Trash they can still open. Everyone else gets `item.removed`, named at the masked path under the folder *they* hold, not the owner's Trash path. That is the case where a third party's window used to keep showing a file that was gone and 404 on click.

**The shared root itself** — a folder and a file each shared directly with a recipient, owner then acts on that very item. Run on this branch and on the same build with only `ShareService` reverted:

| Owner does | before | after |
|---|---|---|
| rename the shared folder | `item.renamed` | `item.renamed` |
| move the shared folder | `item.moved`, `from_path` == `path` | silent — their address is unchanged |
| **trash the shared folder** | `item.moved` → **`/owner/<uuid>/79fc5f26-2110-…`** | `item.removed` → `/owner/<uuid>/top-yoe6p0` |
| delete the shared folder | `item.removed` | `item.removed` |
| rename the shared file | `item.renamed` | `item.renamed` |
| move the shared file | `item.moved`, `from_path` == `path` | silent |
| **trash the shared file** | `item.moved` → **`…/5998f0ad-….txt`** | `item.removed` → `…/topfile-2qnyqa.txt` |
| delete the shared file | `item.removed` | `item.removed` |

The two bolded rows are the reported bug: the recipient's shared item was renamed to a Trash GUID instead of disappearing.

**Real browser (Playwright)** — the recipient signs into the desktop and lists the shared folder, the owner then adds a file from the API, nothing is clicked in the browser:

| | `main` (pre-fix) | this branch |
|---|---|---|
| `item.added` reaches the page's socket | never arrives | arrives, masked path |
| `readdir(dir, { consistency: 'eventual' })` after | **1 item — stale** | **2 items** |

That is the chain the ticket describes — socket → `puter.fs` cache invalidation — reproduced broken on `main` and working here.

**Caveat worth recording:** `item.added` evicts one specific cache key, so it lands only when the recipient's cached key is the masked path — i.e. when they navigated the normal way. A first run that read the owner's *real* path missed the eviction. Not an issue for the GUI; an API consumer holding a real path won't be invalidated by `item.added`. The other events flush wholesale and are immune.

## Known, not fixed

A recipient with write access is notified of their own create, and the actor of a delete is now told about their own delete. The fs events carry no acting socket, so the client's "don't update if this is the original client" guard can't fire; removing an item already gone is a no-op, so this is cosmetic. Pre-existing for updates, moves and deletes. Fixing it means threading the actor through the fs event payloads, which is a bigger change than this bug warrants.

A non-owner cannot delete a file shared **directly** with them at all — delete authority comes from the parent folder, which is the owner's, so it 403s. Unchanged here, and worth a separate look if that is not the intended permission model.


src/backend/services/share/ShareService.ts
```diff
@@ -337,13 +383,37 @@ export class ShareService extends PuterService {
...
             if (!entry?.uuid) return;
+            return this.#fanOutToHolders(entry, 'outer.gui.item.updated').catch(
+                () => {
+                    // Same — never fail a write over its notification.
+                },
+            );
+        });
+
+        // A create is `fs.create.<flavor>`, not `fs.write.file`.
+        this.clients.event.on('fs.create.*', (_key, data) => {
```
Salazareo: please make sure there's no outside event - like from other nodes - sending this, could cause a loop, but should be blocked by broadcast service so I think its ok

Salazareo (approved):
lgtm

src/backend/services/share/ShareService.ts
jfcastro92: Confirmed, it's blocked. Replication is opt-in by prefix: webhooks carry `outer.*`, redis carries `pubsub.*`. The triggers here are `fs.create.* / fs.rename / fs.write.file / fs.move.node / fs.remove.node`, so they never leave the node — only the node that did the write fans out. 

What we emit (`outer.gui.item.*`) does replicate, but the consumer on the far side is `SocketService`, and `ShareService` only listens to `fs.*`, so it can't re-enter the fan-out. No loop, no duplicates. Socket rooms are cluster-wide anyway, so one node's push already reaches recipients on any node.

Added a guard regardless, since that safety lives in another service: the handlers now skip anything tagged `from_outside`, with a test. Separately and pre-existing — inbound webhook events are re-emitted under whatever key the sender supplies, with no allowlist, so a peer could inject `fs.remove.node` (which retires grants).

Salazareo (approved):

PR #3646, opened 2026-08-26
:wrench: PUT-1590: Mark shared items in the file listings

jfcastro92:
Stacked on #3644 (PUT-1585), which is where `is_shared` comes from. Base branch is that one, not `main`.

## What this does

A shared file looked exactly like a regular one. The data to tell them apart arrived with the readdir/stat share flag; nothing rendered it.

A blue dot on the item's icon, lower-right, in both listings:

| File | Change |
|---|---|
| `UI/UIItem.js` | marker inside `.item-icon` — desktop, explorer windows, file dialogs |
| `UI/Dashboard/TabFiles.js` | same marker in the dashboard's own row markup |
| `helpers/sharedBadge.js` *(new)* | `has_direct_share(shares)` + `mark_item_shared(path, on)` |
| `UI/UIWindowShare.js`, `UI/Dashboard/UIShareModal.js` | hook the marker into each dialog's single `render()`, so a grant or revoke updates it without waiting for a re-listing |
| `helpers/refresh_item_container.js`, `helpers/apply_item_added_to_containers.js` | pass the flag through when building items |
| `css/style.css`, `css/dashboard.css` | 9px on the desktop's 45px icons, 7px on the dashboard's 24px rows |

The string was already in the tree and unused — `item_shared_by_you` ("You have shared this item with at least one other user"), present in 40 locales. No new i18n key.

## Design notes

- **On the icon, not in the badge cluster.** `.item-badges` is pinned to the row's top-left in the dashboard, not to the icon, so a badge there could never sit on the file. Anchoring to `.item-icon` keeps the marker on the corner at both icon sizes.
- **A dot, not a glyph.** At 7–9px the two-person icon is an illegible smudge; colour is the signal the ticket asked for. The white ring is what keeps it visible against both a light row and a dark icon.
- **Inherited access is not marked.** It is a state of the folder that was shared, so marking every file inside would repeat one fact across hundreds of rows — the backend flag is direct-only for the same reason.

## Validated in a running local Puter

Playwright against `npm start`: one shared and one plain item of each kind, badge state read from the rendered DOM. Nothing clicked between the share and the read.

| Item | `data-is_shared` | Marker |
|---|---|---|
| shared file | `1` | visible |
| plain file | `0` | hidden |
| shared folder | `1` | visible |
| plain folder | `0` | hidden |
| system folders (Desktop, Documents, …) | `0` | hidden |

Captured in both surfaces — the dashboard's Files list and the classic desktop's icon view. GUI suite: 351 passing, including 5 new cases for `has_direct_share` (direct, inherited-only, mixed, unclaimed invite, empty).

## Worth knowing before testing this on a deployment

The GUI loads puter.js from the CDN (`https://js.puter.com/v2/`), not from a local build, so **the marker stays hidden until #3644's SDK change is published** — `is_shared` arrives as `null` until then. The backend wire is already correct (`/fs/readdir` returns `isShared: true`); it is purely the SDK in the page being the published one. The test routes that URL to `src/puter-js/dist/puter.dev.js`, the way the repo's own browser runner does.

## Not verified

The live update when sharing from a dialog — the marker appearing without a re-listing. It is a one-liner in each dialog's single `render()` funnel and `has_direct_share` is unit-tested, but I did not drive the dialog through the UI.


jfcastro92:
## Validated in a running local Puter

Playwright against `npm start`: created one shared + one plain item of each kind, then read the rendered badge state. Nothing was clicked in the browser between the share and the read.

| Item | `data-is_shared` | Badge |
|---|---|---|
| shared file | `1` | visible |
| plain file | `0` | hidden |
| shared folder | `1` | visible |
| plain folder | `0` | hidden |
| system folders (Desktop, Documents, …) | `0` | hidden |

Tooltip resolves to "You have shared this item with at least one other user." (`item_shared_by_you`, already present in 40 locales).

## Screenshots 

put-1590-icon-crop

put-1590-list-crop

put-1590-list-view




GUI suite: 351 passing, including 5 new cases for `has_direct_share` — direct share, inherited-only, mixed, unclaimed invite, and empty.

## Not verified

The live update path — sharing from the dialog and watching the badge appear without a re-listing. It is a one-liner in each dialog's single `render()` funnel and `has_direct_share` is unit-tested, but I did not drive the dialog through the UI.

## Deliberate omission

Inherited access is not badged. It is a state of the folder that was shared, so marking every file inside would repeat one fact across hundreds of rows — the backend flag is direct-only for the same reason.

Salazareo (approved):

PR #3654, opened 2026-08-27
:wrench: PUT-1619: Publishing a directory takes ownership, not write

jfcastro92:
Closes PUT-1619. Found via internal source audit.

Creating a hosted subdomain gated `root_dir` on `write` — `checkFSAccess` defaults to it — and hosting serves everything under that directory with the ACL deliberately bypassed. So a recipient of a `write` share could point a `*.puter.site` subdomain at the owner's folder and make that subtree world-readable. The recipient could already read those files, so this is not a read escalation; what was new is that it is continuous, covers whatever the owner writes there later, and the owner gets no signal — the subdomain row belongs to the recipient, so nothing the owner can list would show it. `update` had the same gate for a changed `root_dir`.

## The fix

`#checkPublishAccess` now decides both call sites:

- **The actor's own tree** → `write`, exactly as before.
- **Anyone else's** → `manage` required. That is the "Can edit & share" level, which is the owner delegating the decision.

Keyed on who owns the entry rather than requiring `manage` outright, which is what the ticket proposed. `manage`'s `is-owner` implicator declines to answer for app actors ([FSService.ts:238](https://github.com/HeyPuter/puter/blob/main/src/backend/services/fs/FSService.ts#L238)), so a flat `mode: 'manage'` would refuse every app publishing a directory its user handed it — with no way for an app to obtain that grant through the SDK. Apps publishing their own AppData are unaffected either way, via the ACL short-circuit.

The `write` check still runs first: it is what masks a directory the caller cannot see at all as a 404 rather than a 403. `manage` satisfies every lower mode, so a manage-holder is not tripped by the ordering.

## Tests

The two exposure tests **fail without the driver change** — verified by stashing it, they create the subdomain instead of throwing 403:

- a write-share recipient is refused on `create`;
- and refused when repointing a subdomain they own at the shared directory.

The other two are compatibility guards that pass either way: a `manage` recipient is accepted, and so is the actor's own directory.

Full backend suite: 6477 passed, 0 failed. Typecheck clean.

## Also changed

- **Docs** — `hosting.create()` states the ownership rule; `share()`'s level list now says `write` does not allow publishing a directory and `manage` does. Nothing was added to "What sharing does not promise" in `FS/share.md`, since this is fixed rather than accepted.
- **GUI** — the item context menu's Publish As Website reuses the own-it-or-`manage` answer it already computes for sharing, so the option is not offered where the backend would now refuse.

## Deliberately left alone

- **The window head context menu** still offers Publish for someone else's open folder. Its handler is synchronous and returns `true` to preserve the native context menu, so it cannot await the share-mode lookup, and the sync alternative (`is_owned_by_me`) would wrongly disable it for `manage` delegates. Clicking it now answers "Publishing this directory is up to whoever owns it", which beats a wrong disable.
- **`WorkerDriver`** also writes `root_dir_id`, but from the read-gated worker *source file*, and `__workers/` paths are carved out of public serving — so it does not produce the directory exposure this fixes.
- **The owner still cannot see a subdomain rooted in their own tree.** After this, only a `manage` delegate can create one, but the row lives under the delegate's account. That is the visibility half of the ticket and a separate change — worth its own issue if we want it.

### DIFF
--- a/src/backend/drivers/subdomain/SubdomainDriver.ts
+++ b/src/backend/drivers/subdomain/SubdomainDriver.ts
@@ -30,6 +30,7 @@ import type { Actor } from '../../core/actor.js';
 import type { DriverConcurrentConfig, DriverRateLimitConfig } from '../meta.js';
 import type { FSEntry } from '../../stores/fs/FSEntry.js';
 import type { UserRow } from '../../stores/user/UserStore.js';
+import { MANAGE_PERM_PREFIX } from '../../services/permission/consts.js';
 import { expandTildePath } from '../../services/fs/resolveNode.js';
 import { isUniqueViolation } from '../../util/dbError.js';
 import { buildHostedSubdomainIndexUrlCandidates } from '../../util/hostedAppBacking.js';
@@ -176,7 +177,7 @@ export class SubdomainDriver extends PuterDriver {
                 legacyCode: 'bad_request',
             });
         }
-        await this.services.fs.checkFSAccess(entry, actor);
+        await this.#checkPublishAccess(entry, actor);
 
         // A name some other user's app still points at is not free either.
         // Deleting a hosted subdomain leaves the app row's `index_url` intact,
@@ -367,7 +368,7 @@ export class SubdomainDriver extends PuterDriver {
                 });
             }
             if (rootDirId !== (row.root_dir_id ?? null)) {
-                await this.services.fs.checkFSAccess(entry, actor);
+                await this.#checkPublishAccess(entry, actor);
             }
             patch.root_dir_id = rootDirId;
         }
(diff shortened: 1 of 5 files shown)


src/backend/drivers/subdomain/SubdomainDriver.ts
```diff
@@ -574,6 +575,39 @@ export class SubdomainDriver extends PuterDriver {
...
+     * list would show it. `manage` — "Can edit & share" — is the grant that
+     * delegates the decision.
+     */
+    async #checkPublishAccess(entry: FSEntry, actor: Actor): Promise<void> {
+        // First, because this is the check that masks a directory the caller
+        // cannot see at all as a 404.
+        await this.services.fs.checkFSAccess(entry, actor);
+        if (entry.userId === actor.user?.id) return;
+        try {
+            await this.services.fs.checkFSAccess(
```
Salazareo, line 595: im not 100% if we should let people who have manage access deploy actually...

technically its sharing, but I think its a bit odd what do you think @ProgrammerIn-wonderland

Salazareo (commented):
want @ProgrammerIn-wonderland opinion here. but code itself looks good

Salazareo:
gonna merge, well change as needed

PR #3655, opened 2026-08-27
Comprehensive Normalization.

404oops:
# Comprehensive Normalization.

This PR makes Puter's chat API return one response shape — the OpenAI `choices[0]` format — across all vendors, on an opt-in/opt-out basis. Callers get a new tri-state `normalize` option (per call and SDK-wide via `puter.ai.normalize`); models released on or after **2026-09-01** are normalized by default, while older models keep their vendor-native shape unless the caller asks otherwise. Normalization is lossless for reasoning models: Anthropic thinking-block signatures and OpenAI Responses reasoning items are preserved on a new `message.reasoning_details` field and can be replayed verbatim to continue an extended-thinking tool-use turn. Alongside the gated work, a small set of *unconditional* reasoning-field equalizations ships for existing models (documented in `chat.md`), and the provider catalogs get ~53 new model entries and ~23 new aliases. A 23-provider consistency test matrix pins the equalized contract.

### Response normalization core
- **`src/backend/drivers/ai-chat/utils/normalizeToOpenAI.ts` (new)** — the whole policy in one module: `OPENAI_SHAPE_CUTOFF = '2026-09-01'`, `isPostCutoffRelease` (timestamp compare, tolerant of month-precision dates, missing/unparseable ⇒ pre-cutoff), `needsOpenAICoercion` (Anthropic envelope or content-block array), `shouldPresentAsOpenAI` (the single precedence rule), and `normalizeResultToOpenAI` — an idempotent coercer that joins text blocks into string `content`, converts `tool_use` → `tool_calls` (arguments stringified), joins `thinking` into `reasoning` (blank-line separated), preserves `thinking`/`redacted_thinking` verbatim in `reasoning_details`, maps Anthropic `stop_reason` → OpenAI `finish_reason`, and passes unmapped vendor reasons (e.g. `pause_turn`) through verbatim. Already-OpenAI-shaped results return **by reference**.
- **`ChatCompletionDriver.ts`** — applies the coercer to non-streaming message results per `shouldPresentAsOpenAI(args, model.release_date)`, where `model` is the model that actually **served** the request (fallback/reroute included). Stamps `normalized: true` only on OpenAI-shaped output. ⚠️ One unconditional change here: the legacy `response.normalize` path (Anthropic-block normalization) **no longer stamps `normalized: true`** — the field now exclusively means "OpenAI shape".
- **`PuterAIController.ts`** — all four wire-compat routes (`/openai/chat/completions`, completions, responses, `/anthropic/messages`) pin `normalize: false`, so the release-date cutoff can never change what those routes' own translators receive.
- **`types.ts`** — `ICompleteArguments.normalize?: boolean` documented as the tri-state.

### Per-provider equalization
- **`MistralAiProvider.ts`** — (1) **Gated on the policy**: non-streaming remap of the SDK's camelCase dialect — `finishReason` → mapped `finish_reason` (unmapped values pass through), `toolCalls` → OpenAI `tool_calls` with stringified arguments, and magistral chunked `content` arrays split into string `content` + `reasoning`. Without the flag / pre-cutoff, callers keep seeing the SDK-native keys. (2) **Unconditional**: the *streaming* chunk splitter — chunked `delta.content` is split into text and `reasoning` deltas regardless of policy (previously the raw array reached `addText` and stream consumers got stringified objects; streamed chunk types must not depend on a response-format flag). (3) New opt-in `custom.prompt_mode` → SDK `promptMode` passthrough.
- **`OpenAIUtil.js` `handle_completion_output`** — `normalizeReasoningContent(ret)` now runs for **every** chat-completions provider, **unconditionally**: `reasoning_content` (DeepSeek wire convention) is renamed to `reasoning` and the vendor key deleted on all non-streaming responses. This removes a field from existing models and is *not* behind the normalize policy — the one field removal, called out in the docs.
- **`OpenAIUtil.js` `handle_completion_output_responses_api`** — three **unconditional** changes to existing OpenAI/Azure Responses-API models: `finish_reason` is now `'tool_calls'` on tool turns (was hardcoded `'stop'`); `reasoning: null` is replaced by string-or-absent (populated from reasoning summaries); `reasoning_details` (item `id`, `encrypted_content`, `summary`) is added for replay.
- **Streaming (Responses API)** — `create_chat_stream_handler_responses_api` now routes `response.reasoning_summary_text.delta` events to the `reasoning` channel (blank line between summary parts). Unconditional and additive: those deltas were previously dropped.

### Reasoning round-trip / replay
- **`ClaudeProvider.ts`** — inbound messages carrying round-tripped artifacts are handled: `thinking`/`redacted_thinking` blocks from `reasoning_details` are spliced back **ahead of** the content array (Anthropic requires leading thinking blocks with intact signatures), and output-only fields (`reasoning`, `refusal`, `reasoning_details`) are stripped — on a **copy**, because the driver reuses one `messages` array across fallback attempts. The copy-on-write invariant is pinned by tests at both the provider level (frozen objects, double-send) and the driver level (fallback attempts share the array by reference).
- **`OpenAIUtil.js` `process_input_messages_responses_api`** — the mirror for the Responses API: `reasoning_details` items expand into standalone `reasoning` input items preceding their message, output-only fields (`reasoning`, `refusal`, `normalized`) stripped via destructuring rebind (caller objects untouched).

### SDK surface (puter-js)
- **`modules/ai/index.js`** — `puter.ai.normalize = true` default. Note the asymmetry: SDK-wide `true` means "the release-date policy" (nothing sent on the wire), while per-call `true` means "force OpenAI shape". Only SDK-wide `false` rides the wire (`normalize: false` on every call). Per-call always wins.
- **`modules/ai/chat.js`** — the forwarding logic above; **`modules/ai/types.js`** — JSDoc for `normalize`, `finish_reason`, `normalized`, `reasoning`, `reasoning_details`, `refusal`.

### Catalog backfill (billing-bearing — all unconditional)
- **Alibaba (`alibaba/models.ts`, +~1,000 lines)**: 46 new entries — Qwen Plus/Max/Flash dated snapshots, Qwen3 2507 open-weight refreshes, coder, VL, Omni, MT translation, plus **8 third-party models hosted on Model Studio** (GLM-5.1/5.2/5.2-fast-preview, DeepSeek V3.2/V4 Flash/Pro + dated variants, Kimi K2.7 Code) at Alibaba's own rates. 4 dated-snapshot aliases added to existing entries.
- **OpenAI**: 12 dated-snapshot aliases on existing entries + 2 new entries (`chat-latest`, `gpt-4o-2024-11-20`).
- **Gemini**: rolling `gemini-{flash,flash-lite,pro}-latest` aliases (×2 spellings each) pinned to the newest catalog entry, + 2 free Gemma 4 entries.
- **DeepSeek**: `deepseek-v4-flash-vision-exp`. **Mistral**: `zai-glm-5-2` (Mistral-hosted) + a `mistral-medium-3.5` spelling alias. **Z.AI**: `glm-5.3-flash` at list price (50% promo deliberately not encoded).

Backfill rules enforced: nothing vendor-deprecated (15 already-written OpenAI entries were culled against the deprecations page), nothing without a price confirmed on the vendor's official pricing page, nothing absent from the vendor's live `/models` listing, and nothing that fails a live routing probe (one entry removed on that rule).

### Docs
- **`chat.md`** — a full "Response Normalization" section: the flag, the cutoff, the SDK-wide switch, the served-model caveat, the OpenRouter live-release-date caveat, streaming unaffected, replay guidance, and a before/after table for the four unconditional reasoning-field changes. **`chatresponse.md`** — `content` string-or-array, `finish_reason` mapping table (open set), `normalized`, `reasoning`, `reasoning_details`, `usage`. The Claude cache-control example (docs + playground) switched to `normalize: true` + string content.

### Tests
New: `providerConsistency.test.ts` (23 providers × text/tool/reasoning against mocked upstreams — the equalized-contract matrix), `normalizeToOpenAI.test.ts` (21 cases), plus additions to the driver (12, including full precedence coverage and the fallback shared-array invariant), Claude (4), Mistral (6), OpenAIUtil (9), puter-js unit (7), and one live API-suite test. Controller tests grew `normalize: false` assertions.

### Type/lint cleanup
Annotation-only changes across ~12 files (casts, callback params, return types, `context.ts` typing the pre-existing `driverName` key). Two slightly more than formatting: `ICompleteArguments.text`/`verbosity` retyped from `'concise' | 'detailed'` to `'low' | 'medium' | 'high'` (matches what vendors accept; compile-time only), and the xAI STT `Blob` construction now copies the buffer into a `Uint8Array` (type fix; runtime copy, bytes identical).

## Behavior contract after this PR

Precedence for non-streaming results, resolved in `shouldPresentAsOpenAI` (used identically by driver and Mistral):
1. Per-call `normalize: true` ⇒ OpenAI shape, any model. `normalize: false` ⇒ provider-native, any model.
2. Else, legacy `response.normalize` (internal) ⇒ Anthropic-block normalization, suppressing the OpenAI presentation (and no longer stamping `normalized`).
3. Else, release-date cutoff: served model's `release_date >= 2026-09-01` ⇒ OpenAI shape; earlier, missing, or unparseable ⇒ native. The **served** model decides — a fallback or content-block reroute to a date-less model yields native shape even if the requested model was post-cutoff.

SDK: `puter.ai.normalize` defaults to `true` = "the policy" (nothing on the wire); only `false` is transmitted; per-call overrides both ways.

Normalized responses: string-or-null `message.content`, `message.tool_calls`, `refusal: null`, optional `reasoning` / `reasoning_details`, mapped `finish_reason` (`end_turn`/`stop_sequence`→`stop`, `max_tokens`→`length`, `tool_use`→`tool_calls`, `refusal`→`content_filter`, unmapped values verbatim — treat as an open set), `normalized: true`, `usage` untouched (key names remain provider-specific by design). Native responses are byte-for-byte what the provider path returned.

Streaming is **never** touched by `normalize` (the driver returns stream results before the coercion branch, test-pinned); the two unconditional streaming changes (Mistral chunk split, Responses reasoning-summary routing) exist precisely so chunk types don't depend on the flag.

## Provider spec audit

The precondition for this work was to establish which providers already speak
the OpenAI format and which are out of spec. Result: 20 of 23 needed no change
at all, because they funnel through `OpenAIUtil.handle_completion_output` and
already return `choices[0]`-shaped results.

The table has 21 rows for the 23 entries the conformance matrix drives: openai and
azure each register two providers, one per wire API.

| Provider | Wire API | Endpoint | Verdict |
| --- | --- | --- | --- |
| openai | Chat Completions + Responses | SDK default (`api.openai.com/v1`) | native OpenAI |
| azure | Chat Completions + Responses | per-deployment `config.apiURL` | native OpenAI |
| **meta** | Chat Completions | `https://api.meta.ai/v1` | **vendor's own OpenAI-compatible layer — conforms, no change needed** |
| gemini | Chat Completions | `https://generativelanguage.googleapis.com/v1beta/openai/` | Google's OpenAI compat endpoint — conforms |
| alibaba | Chat Completions | `https://dashscope-intl.aliyuncs.com/compatible-mode/v1` | conforms |
| byteplus | Chat Completions | `https://ark.ap-southeast.bytepluses.com/api/v3` | conforms |
| deepseek | Chat Completions | `https://api.deepseek.com` | conforms; `reasoning_content` renamed centrally |
| groq | Chat Completions | `groq-sdk` default (`api.groq.com/openai/v1`) | conforms |
| hoonify | Chat Completions | `https://api.hoonify.ai/v1` | conforms |
| infron | Chat Completions | `https://llm.onerouter.pro/v1` | conforms |
| minimax | Chat Completions | `https://api.minimax.io/v1` | conforms |
| moonshot | Chat Completions | `https://api.moonshot.ai/v1` | conforms |
| neuralwatt | Chat Completions | `https://api.neuralwatt.com/v1` | conforms |
| ollama | Chat Completions | `http://localhost:11434/v1` (configurable) | OpenAI compat endpoint — conforms |
| openrouter | Chat Completions | `https://openrouter.ai/api/v1` | conforms; note its `release_date` is live-derived (see above) |
| together | Chat Completions | `together-ai` SDK default (`api.together.xyz/v1`) | conforms |
| xai | Chat Completions | `https://api.x.ai/v1` | conforms |
| zai | Chat Completions | `https://api.z.ai/api/paas/v4` | conforms |
| **mistral** | Mistral SDK | `@mistralai/mistralai` default (`api.mistral.ai/v1`) | **out of spec: camelCase dialect (remapped behind the policy gate) + chunked reasoning content (split into text/reasoning chunks, ungated)** |
| **claude** | Anthropic Messages | `@anthropic-ai/sdk` default (`api.anthropic.com/v1`) | **out of spec: content blocks + `stop_reason` — the coercer's whole reason for existing** |
| fake-chat | internal fixture | none | Anthropic-shaped by design (drives the e2e normalize test) |

Meta was the suspected edge case going in. It is not one: Meta
ships its own OpenAI-compatible layer at `api.meta.ai/v1` and the provider
drives it through the standard OpenAI SDK, which is why Meta appears nowhere in
this diff. Anthropic's own OpenAI-compat beta endpoint was evaluated and
rejected for Claude — routing through it would cost prompt-caching control,
compaction, and the cache-token usage detail metering bills on. Hence a central
coercer rather than per-provider endpoint swaps.

## Risk notes

- **Unconditional changes to existing models** (the review-critical list): (1) `reasoning_content` → `reasoning` rename on all non-streaming chat-completions responses — a field removal; (2) Responses-API `finish_reason` now `tool_calls` on tool turns; (3) Responses-API `reasoning` null → absent-or-string (`'reasoning' in msg` changes); (4) Responses-API `reasoning_details` added; (5) Mistral streaming chunk split; (6) Responses streaming reasoning-summary deltas now emitted; (7) legacy `response.normalize` path no longer sets `normalized: true`. All are documented and bug-fix-shaped, and each was explicitly approved during review rather than shipped silently — but none is behind the policy, so sign off on them consciously.
- **Cross-provider id collisions in the catalog** (verified against `compareModelPreference`): the Alibaba-hosted third-party entries reuse the direct vendors' ids, so they merge into shared routing buckets — which is this repo's deliberate vendor+reseller fallback design. Concretely: `deepseek-v4-flash`/`-pro` and `kimi-k2.7-code` keep the direct vendor primary on the cost tiebreak (Alibaba becomes a more expensive fallback route); `glm-5.2` ties Z.AI **exactly** (same price, same id length), so the serving provider falls to provider-registration order. Billing-neutral either way — both routes serve the same upstream model at the same price — but reviewers should confirm the fallback-bucket effect is wanted.
- **Pricing-bearing entries** (~53): tiered Alibaba models are encoded at the **base (≤32K-input) tier** — large-context calls under-bill; DeepSeek time-of-day models are pinned at the **busy** rate (off-peak over-bills, deliberate); Z.AI `glm-5.3-flash` ignores the current 50% promo (over-bills, deliberate); qwen3-30b-a3b thinking-mode output rate not modeled (under-bills); Gemma 4 entries are $0 per Google's pricing page.
- **Alias/date oddities to double-check against vendor docs**: `gpt-5.4-nano` release `2026-03-19` vs alias `...-2026-03-17` (pre-existing quirk); `qwen3.6-flash` release `2026-04-27` vs alias `...-2026-04-16`; `qwen3.7-plus` release `2026-06-02` vs alias `...-2026-05-26` (snapshot-name-vs-GA-date is plausible, but worth confirming). New OpenAI entry with bare id `chat-latest` — verified live: the upstream API accepts exactly that id.
- **Rolling aliases pinned to fixed entries**: `gemini-*-latest` and `qwen-plus-latest` are hot-swapped upstream; Puter's pinning goes stale (and mis-prices) the day the vendor moves them — a maintenance obligation, acknowledged in comments.
- **OpenRouter caveat (documented, not new code)**: OpenRouter release dates come from its live listing, so a model newly listed there on/after the cutoff flips to normalized-by-default without any Puter change.
- **Docs asymmetry**: SDK-wide `normalize = true` means "policy", per-call `true` means "force" — correct and tested, but easy for users to misread.
- **Vendor-deprecation flags left for maintainers**: pre-existing `o1`, `o3-mini`, `o4-mini` entries are now vendor-deprecated (shutdown 2026-10-23) — untouched here since they predate this work; the pre-existing `deepseek-v4-flash`/`-pro` prices no longer match DeepSeek's current pricing page (billing drift found during review, not introduced or changed here); gemma-4 emits its own `<thought>` markup inline in content (model behavior).

## Known gaps

- **`reasoning_details` is not portable across vendors.** The artifacts are
  provider-specific (an Anthropic signature means nothing to OpenAI), so replay
  only works against the model that produced them. Documented, not enforced — a
  fallback reroute mid-conversation cannot replay a reasoning turn.
- **Unrelated fix carried along:** the xAI speech-to-text provider now copies
  each upload buffer into a `Uint8Array` before wrapping it in a `Blob`. Node's
  `Buffer` does not satisfy the DOM `BlobPart` signature, so the old code was a
  type error and a latent bug; the fix costs one extra in-memory copy per audio
  upload. Not part of the normalization work — flagged so reviewers aren't
  surprised by a non-chat driver in the diff.
- **`ChatProvider` no longer stubs `checkModeration`.** A no-op stub added
  during the type cleanup has been removed: it had no subclasses and no
  production callers, and it was suppressing a genuine design smell — the base
  class does not implement `IChatProvider`. That pre-existing `TS2420` is back
  and is covered by `origin/main`'s baseline entry, so the gate passes.
- **`tools/typecheck-baseline.json` is untouched by this branch.** An earlier
  commit had removed 31 entries (49 error instances) as a side effect of fixing
  the AI-code type errors they suppressed; that removal has been reverted. The
  errors stay fixed in the source, so `npm run typecheck` prints "48 baselined
  error(s) fixed" and exits 0. Regenerating the ledger is the maintainer's call.
- **A reviewer decision was left open, not settled.** `renameReasoningContent`
  deletes `message.reasoning_content` even when `message.reasoning` was already
  present, so a provider sending both keys loses the vendor one. That looks like
  data loss, and changing it was requested during review — but
  `BytePlusProvider.test.ts` and `ZAIProvider.test.ts`, both pre-existing on
  `main`, pin the current behavior deliberately, with a fixture value named
  `'should-be-dropped'`. Honouring the request would mean rewriting two tests
  this branch does not own and changing BytePlus/ZAI behavior nobody asked to
  change, so it was left alone. A reviewer should decide: keep the drop as those
  tests specify, or change it and update them.
- **The copy-on-write guarantee is narrower than it sounds.** Both reasoning-replay
  input paths now strip output-only fields from a copy, so a replayed message's
  thinking signature survives a fallback retry. But the driver still normalizes
  every inbound message in place before any provider runs
  (`normalize_single_message` in `utils/Messages.js`, pre-existing), rewriting
  string `content` into `[{type:'text'}]` blocks on the caller's own objects. The
  fix protects the reasoning artifacts, not whole-message immutability, and the
  new driver-level test says so explicitly.
- **Beyond the original request:** the shipped Claude cache-control playground
  example (`src/docs/src/playground/examples/ai-claude-cache-control.html`) was
  switched to `normalize: true` with string `content`. Nobody asked for it; it
  was changed because its `content[0].text` reading goes stale the moment
  post-cutoff Claude models normalize by default. Flagged so the diff's
  user-facing surface is not a surprise.
- **Doc claims corrected late in review**, listed because they were wrong in an
  earlier push of this branch: `content` is string-**or-null** on normalized
  responses (tool-only turns have no text), `reasoning_details` is not scoped to
  normalized responses, `finish_reason` is an open set, and the release-date
  trigger depends on the serving provider's own dates.
- Cosmetic: repo-wide ESLint reports 37 errors / 61 warnings across 20 files (5 of them stale copies under `.claude/worktrees/`), concentrated in `src/backend/stores`. **Zero** land in any file this branch touches; every file in this diff is clean.

## Test results

All runs on the final rebased tree (14 commits on top of `origin/main`), fresh builds first, exit codes read directly — no output filtering.

### Automated suites

| Suite | Result | Exit |
| --- | --- | --- |
| `src/puter-js` build (webpack) | compiled | 0 |
| `npm run build:workerLib` | compiled | 0 |
| Backend (`npm run test:backend`, 274 files) | **6,598 passed**, 26 skipped | 0 |
| puter.js API suite (`npm run test:puterjs` — same tests on node, browser/Playwright, and workerd/miniflare, against the built bundle) | **1,889 passed**, 146 skipped | 0 |
| GUI (`vitest --config src/gui/vitest.config.js`) | **418 passed** | 0 |
| Typecheck gate (`npm run typecheck`, read-only) | no new errors (35 known, baselined; `tools/typecheck-baseline.json` byte-identical to `main`) | 0 |
| puter.js declarations (`npm run check:puterjs:types`) | generate + type-check cleanly | 0 |
| ESLint over `src/backend/drivers/ai-chat` | 0 errors (13 pre-existing `any` warnings) | 0 |

New coverage added by this branch: ~90 tests — the 23-provider × text/tool/reasoning conformance matrix (`providerConsistency.test.ts`, driven with `normalize: true` so the equalized contract is what's asserted), 25 coercer unit tests, 13 driver precedence/cutoff tests, reasoning round-trip tests on both input paths (thinking-signature preservation sabotage-verified: the test fails if the copy-on-write is reverted), a driver-level two-attempt fallback test that exercises the real fallback loop, Mistral chunk-splitting tests (both paths, both `normalize` values asserted identical for streams), Responses-API reasoning tests, SDK flag-precedence tests, and an e2e `normalize: true` case that runs in all three API-suite runners.

### Live end-to-end verification (local Docker deployment, real provider APIs)

The branch was built into the self-hosted Docker image and exercised over HTTP through Caddy against real vendor keys:

**Normalization matrix — 25/25 meaningful checks passed:**
- Claude: native blocks with flag unset; `normalize: true` → string content / `finish_reason: stop` / `normalized: true` / `refusal: null`; `normalize: false` → native; tools → `tool_calls` with stringified arguments, `finish_reason: tool_calls`, `content: null` on the tool-only turn; extended thinking → `reasoning` string + `reasoning_details` with signatures; **replaying the normalized message verbatim was accepted by the live Anthropic API and answered correctly** — the round-trip contract holds against the real vendor.
- Mistral: policy gate verified live (native keeps `finishReason`; force-on maps, cleans camelCase, stamps `normalized`).
- OpenAI: idempotent under force-on; `normalized: true` stamp only when asked.
- Magistral (Mistral reasoning model): the live API currently inlines reasoning as prose (no ThinkChunk content, no markers) — separable-thinking handling is unit-pinned for when Mistral serves the chunked shape, and the new `custom.prompt_mode` passthrough was verified delivered end-to-end (Mistral's own `3051` mode-not-enabled error returns through the stack).

**Catalog backfill — 58/58 surviving additions answered a live prompt** through the deployment (all 46 Alibaba entries, both gemma-4 models, `chat-latest`, `gpt-4o-2024-11-20`, `zai-glm-5-2`, `deepseek-v4-flash-vision-exp`, `glm-5.3-flash`, plus spot-checked snapshot/rolling aliases). The one entry that failed live routing (`qwen3-vl-flash-2025-10-15`, upstream 400 twice) was removed. A post-edit re-audit of every keyed provider's `/models` endpoint shows **zero stale catalog entries** across all 11 verifiable catalogs.

## Numbers

- **48 files changed, +4,143 / −101.**
- **~53 new catalog entries** (46 Alibaba — incl. 8 third-party-hosted — 2 OpenAI, 2 Gemini/Gemma, 1 DeepSeek, 1 Mistral-hosted, 1 Z.AI) and **~23 new aliases** on existing entries (12 OpenAI dated, 4 Alibaba dated, 6 Gemini rolling, 1 Mistral spelling).
- **~63 new test blocks** across 8 files, plus `normalize: false` assertions added to 4 existing controller tests; the provider-consistency matrix expands to ~67 runs over 23 providers.
- 2 new source modules (`normalizeToOpenAI.ts` at 227 lines, plus its 309-line test); largest single file change is `alibaba/models.ts` (+1,031).


leilacarolyn4-coder (approved):

Salazareo (approved):
code looks good to me

leilacarolyn4-coder:
> code looks good to me

Nice

PR #3664, opened 2026-08-28
fix: honor defaultValue for showFontPicker() in the app environment

### LINKED ISSUE #2896: Add support for `defaultValue` option for `puter.ui.showFontPicker` in puter app environment
Currently the `defaultValue` options only works in web environment. Ideally we match this functionality in puter apps env

app

Image

web

Image

Also somehow our font picker UI is different in web and in puter app env

Best to try and compare it here
https://ui-playground.puter.site/
https://puter.com/app/ui-playground

M001N:
Fixes #2896

`puter.ui.showFontPicker({ defaultValue })` only worked when the SDK
fell back to the standalone `<puter-font-picker>` web component. When
the app runs embedded in the Puter desktop (env=app), the call is
forwarded over postMessage to `IPC.js`, which opens
`UIWindowFontPicker` directly — and that window only ever read
`options.default`, never `defaultFont` or `defaultValue`, so the
pre-selected font was always lost in that environment.

- `src/puter-js/src/modules/UI.js`: `showFontPicker()` now resolves
  `defaultValue`/`defaultFont`/`default` to a single effective value
  once, up front, and uses it identically for both the app (postMessage)
  and standalone (web component) code paths — mirroring the alias
  pattern `showColorPicker()` already uses.
- `src/gui/src/UI/UIWindowFontPicker.js`: now reads
  `options.defaultValue || options.defaultFont || options.default`
  instead of only `options.default`, so the app-embedded picker window
  honors the same aliases the SDK documents.

No behavior change for existing callers using `default`/`defaultFont`.

Tests: `src/gui/**` has no CI test coverage in this repo (verified
against `.github/workflows/*.yaml|yml` — none filter on that path);
verified `src/puter-js` changes with `npm run check:puterjs:types`
and `npm run test:puterjs` locally. Manually traced both the app and
web code paths by hand (no existing test in `components.suite.ts`
covers the font picker's default value either way).

### DIFF
--- a/src/gui/src/UI/UIWindowFontPicker.js
+++ b/src/gui/src/UI/UIWindowFontPicker.js
@@ -41,13 +41,16 @@ const font_list = new Set([
 async function UIWindowFontPicker (options) {
     // set sensible defaults
     if ( arguments.length > 0 ) {
-        // if first argument is a string, then assume it is the default color
+        // if first argument is a string, then assume it is the default font
         if ( window.isString(arguments[0]) ) {
             options = {};
             options.default = arguments[0];
         }
     }
     options = options || {};
+    // `defaultValue` and `defaultFont` are aliases for the legacy `default`
+    // option, matching the names puter.js accepts in showFontPicker().
+    const defaultFont = options.defaultValue || options.defaultFont || options.default;
 
     return new Promise(async (resolve) => {
         let h = '';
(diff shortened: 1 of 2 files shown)


Salazareo:
hi @M001N please sing the CLA

Salazareo:
closing because of no CLA

PR #3667, opened 2026-08-29
:bug: PUT-1655: Open read-only shared files instead of forwarding write_url=undefined

jfcastro92:
## Problem

Opening a file you hold **read-but-not-write** access on (a read-only share) failed silently. For such a file the backend correctly omits `write_url` from the `/open_item` signature, but `launchApp` appended it to the app iframe URL unconditionally. `URLSearchParams` coerces `undefined` to the string `"undefined"`, so the app received `puter.item.write_url="undefined"` — **truthy**, so the editor believed the file was writable, and an **invalid URL**, so it broke on open. Only read-only shares hit this; files you can write carry a real `write_url`.

## Fix

Extract the `puter.item.*` param building into `append_signed_item_params` and guard the `write_url` append so it is only added when present. `launchApp.js` is too coupled to `UIWindow`/jQuery/window globals to unit-test directly, so the pure helper carries the logic and its own test, matching the `helpers/` pattern.

## Changes

- `src/gui/src/helpers/appendSignedItemParams.js` — new pure helper; `write_url` appended only when present.
- `src/gui/src/helpers/launchApp.js` — call the helper instead of the inline appends.
- `src/gui/src/helpers/appendSignedItemParams.test.js` — regression tests.

### Unit (`appendSignedItemParams.test.js`) — 3/3 passing

Pins both directions, and both **fail on the pre-fix code, pass on the fix**:
- a read-only signature omits `write_url` entirely (no `"undefined"`);
- a writable signature still forwards `write_url`;
- the passed path is advertised over the signature path.

### Real end-to-end against a local instance — 7/7 passing

Two real, email-confirmed accounts. **A** wrote `shared.txt` and shared it **read-only** with **B**. **B** opened the genuine share link (`/desktop?shared=…`), which runs the real open path (`handle_shared_link → open_shared_item → open_path_target → POST /open_item → launch_app → append_signed_item_params`) and rendered the file in an editor.

The share, B's access, and the read-only `/open_item` signature (no `write_url`) are all real backend behavior — verified end to end:

- ✅ A shared the file read-only with B
- ✅ an app window opened for the shared file
- ✅ iframe carries `read_url`
- ✅ **iframe has NO `write_url`** (pre-fix: `write_url="undefined"`)
- ✅ no literal `"undefined"` write_url
- ✅ editor shows READ-ONLY mode
- ✅ editor displays the shared file bytes

### Screenshots

_Shared file opens in the editor, read-only:_
put-1655-01-editor-opened
put-1655-02-desktop

### DIFF
--- a/src/gui/src/helpers/appendSignedItemParams.js
+++ b/src/gui/src/helpers/appendSignedItemParams.js
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2024-present Puter Technologies Inc.
+ *
+ * This file is part of Puter.
+ *
+ * Puter is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Append a signed file's `puter.item.*` params onto the launch iframe URL.
+ *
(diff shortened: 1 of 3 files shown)


Salazareo (approved):
nit: prefer camelCase, but I think this looks sane

PR #3679, opened 2026-08-30
feat: durable event subscriptions store, cache, and routes (PUT-1673)

Salazareo:
Stacked on `DS/put-1675` — review only this PR's diff; merges bottom-up.


Salazareo (commented):
ready to go this one I think

PR #3694, opened 2026-08-30
feat: global outbound share listing (PUT-1664)

Salazareo:
Bottom of a stacked-PR chain (targets `main`).

### DIFF
--- a/src/backend/controllers/share/ShareController.ts
+++ b/src/backend/controllers/share/ShareController.ts
@@ -283,6 +283,36 @@ export class ShareController extends PuterController {
         });
     }
 
+    /**
+     * GET /share/shared-by-me — paginated listing of everything the caller has
+     * shared out. `GET /share/shares` answers this for one item at a time,
+     * which cannot answer it at all for a caller who doesn't know what to ask
+     * about.
+     */
+    @Get('/shared-by-me', {
+        subdomain: 'api',
+        requireVerified: true,
+        rateLimit: SHARE_LIST_LIMIT,
+    })
+    async listSharedByMe(req: Request, res: Response): Promise<void> {
+        const actor = this.#requireActor(req);
+        const query = this.#query(req);
+
+        const page = await this.services.share.listSharedByMe(actor, {
+            limit: normalizeLimit(query.limit, { cap: LIST_LIMIT_CAP }),
+            cursor: typeof query.cursor === 'string' ? query.cursor : undefined,
+            includeTotal: query.includeTotal === 'true',
+        });
+
+        res.json({
+            items: await Promise.all(
+                page.items.map((share) => this.#toClientShare(share)),
+            ),
+            ...(page.cursor ? { cursor: page.cursor } : {}),
+            ...(page.total !== undefined ? { total: page.total } : {}),
+        });
+    }
+
     /** GET /share/shares — who can reach one item. */
     @Get('/shares', {
         subdomain: 'api',
--- a/src/backend/controllers/share/ShareController.ts
+++ b/src/backend/controllers/share/ShareController.ts
@@ -265,22 +265,9 @@ export class ShareController extends PuterController {
         rateLimit: SHARE_LIST_LIMIT,
     })
     async listSharedWithMe(req: Request, res: Response): Promise<void> {
-        const actor = this.#requireActor(req);
-        const query = this.#query(req);
-
-        const page = await this.services.share.listSharedWithMe(actor, {
-            limit: normalizeLimit(query.limit, { cap: LIST_LIMIT_CAP }),
-            cursor: typeof query.cursor === 'string' ? query.cursor : undefined,
-            includeTotal: query.includeTotal === 'true',
-        });
-
-        res.json({
-            items: await Promise.all(
-                page.items.map((share) => this.#toClientShare(share)),
-            ),
-            ...(page.cursor ? { cursor: page.cursor } : {}),
-            ...(page.total !== undefined ? { total: page.total } : {}),
-        });
+        await this.#listSharePage(req, res, (actor, opts) =>
+            this.services.share.listSharedWithMe(actor, opts),
+        );
     }
 
     /**
@@ -295,10 +282,31 @@ export class ShareController extends PuterController {
         rateLimit: SHARE_LIST_LIMIT,
     })
     async listSharedByMe(req: Request, res: Response): Promise<void> {
+        await this.#listSharePage(req, res, (actor, opts) =>
+            this.services.share.listSharedByMe(actor, opts),
+        );
+    }
+
+    /**
+     * Parse-and-shape shared by the paginated share listings, so the two cannot
+     * drift: same query contract in, same envelope out.
+     */
+    async #listSharePage(
+        req: Request,
+        res: Response,
+        list: (
+            actor: Actor,
+            opts: { limit?: number; cursor?: string; includeTotal?: boolean },
+        ) => Promise<{
+            items: ResolvedShare[];
+            cursor?: string;
+            total?: number;
+        }>,
+    ): Promise<void> {
         const actor = this.#requireActor(req);
         const query = this.#query(req);
 
-        const page = await this.services.share.listSharedByMe(actor, {
+        const page = await list(actor, {
             limit: normalizeLimit(query.limit, { cap: LIST_LIMIT_CAP }),
             cursor: typeof query.cursor === 'string' ? query.cursor : undefined,
             includeTotal: query.includeTotal === 'true',
--- a/src/backend/services/share/ShareService.ts
+++ b/src/backend/services/share/ShareService.ts
@@ -21,6 +21,7 @@ import { contentType as contentTypeFromMime } from 'mime-types';
 import { posix as pathPosix } from 'node:path';
 import { userRelatedActor, type Actor } from '../../core/actor';
 import { HttpError, isHttpError } from '../../core/http/HttpError.js';
+import { runWithConcurrencyLimitSettled } from '../../util/concurrency.js';
 import { isUniqueViolation } from '../../util/dbError.js';
 import {
     abuseKey,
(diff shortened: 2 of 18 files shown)


src/backend/services/share/ShareService.ts
```diff
@@ -1324,6 +1333,118 @@ export class ShareService extends PuterService {
...
+            ),
+            this.#reachableBy(actor, [...entries.values()]),
+        ]);
+
+        const items: ResolvedShare[] = [];
+        for (const row of rows) {
+            const entry = entries.get(Number(row.fsentry_id));
+            if (!entry) continue;
+            if (!reachable.has(entry.uuid)) continue;
+            const pending = !row.holder_user_id;
```
jfcastro92, line 1524: I would hold the PR on this one. Pending rows skip the `stillReaches` check by construction, and nothing else catches them: `#revokeDownstream` goes via `listByFsentrySubtree` (filters `holder_user_id IS NOT NULL`) and `deleteActive` is holder-keyed, so revoking a manage delegate never deletes the invites they sent. `#reachableBy` is a no-op for plain sessions, so there is no caller-side gate left.

So after the owner revokes D, D can still call this forever and read the file's current name, size, mtime and a fresh signed thumbnail URL. listSharesOf and listSharedWithMe both gate the caller; this path does not. Cleanest fix is probably deleting pending invites when their issuer's grant goes away, since those orphans are a problem either way.

src/backend/services/share/ShareService.ts
```diff
@@ -1324,6 +1333,118 @@ export class ShareService extends PuterService {
...
+            if (
+                !pending &&
+                !stillReaches.has(`${Number(row.holder_user_id)}:${entry.id}`)
```
jfcastro92: `stillReaches` is keyed `holder:entry`, so it answers "can H still reach N through anyone", not "is my grant to H still alive". If O and D both shared N with H and O's permission row goes without the share row being deleted (`onGrantRevoked` bails early when `issuerId` is not a number), D's grant keeps the pair live and O sees a dead share they cannot clear.

`#liveGrants` already has `issuer_user_id` on hand (`#hasGrantFrom` uses it). listSharesOf is equally coarse, so I would fix it inside `#reachingHolders` rather than here.

src/backend/services/share/ShareService.ts
```diff
@@ -1324,6 +1333,118 @@ export class ShareService extends PuterService {
...
+                ...(pending
+                    ? {
+                          pending: true,
+                          recipientEmail:
+                              (row.data as { invitedAddress?: string } | null)
+                                  ?.invitedAddress ?? row.recipient_email,
+                      }
+                    : {}),
+                createdAt: row.created_at,
+                issuedByApp: issuedByApp(row),
```
jfcastro92: Always null for pending rows: `upsertPending` writes `{ issuerAppUid }` into `data` (ShareStore.js:375), but this reads `data.issuedByApp`, the key `upsertActive` uses. So an app-sent invite looks self-issued, which is the one case the field exists for.

Pre-existing (listSharesOf:1595 too), but this is where someone would audit what an app did in their name.

jfcastro92 (commented):
Review from Aug 31 against 7ce4acbb8, the branch head at the time. All ten points were addressed in `1725fd12c` ("fix: address review on outbound share listing") before this merged as 66a975f65.

I deleted the inline comments on a bad inference — I saw the fixes in main and concluded I had reviewed a stale head post-merge, which was wrong. The feedback was live and it was acted on. The merge commit body records what each fix changed, and the mapping comment below records where each landed.

src/backend/services/share/ShareService.ts
```diff
@@ -1324,6 +1333,118 @@ export class ShareService extends PuterService {
...
+        ]);
+
+        const nodeById = new Map<number, FSEntry>(
+            [...entries.values()].map((entry) => [entry.id, entry]),
+        );
+        // Same two bounds as the inbound listing: the grant behind a row has to
+        // still be there, and an app sees only the part of it the credential
+        // reaches in its own right.
+        const [stillReaches, reachable] = await Promise.all([
+            this.#reachingHolders(
```
jfcastro92: Fan-out worth a look: `#reachingHolders` does one `#liveGrants` per distinct holder in an unbounded `Promise.all`, ~2 reads each. Inbound only ever had one holder, this can be 200 a page, and 600 req/min on `share:list` makes it cheap to trigger. Cold cache that is ~200 SELECTs queued through the batcher per request.

The flat-perm keys already embed the holder id, so one mget plus a `holder_user_id IN (...)` read would do a whole page in ~2.

src/backend/controllers/share/ShareController.ts
```diff
@@ -283,6 +283,36 @@ export class ShareController extends PuterController {
         });
     }
 
+    /**
+     * GET /share/shared-by-me — paginated listing of everything the caller has
+     * shared out. `GET /share/shares` answers this for one item at a time,
+     * which cannot answer it at all for a caller who doesn't know what to ask
+     * about.
+     */
+    @Get('/shared-by-me', {
```
jfcastro92, line 279: AGENTS.md sends anything apps can call to doc/contributing-apis.md, which says the PR is not done until all seven steps are, including the puter.js method, docs page and SDK suite. Nothing under src/puter-js or src/docs here.

Not backend-only by precedent either: `/shared-with-me` ships as `puter.fs.listShared()` with docs. As-is no app can reach this listing.

src/backend/controllers/share/ShareController.ts
```diff
@@ -283,6 +283,36 @@ export class ShareController extends PuterController {
...
+        subdomain: 'api',
+        requireVerified: true,
+        rateLimit: SHARE_LIST_LIMIT,
```
jfcastro92, line 282: This joins the shared 600/min `share:list` bucket, but rate-limits-and-quotas.md:138 still lists only `getShares`/`listShared`, and AGENTS.md says a limit is not done until the docs are. Since the bucket is shared, polling this quietly eats the budget other share reads need.

src/backend/stores/share/ShareStore.js
```diff
@@ -103,6 +99,72 @@ export class ShareStore extends PuterStore {
         };
     }
 
+    /**
+     * Shares a user has made, keyset-paginated on `id`: the rows they issued
+     * themselves, plus the rows a manage delegate issued on a node they own —
+     * which is the only place those are visible, since the permission tables
+     * are keyed issuer to holder.
+     *
+     * Two indexed reads rather than one `OR` spanning the join: the optimizer
```
jfcastro92: The comment promises a plan the migrations do not back. SQLite has no issuer index at all (0014 creates none; 0067 adds holder, fsentry and the unique triple), so the issued half walks the PK there.

Bigger one: the delegated half joins `fsentries.user_id` then orders by `share.id`, which no index gives us in any dialect, so each page re-sorts every share on every node the user owns. `idx_share_fsentry` only helps the join probe. Either add the composite issuer index or reword this.

src/backend/services/share/ShareService.ts
```diff
@@ -1324,6 +1333,118 @@ export class ShareService extends PuterService {
...
+            if (!entry) continue;
+            if (!reachable.has(entry.uuid)) continue;
+            const pending = !row.holder_user_id;
+            if (
+                !pending &&
+                !stillReaches.has(`${Number(row.holder_user_id)}:${entry.id}`)
+            ) {
+                continue;
+            }
+            items.push({
```
jfcastro92: Fifth hand-built `ResolvedShare` in this file (inbound at 1307, three in listSharesOf), and they have drifted: inbound has name/type/thumbnail/owner but no issuedByApp/inheritedFrom, listSharesOf the reverse, this one the union. Same object, different shape per endpoint.

Something like `#resolveShareRows(rows, { trash, grantCheck, pending })` would make the differences arguments instead of three near-identical loops.

src/backend/controllers/share/ShareController.ts
```diff
@@ -283,6 +283,36 @@ export class ShareController extends PuterController {
...
+    })
+    async listSharedByMe(req: Request, res: Response): Promise<void> {
```
jfcastro92, line 284: `listSharedWithMe` above with one line changed. You added a test to keep the decorator options in sync, so the drift risk is clearly real, but nothing guards the parsing or response shaping. A small `#listSharePage(req, res, list)` used by both would cover all of it.

src/backend/stores/share/ShareStore.js
```diff
@@ -530,6 +592,20 @@ export class ShareStore extends PuterStore {
...
+    /** @param {number} [limit] */
+    #pageSize(limit) {
+        return Math.min(
+            Math.max(1, Math.floor(Number(limit) || DEFAULT_PAGE_SIZE)),
+            MAX_PAGE_SIZE,
+        );
+    }
+
+    /** The id a keyset page resumes after; 0 for the first page. */
+    #afterId(cursor) {
```
jfcastro92, line 646: Minor and pre-existing, but it now backs two endpoints. `decodeCursor` only 400s on unparseable input, so `{}`, `{"id":"abc"}` or another endpoint's cursor all decode and land here as 0, silently restarting from page 1 — a client that walked N pages re-ingests everything.

Same pattern is in AppDriver/FSEntryStore etc, so not a blocker; just cheaper to validate once here than in four places later.

jfcastro92:
Retracting my review above — apologies for the noise. I reviewed commit 7ce4acbb8, which was a stale branch head; this PR merged on Sep 1 as 66a975f65, and the merged version already handles every point I raised. For the record, matched against origin/main:

- Pending invites bypassing the liveness gate -> fixed by `#pendingStillAuthorized`, gated at `if (pending && !pendingAllowed.has(row.uid)) continue;`
- Holder-aggregate liveness -> `#reachingGrants` now keys `holder:entry:issuer`
- `issuerAppUid` vs `issuedByApp` -> now writes the `issuedByApp` key
- Per-holder fan-out in `#reachingHolders` -> batched via `readLinkedUserUserPermsForHolders`
- Missing puter.js method and docs -> `listSharedByMe.js` plus `src/docs/src/FS/listSharedByMe.md`
- Rate-limit table -> row now reads `getShares`, `listShared`, `listSharedByMe`
- The index comment I pushed back on -> `0071_share_issuer_index.sql` and `postgres_mig_14.sql` add `idx_share_issuer (issuer_user_id, id)`, and the comment is reworded to match
- Fifth inline `ResolvedShare` -> extracted as `#resolvedShareRow`
- Duplicated controller method -> extracted as `#listSharePage`
- Cursor shape validation -> `#afterId` now checks the decoded payload

One operational note that outlives the review: `0071_share_issuer_index.sql` is registered at threshold 66, so any database below user_version 66 is running the outbound listing without `idx_share_issuer`. On sqlite that is a full table walk on the issued half. Worth confirming the migration has actually run wherever this is deployed, not just that it is in main.

PR #3696, opened 2026-08-30
fix: cover the apps summary's no-app-group first page (PUT-1670)

Salazareo:
listOutboundApps sorts the no-app group first via an empty-string
sentinel. Add a regression test pinning that a first page (no cursor)
actually returns it, and that the cursor it hands back resumes past it
into the app-keyed groups rather than skipping or repeating.

---
Stacked on `DS/put-1670` — review only this PR's diff; merges bottom-up.


jfcastro92 (approved):
lgtm

PR #3699, opened 2026-08-31
allow concept of "home regions"

ProgrammerIn-wonderland:


config.template.jsonc
```diff
@@ -142,6 +142,13 @@
             }
         }
     },
+    // Optional: where each node stores files, keyed by node id. A write can
+    // name one to place bytes near the account that owns them. Omit it and
+    // every write uses `s3_bucket` / `s3_region` below.
+    "_servers_example": {
```
Salazareo: why _servers_example? this should just be the actual example no?

Salazareo (approved):
small nit on the config there, but seems ok

PR #3709, opened 2026-09-01
🏗️ PUT-1702: TeamStore — workspace CRUD, handle uniqueness, reserved words

jfcastro92:
> Stacked on #3708 (PUT-1701) → #3705 → #3704. Review the schema stack first; the diff here is only `TeamStore`.

First code of phase 2. `GroupStore` has only `addUsers`/`removeUsers` — nothing creates, reads back or lists a group at runtime. `TeamStore` is that missing half, scoped to rows with `kind = 'team'`.

```
getByUid(uid)            create({ ownerUserId, name, handle?, planId? })
getByHandle(handle)      update(uid, { name?, handle? })
isHandleAvailable(h)     softDelete(uid)
listByOwner(userId)
```

## A workspace is addressed by `uid`, not by handle

`group.uid` has been `NOT NULL UNIQUE` since `0015`, so the identity already existed — this just uses it. `handle` is a mutable display label with no addressing role.

That matters because the alternative fails in a way worse than an error. If routes keyed on handle, a rename would invalidate every stored reference — and once the freed handle was claimed by someone else, an old reference would not 404, it would resolve to a **different workspace**. There is a test (`keeps the uid valid after the handle changes`) pinning the property.

⚠ **This changes the API surface** already written up in the design doc and in PUT-1708 / PUT-1735 / PUT-1736 / PUT-1746: every route and SDK method now takes `:uid` instead of `:handle`. Those are updated. One consequence is worth a second opinion — see the open question at the bottom.

## Soft delete releases the handle

`softDelete` sets `deleted_at` **and** nulls `handle`, keeping `name`.

Without this, a global unique index with no dead-row exclusion reserves a deleted workspace's handle permanently. **MySQL has no partial indexes**, so "exclude soft-deleted rows from the index" was never available as a fix — it had to be handled in the write path. Since nothing addresses a workspace by handle, releasing it dangles nothing; `name` survives so the row still reads sensibly in history.

## Handle validation

`^[a-z0-9]+(-[a-z0-9]+)*$`, 3–64 characters, plus a 42-entry reserved list (`admin`, `puter-support`, `security`, `billing`, …) because a handle appears in activation email and share dialogs, where one that reads like Puter itself makes impersonation more convincing.

The charset is **deliberately narrower than the column accepts**, and that is the interesting part. Case-insensitive uniqueness is spelled three different ways across our engines, and they do not mean the same thing:

| | Mechanism | Also folds |
| --- | --- | --- |
| mysql | `utf8mb4_unicode_ci` | accents, `ß`→`ss` |
| sqlite | `COLLATE NOCASE` | ASCII only |
| postgres | `lower(handle)` | full Unicode, accents preserved |

So `café` / `cafe` collide on mysql alone, and `ÄÖ` / `äö` on mysql and postgres but not sqlite — a handle accepted on self-hosted sqlite could be refused on prod mysql. Restricting to `[a-z0-9-]` makes all three differences unreachable rather than relying on the engines to agree.

## Two details a reviewer should check

**Every read filters `kind = 'team' AND deleted_at IS NULL`.** That predicate is what makes the seeded `admin` / `system` groups *unreachable* rather than merely absent — the criterion inherited from PUT-1699. `never returns a seeded system group from any method` walks every `kind IS NULL` row and asserts the store returns null for each. A team admin is not a platform admin.

**Handle lookups are dialect-aware.** Postgres indexes `lower(handle)`, so `#handleMatch()` compares `lower(handle) = lower(?)` there and a plain `=` elsewhere, where the column collation already handles it.

---

### sqlite and postgres — both verified

```
$ npx vitest run --config src/backend/vitest.config.ts src/backend/stores/team/
 Test Files  1 passed (1)
      Tests  17 passed (17)

$ PUTER_TEST_DB_ENGINE=postgres npx vitest run --config src/backend/vitest.config.ts src/backend/stores/team/
 Test Files  1 passed (1)
      Tests  17 passed (17)
```

<details>
<summary>Confirming the postgres branch is load-bearing, not defensive</summary>

Removing the dialect branch so `#handleMatch()` always returns `` `handle` = ? ``, then re-running on postgres:

```
 ❯ TeamStore.test.ts (17 tests | 1 failed)
     × resolves a handle case-insensitively
 Tests  1 failed | 16 passed (17)
```

Note which test survives: `refuses a duplicate handle, including one differing only in case` still passes, because the **unique index** catches that regardless. It is the *lookup* that breaks. That is the same distinction flagged in #3704 — uniqueness and lookup are separately case-sensitive, and getting one right does not give you the other.
</details>

### Full suite and typecheck

```
$ npm run test:backend
 Test Files  251 passed | 24 skipped (275)
      Tests  6718 passed | 26 skipped (6744)      # +17, no regressions

$ npm run typecheck
```

---

## Open question this raises — not decided here

`PUT-1726` specifies sharing as `puter.fs.share('/me/Docs', { team: 'design-team' }, 'write')` — a **handle**.

That is genuinely not the same case as a route: it is typed by a person and resolved immediately, so it is an input rather than a stored reference, and a uuid there would be miserable to write. But a **scripted** share to a handle keeps working after that handle is released and reclaimed, and would silently start sharing with an unrelated workspace. A 404 would be safe; retargeting is not.

Left open in PUT-1726 and §11 of the design doc rather than settled quietly. Three options are written up there.

---

Closes PUT-1702.

### DIFF
--- a/src/backend/stores/team/TeamStore.ts
+++ b/src/backend/stores/team/TeamStore.ts
@@ -1,0 +1,239 @@
+/*
+ * Copyright (C) 2024-present Puter Technologies Inc.
+ *
+ * This file is part of Puter.
+ *
+ * Puter is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import { v4 as uuidv4 } from 'uuid';
+import { PuterStore } from '../types';
+
+/** A workspace: a `group` row with `kind = 'team'`. */
+export interface TeamRow {
+    id: number;
+    uid: string;
+    owner_user_id: number;
+    kind: string;
+    name: string | null;
+    handle: string | null;
+    deleted_at: string | null;
+    created_at: string;
+}
+
+export const TEAM_KIND = 'team';
+
+/** Longest handle mysql can store — `varchar(64)` in mysql_mig_26. */
+export const HANDLE_MAX_LENGTH = 64;
+export const HANDLE_MIN_LENGTH = 3;
+
+/** Mysql declares `name varchar(255)`; sqlite and postgres use TEXT. */
+export const NAME_MAX_LENGTH = 255;
+
+/** Narrower than the column, so the engines' collations cannot disagree. */
+const HANDLE_PATTERN = /^[a-z0-9]+(?:-[a-z0-9]+)*$/u;
+
+/** A handle reaching users in email makes an impersonation more convincing. */
+const RESERVED_HANDLES = new Set([
+    'about',
+    'account',
+    'admin',
+    'administrator',
+    'api',
+    'app',
+    'apps',
+    'billing',
+    'blog',
+    'ceo',
+    'contact',
+    'dashboard',
+    'dev',
+    'developer',
+    'docs',
+    'help',
+    'internal',
+    'legal',
+    'login',
+    'mail',
+    'moderator',
+    'official',
+    'owner',
+    'payment',
+    'payments',
+    'puter',
+    'puter-support',
+    'puterteam',
+    'root',
+    'security',
+    'settings',
+    'signup',
+    'staff',
+    'status',
+    'support',
+    'system',
+    'team',
+    'teams',
+    'trust',
+    'trust-safety',
+    'verify',
+    'workspace',
+    'workspaces',
+]);
+
+export type HandleRejection =
+    'too_short' | 'too_long' | 'malformed' | 'reserved';
+
+/** Trimmed and capped, so the same name is accepted on every engine. */
+export const normalizeTeamName = (name: string): string => {
+    const trimmed = name.trim();
+    if (trimmed === '') throw new Error('team name is required');
+    if (trimmed.length > NAME_MAX_LENGTH)
+        throw new Error('team name is too long');
+    return trimmed;
+};
+
+/** Why `handle` is unusable, or null when it is fine. */
+export const checkHandle = (handle: string): HandleRejection | null => {
+    if (handle.length < HANDLE_MIN_LENGTH) return 'too_short';
+    if (handle.length > HANDLE_MAX_LENGTH) return 'too_long';
+    if (!HANDLE_PATTERN.test(handle)) return 'malformed';
+    if (RESERVED_HANDLES.has(handle)) return 'reserved';
+    return null;
+};
+
+export class TeamStore extends PuterStore {
+    /**
+     * Postgres indexes `lower(handle)`; the others are case-insensitive
+     * already.
+     */
+    #handleMatch(): string {
+        return this.clients.db.case({
+            postgres: 'lower(`handle`) = lower(?)',
+            otherwise: '`handle` = ?',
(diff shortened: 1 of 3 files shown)


src/backend/stores/team/TeamStore.ts
```diff
@@ -0,0 +1,250 @@
...
+export const HANDLE_MIN_LENGTH = 3;
+
+/** Mysql declares `name varchar(255)`; sqlite and postgres use TEXT. */
+export const NAME_MAX_LENGTH = 255;
+
+/** Narrower than the column, so the engines' collations cannot disagree. */
+const HANDLE_PATTERN = /^[a-z0-9]+(?:-[a-z0-9]+)*$/u;
+
+/** A handle reaching users in email makes an impersonation more convincing. */
+const RESERVED_HANDLES = new Set([
```
Salazareo, line 48: should this maybe be in config to expand?

can leave as is for now tho

Salazareo (approved):
since were calling feature "team" maybe we change table name to teams also, so groups can be more akin to the workspace concept of groups, like within a team?

src/backend/stores/team/TeamStore.ts
jfcastro92: Good call on the naming — I went back and forth on this one.

What decided it for me is that group isn't really ours: it's been there since migration 0015, along with jct_user_group and user_to_group_permissions, with the system groups seeded in 0025 and ~45k membership rows in prod today. The permission scan already joins those two tables, and reusing that path is most of the reason this is 13 PRs instead of 40. A separate teams table would mean either duplicating that scan for a second entity, or FK'ing back to group anyway.

On groups-within-a-team, I think kind actually gets us there: it's kind='team' today, and a later kind='subgroup' with a parent_group_id would inherit the same membership and permission path for free. A separate table would make that nesting harder, not easier.

Naming-wise I do agree with you, which is why everything above the DB says team — TeamStore, TeamService, TeamController, /teams. Only the physical table still says group. Happy to revisit if subgroups end up wanting columns that don't fit.

jfcastro92:
## Local validation

Engine: mysql 8 · `teams_enabled: true`

**Create, read, rename**

```
POST /teams {"name":"Acme Corp","handle":"acme"}
  -> 200 {"uid":"13341565-5223-4c0a-9d28-ca71c9da6682","name":"Acme Corp",
          "handle":"acme","is_owner":true,"created_at":"2026-09-03T23:25:43.000Z"}
PUT  /teams/<uid> {"name":"Acme Renamed"} -> 200, name updated
```

**Handle uniqueness across owners** — a second user taking the same handle:

```
POST /teams {"name":"Other Acme","handle":"acme"}
  -> 409 {"error":"That handle is taken","code":"conflict"}
```

A conflict, not a 500 — the unique index is caught and translated.

**Handle validation runs before the index.** `HANDLE_PATTERN` is
`/^[a-z0-9]+(?:-[a-z0-9]+)*$/`, commented *"Narrower than the column, so the
engines' collations cannot disagree"* — and that is what happens in practice:

```
POST /teams {"handle":"ACME"} -> 400 {"error":"Unusable handle: malformed","code":"bad_request"}
```

Worth noting for review: this means a case-differing handle is rejected at
validation and never reaches the case-insensitive index, so on mysql the
`utf8mb4_unicode_ci` collation is defense-in-depth rather than the primary
mechanism. That is the stronger arrangement, and it is the first time the mysql
collation has been exercised at all — the automated suite runs sqlite and
postgres only.

**Reserved handles**

```
admin  -> 400 {"error":"Unusable handle: reserved","code":"bad_request"}
system -> 400 {"error":"Unusable handle: reserved","code":"bad_request"}
api    -> 400 {"error":"Unusable handle: reserved","code":"bad_request"}
```

**One observation, not a defect:** `www` is accepted. That is consistent with the
list's stated purpose — the docblock says *"A handle reaching users in email makes
an impersonation more convincing"*, so the list guards impersonation, not routing.
If handles ever become subdomains or path prefixes (open question 2 in the design
doc), the list would want a routing-oriented pass — `www`, `cdn`, `static`,
`assets`, `site`, `host`. Flagging so the decision is deliberate rather than
inherited.

PR #3710, opened 2026-09-01
🏗️ PUT-1703: TeamStore — membership reads and writes (single writer)

jfcastro92:
> Fifth in the stack: #3704 → #3705 → #3708 → #3709 → this. Review bottom-up; the diff here is the membership half of `TeamStore`.

The management side of workspace membership. `readUserGroupPerms` already joins `jct_user_group` and resolves group grants for the permission scan — that path is untouched.

```
getMembership(teamUid, userId)      addMember(teamUid, userId, { orgOwned })
isMember(teamUid, userId)           removeMember(teamUid, userId)
listMembers(teamUid, { limit, cursor })
listTeamsForUser(userId)
```

## The "single writer" decision

The ticket asked to decide whether `TeamStore` supersedes `GroupStore.addUsers`/`removeUsers`, and not to leave two writers.

**First, what the callers actually do.** All five production call sites target a *seeded system group*:

| Call site | Group |
| --- | --- |
| `AuthController` signup ×2, save-account ×1 | `default_user_group`, `default_temp_group` |
| `OIDCService` first login | `default_user_group` |
| `DefaultUserService` bootstrap | `ADMIN_GROUP_UID` |

Never a team — those uids come from config and constants. So the two stores were already disjoint in practice, and the ticket's stated fear ("the one without the conflict clause will eventually be the one that is called") was defused by #3705 making `addUsers` conflict-tolerant.

**So rather than delete `GroupStore` and churn the signup and OIDC paths inside a teams PR, the split is now enforced in SQL:**

- `TeamStore` writes select `group_id` from a `kind = 'team'` subquery.
- `GroupStore.addUsers`/`removeUsers` carry `AND kind IS NULL`.

Neither store can reach the other's rows. That costs **zero extra queries** — it folds into subqueries both methods already had — and a team uid becomes a no-op, which is how `addUsers` already treats an unknown username (there's a pre-existing test for that). Two new tests on the `GroupStore` side assert it cannot touch a team.

⚠ If you'd rather have one literal writer and accept the blast radius, that's a `GroupStore` deletion plus five call-site rewrites through signup, OIDC and the self-hosted bootstrap. I judged that too invasive to smuggle into this PR, but say so and I'll do it separately.

## A cross-dialect bug worth seeing

`addMember` first used `db.booleanValue(orgOwned)`. That looked right and passed on sqlite — 27/27 green.

`PostgresDatabaseClient` **overrides** `booleanValue` to return a real boolean, because postgres has a real `boolean` type. But `org_owned` is an integer flag in all three dialects (`tinyint(1)` / `INTEGER` / `smallint`), so postgres rejected it outright:

```
error: invalid input syntax for type smallint: "true"
  Tests  10 failed | 26 passed (36)
```

SQLite is loosely typed and stored `true` without complaint. Now passing `1`/`0` explicitly, with a comment saying why so nobody "tidies" it back. This only surfaced because both engines were run.

## Other details

`org_owned` is written here but **never accepted from a request** — `TeamService` sets it at provisioning and workspace creation. It decides who pays, not who may read, and a test asserts the workspace owner's row carries `0` while a provisioned account carries `1`.

`listMembers` is keyset-paginated on `id` per `doc/pagination.md`, using the shared `encodeCursor` / `decodeCursor` / `normalizeLimit` helpers and fetching one row past the limit to decide whether a cursor is warranted.

Reads inherit the `kind = 'team' AND deleted_at IS NULL` predicate, so soft-deleting a workspace drops it out of `listTeamsForUser` and `isMember` without touching the membership rows.

---

### Both engines

```
$ npx vitest run --config src/backend/vitest.config.ts src/backend/stores/team/ src/backend/stores/group/
 Test Files  2 passed (2)
      Tests  36 passed (36)

$ PUTER_TEST_DB_ENGINE=postgres npx vitest run --config src/backend/vitest.config.ts src/backend/stores/team/ src/backend/stores/group/
 Test Files  2 passed (2)
      Tests  36 passed (36)
```

Eleven tests added — membership round-trip, `org_owned` distinguishing the workspace owner, repeat-add being a no-op, removal reporting whether a row existed, workspace scoping, soft-delete dropping out of the user's list, `addMember` refusing a non-team group, and two on the `GroupStore` side asserting the boundary.

Two pagination tests: a 5-member workspace walked in pages of 2 terminates in exactly 3 pages with every member seen once, and `limit: 100_000` is capped at 200 rather than trusted.

<details>
<summary>Every write to `jct_user_group` in the codebase, after this change</summary>

```
GroupStore.ts:61   DELETE FROM `jct_user_group`        <- system groups (kind IS NULL)
GroupStore.ts:46   INSERT ... INTO `jct_user_group`    <- system groups (kind IS NULL)
TeamStore.ts:349   DELETE FROM `jct_user_group`        <- teams (kind = 'team')
TeamStore.ts:330   INSERT ... INTO `jct_user_group`    <- teams (kind = 'team')
```

Two stores, disjoint by predicate, one writer per domain.
</details>

### Full suite and typecheck

```
$ npm run test:backend
 Test Files  251 passed | 24 skipped (275)
      Tests  6730 passed | 26 skipped (6756)      # +12, no regressions

$ npm run typecheck
```

---

Closes PUT-1703.

### DIFF
--- a/src/backend/stores/group/GroupStore.test.ts
+++ b/src/backend/stores/group/GroupStore.test.ts
@@ -158,4 +158,31 @@ describe('GroupStore', () => {
         expect(await memberUsernames(a)).toEqual([]);
         expect(await memberUsernames(b)).toEqual([member.username]);
     });
+    // -- the boundary with TeamStore ----------------------------------
+
+    it('refuses to add a member to a team, leaving that to TeamStore', async () => {
+        const uid = uuidv4();
+        await server.clients.db.write(
+            'INSERT INTO `group` (`uid`, `owner_user_id`, `kind`, `name`, `extra`, `metadata`) ' +
+                'VALUES (?, ?, ?, ?, ?, ?)',
+            [uid, owner.id, 'team', 'Not Yours', '{}', '{}'],
+        );
+
+        // Two writers produced the duplicates; the split is enforced in SQL.
+        await store.addUsers(uid, [member.username]);
+        expect(await memberUsernames(uid)).toEqual([]);
+    });
+
+    it('refuses to remove a member from a team', async () => {
+        const uid = uuidv4();
+        await server.clients.db.write(
+            'INSERT INTO `group` (`uid`, `owner_user_id`, `kind`, `name`, `extra`, `metadata`) ' +
+                'VALUES (?, ?, ?, ?, ?, ?)',
+            [uid, owner.id, 'team', 'Also Not Yours', '{}', '{}'],
+        );
+        await server.stores.team.addMember(uid, member.id, { orgOwned: true });
+
+        await store.removeUsers(uid, [member.username]);
+        expect(await memberUsernames(uid)).toEqual([member.username]);
+    });
 });
--- a/src/backend/stores/group/GroupStore.ts
+++ b/src/backend/stores/group/GroupStore.ts
@@ -33,36 +33,29 @@ import { PuterStore } from '../types';
  * which joins the junction table itself.
  */
 export class GroupStore extends PuterStore {
-    /**
-     * Adds users (by username) to the group identified by `uid`. No-op if
-     * `usernames` is empty, and for a user who is already a member.
-     */
+    /** Adds users to a seeded system group. No-op for a team or a member. */
     async addUsers(uid: string, usernames: string[]): Promise<void> {
         if (usernames.length === 0) return;
         const placeholders = `(${usernames.map(() => '?').join(', ')})`;
-        // Ignore conflicts on the unique pair index from 0072; re-adding a member
-        // was a duplicate row before it, and would raise without this.
+        // Ignore conflicts on the unique pair index from 0072.
         await this.clients.db.write(
             `${this.clients.db.insertIgnoreInto('jct_user_group')} ` +
                 '(`user_id`, `group_id`) ' +
                 'SELECT u.id, g.id FROM `user` u ' +
-                'JOIN (SELECT id FROM `group` WHERE uid = ?) g ON 1 = 1 ' +
+                'JOIN (SELECT id FROM `group` WHERE uid = ? AND kind IS NULL) g ON 1 = 1 ' +
                 `WHERE u.username IN ${placeholders}` +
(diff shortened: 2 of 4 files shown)


Salazareo (approved):

jfcastro92:
## Local validation

Engine: mysql 8 · workspace with 4 members (1 owner + 3 provisioned seats)

**Full listing**

```
GET /teams/<uid>/members -> 200
{"items":[{"username":"tmowner","org_owned":false,"created_at":"2026-09-03T23:25:43.000Z"}]}
```

…before seats existed, and after provisioning, paginated below.

**Keyset pagination, page 1** (`?limit=2`)

```json
{"items":[{"username":"tmowner",  "org_owned":false,"created_at":"2026-09-03T23:25:43.000Z"},
          {"username":"acmeseat1","org_owned":true, "created_at":"2026-09-03T23:27:28.000Z"}],
 "cursor":"eyJpZCI6ODh9"}
```

**Page 2** (`?limit=2&cursor=eyJpZCI6ODh9`)

```json
{"items":[{"username":"acmeseat2","org_owned":true,"created_at":"2026-09-03T23:27:29.000Z"},
          {"username":"acmeseat3","org_owned":true,"created_at":"2026-09-03T23:27:29.000Z"}]}
```

Three properties worth calling out, all matching `doc/pagination.md`:

- **No overlap** — page1 `['tmowner','acmeseat1']`, page2 `['acmeseat2','acmeseat3']`, intersection empty.
- **The last page omits `cursor` entirely** rather than returning a null or an empty
  string, so "is there more" is a key-presence test.
- **Ordering is stable across the page boundary** even though `acmeseat2` and
  `acmeseat3` share a `created_at` to the second — the cursor is keyed on `id`
  (`eyJpZCI6ODh9` decodes to `{"id":88}`), not on the timestamp, so the tie does not
  cause a repeat or a skip. That is the case a same-second insert would break under
  timestamp keying, and it is exercised here by real rows rather than by fixture.

`org_owned` is correctly `false` for the owner and `true` for provisioned seats,
returned as a JSON boolean rather than mysql's `1`/`0`.

PR #3713, opened 2026-09-01
🏗️ PUT-1705: TeamService — provision an org account

jfcastro92:
> Seventh in the stack: #3704 → #3705 → #3708 → #3709 → #3710 → #3712 → this. Review bottom-up.

**PUT-1705** on its own. This is the operation the feature exists for, it creates real user accounts and sends mail, so it doesn't share a review with anything.

```
provisionAccount(teamUid, actorUserId, { username, email })
    -> { userId, username, activationLink }
suggestUsernames(username, count?)
resendActivation(teamUid, actorUserId, targetUserId)
```

## Activation reuses password recovery

No new token machinery, no `team_activation` table, no unauthenticated endpoint on the team surface. Provisioning sets the same `pass_recovery_token` that `/send-pass-recovery-email` uses, signs the same one-hour `otp` JWT with `purpose: 'pass-recovery'`, and hands back the same `/action/set-new-password` link.

That inherits, for free: a one-shot token cleared on use, expiry carried in a signed JWT rather than a DB column, atomic single-use consumption, purpose scoping so the link can't be replayed for another operation, and refusal for suspended accounts.

**Activation state needs no column either.** An unactivated account is one with no password — the predicate migration `0066` already uses. `resendActivation` refuses an account that has one, because re-sending then would be a password reset wearing the wrong name.

## Usernames come from the global pool

Usernames are globally unique and case-insensitive (`idx_user_username_nocase`, `0055`), so a workspace provisions from the same namespace as every other Puter user. Two companies cannot both have a `juan`.

What provisioning adds is a check and a suggestion — **never a silent modification**. A suffixed `juan-castro-2` would appear in every share dialog that person ever sees, and they never agreed to it. So a taken name is refused with `username_already_in_use` plus alternatives that have been *verified free*, and the administrator chooses.

⚠ **The check runs before any write.** A taken name failing part-way through account setup would leave an orphaned `user` row and a half-provisioned account. There's a test asserting the workspace's member count is unchanged after a rejected provision.

## Two smaller decisions

`requires_email_confirmation` is set on the new account. The address came from the administrator, not its holder, so it is unverified by definition — and §5.5 later depends on that address being the member's, which makes confirming it load-bearing rather than tidy.

The `team_account_activation` template states what the workspace can and cannot do:

> - This account belongs to {{team_name}}. They pay for it and can close it.
> - {{team_name}} **cannot** see your files, apps or data.
> - {{team_name}} **can** reset your password. You will be emailed each time that happens, and every reset is recorded where you can see it.

That last line is required, not padding. Stating only that admins cannot see the contents would describe a guarantee the system does not make.

The equivalent disclosure on the set-password *screen* is phase 6, since that's GUI.

---

## Verification

```
 Test Files  1 passed (1)
      Tests  21 passed (21)

$ PUTER_TEST_DB_ENGINE=postgres npx vitest run --config src/backend/vitest.config.ts src/backend/services/team/
 Test Files  1 passed (1)
      Tests  21 passed (21)

$ npm run test:backend
 Test Files  252 passed | 24 skipped (276)
      Tests  6751 passed | 26 skipped (6777)      # +8, no regressions

$ npm run typecheck
```

Eight tests added: the account is created password-less and owned by the workspace, it gets its filesystem tree (`trash_uuid` set, which is also `generateDefaultFsentries`' own idempotency guard), the activation link is returned and a recovery token stored, a taken username writes nothing, suggestions are verified free and never equal the taken name, a non-master is refused, re-issuing rotates the token, and an activated account cannot be re-sent.

## Not here

| Deferred | To |
| --- | --- |
| `user.free_storage` stamped from a team policy | **deleted, not deferred** — a seat is an ordinary account on the common tier, so its ceiling is `config.storage_capacity` like any other. See `TEAMS-BILLING-SPLIT.md` |
| The billing event that starts the per-account charge | phase 3, PUT-1712 |
| Audit row for `provision` | PUT-1708 |
| Disclosure on the set-password screen | phase 6 |

---

Closes PUT-1705.

### DIFF
--- a/src/backend/controllers/auth/AuthController.ts
+++ b/src/backend/controllers/auth/AuthController.ts
@@ -81,8 +81,8 @@ import {
 } from '../../services/permission/appDataScopes.js';
 import { PuterController } from '../types.js';
 
-const USERNAME_REGEX = /^\w{1,}$/;
-const USERNAME_MAX_LENGTH = 45;
+export const USERNAME_REGEX = /^\w{1,}$/;
+export const USERNAME_MAX_LENGTH = 45;
 const FINGERPRINT_MAX_LENGTH = 128;
 // One consent prompt covers a handful of scopes at most. The cap keeps a
 // crafted request from turning a single grant call into a bulk write.
@@ -194,7 +194,7 @@ const cardAlreadyVerified = (user: {
     requires_card_verification?: boolean;
 }): boolean =>
     Boolean(user.card_fingerprint) && !user.requires_card_verification;
-const RESERVED_USERNAMES = new Set([
+export const RESERVED_USERNAMES = new Set([
     'admin',
     'administrator',
     'root',
@@ -4463,7 +4463,15 @@ export class AuthController extends PuterController {
     ): Promise<UserRow | null> {
         const existing = await this.stores.user.findEmailOwner(email, opts);
         if (!existing) return null;
-        if (existing.email_confirmed || existing.password !== null) {
+        // A provisioned account looks exactly like a claimable placeholder --
+        // no password, unconfirmed -- but claiming it hands a stranger that
+        // workspace's membership.
+        const orgSeat = await this.stores.team.getOrgSeat(existing.id);
+        if (
+            existing.email_confirmed ||
+            existing.password !== null ||
+            orgSeat !== null
+        ) {
             throw new HttpError(
                 400,
                 'This email already exists in our database. Please use another one.',
--- a/src/backend/clients/email/templates.ts
+++ b/src/backend/clients/email/templates.ts
@@ -347,6 +347,24 @@ support@puter.com immediately.
 <p>Puter</p>
         `,
     },
+    team_account_created: {
+        subject: 'Your {{team_name}} account on Puter',
+        html: `
+<p>Hi there,</p>
+<p>{{team_name}} has created a Puter account for you: <b>{{username}}</b>.
+They will send you a temporary password separately; you will be asked to
+choose your own the first time you sign in.</p>
+<p>What this means:</p>
+<ul>
+<li>This account belongs to {{team_name}}. They pay for it and can close it.</li>
+<li>{{team_name}} <b>cannot</b> see your files, apps or data.</li>
+<li>{{team_name}} <b>can</b> reset your password, which would let them sign in
+as you. Choose your own password promptly.</li>
+</ul>
+<p>Sincerely,</p>
+<p>Puter</p>
+        `,
+    },
     enabled_2fa: {
         subject: '2FA Enabled on your Account',
         html: `
--- a/src/backend/services/team/TeamService.test.ts
+++ b/src/backend/services/team/TeamService.test.ts
@@ -28,7 +28,7 @@ describe('TeamService', () => {
     let owner: { id: number };
 
     const makeUser = async (): Promise<{ id: number; username: string }> => {
-        const username = `svc-${Math.random().toString(36).slice(2, 10)}`;
+        const username = `svc_${Math.random().toString(36).slice(2, 10)}`;
         const created = (await server.stores.user.create({
             username,
             uuid: uuidv4(),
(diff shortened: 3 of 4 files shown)


src/backend/controllers/auth/AuthController.ts
```diff
@@ -4463,7 +4463,18 @@ export class AuthController extends PuterController {
     ): Promise<UserRow | null> {
         const existing = await this.stores.user.findEmailOwner(email, opts);
         if (!existing) return null;
-        if (existing.email_confirmed || existing.password !== null) {
+        // A provisioned account looks exactly like a claimable placeholder --
+        // no password, unconfirmed -- but claiming it hands a stranger that
+        // workspace's membership.
+        const orgOwned = await this.clients.db.read(
```
Salazareo: might be jumping the gun if its already in a PR up the stack, but exposing this in the Actor object might be good

Salazareo (approved):
one comment there

jfcastro92:
## Local validation

Engine: mysql 8 · `teams_enabled: true` · `max_seats_per_workspace: 3`

**Provisioning three seats, then hitting the cap**

```
seat1 [HTTP 200] {"username": "acmeseat1", "temporary_password": "<redacted len=16>"}
seat2 [HTTP 200] {"username": "acmeseat2", "temporary_password": "<redacted len=16>"}
seat3 [HTTP 200] {"username": "acmeseat3", "temporary_password": "<redacted len=16>"}
seat4 [HTTP 409] {"error": "This workspace is limited to 3 seats",
                  "code": "seat_limit_reached", "limit": 3}
```

**The rows that result** — seats are `org_owned=1`, the owner is not:

```
username     email_confirmed  req_email  req_pw  suspended  org_owned
tmowner            1              0        0       NULL         0
acmeseat1          0              1        1       NULL         1
acmeseat2          0              1        1       NULL         1
acmeseat3          0              1        1       NULL         1
```

**Provisioning mail actually delivered** (mailhog, real SMTP — not a spy), carrying the
workspace's current name:

```
to=['acmeseat3@example.test'] subj='Your Acme Renamed account on Puter'
to=['acmeseat2@example.test'] subj='Your Acme Renamed account on Puter'
to=['acmeseat1@example.test'] subj='Your Acme Renamed account on Puter'
```

Body confirms no credential is included, as intended — the password goes out of band.

**Credential reissue returns a fresh one-time password**

```
POST /teams/:uid/members/acmeseat1/activation -> 200, temporary_password len=16
login as acmeseat1 with it                    -> 200, session token issued
```

Provisioning, the seat cap, `org_owned`, and the mail path all behave as described.

---

⚠ One follow-up, tracked in PUT-1745 rather than here: `reissueCredential`'s
`requires_password_change` guard does not currently become effective, so the
"already activated" 409 is not reachable and reissue is not rate-bounded by
activation state. Detail and suggested Done-when changes are on that ticket. It
is not reachable in any deployment — `teams_enabled` is `false` by default.

src/backend/controllers/auth/AuthController.ts
jfcastro92: Good catch on the raw query — fixed, though not via the Actor.

Actor doesn't quite fit this call site: `/signup` is unauthenticated (no `requireUserActor`, just `guiOriginOnly` + captcha), so there's no actor yet. And the thing being tested isn't the caller — it's `existing`, the row that owns the email being claimed. Even with an actor present, it'd be the wrong subject.

But the raw SQL in a controller was wrong regardless, and there was already a store method for it, so it's now:

```ts
const orgSeat = await this.stores.team.getOrgSeat(existing.id);
if (existing.email_confirmed || existing.password !== null || orgSeat !== null) {
```

That's strictly better than what I had — `getOrgSeat` also scopes to `g.kind = 'team'`, which my query didn't, so an `org_owned` row in a non-team group no longer blocks a claim. Same behaviour today since only provisioning sets that flag, but it's the correct scope. Full suite green (7521).

On exposing it on the Actor more generally — I think you're right that it's wanted, just for the "is the *caller* a seat" checks rather than this one. That lands in PUT-1745, which owns the enforcement gate; I've noted it there so it gets designed with the gate instead of bolted on separately.

PR #3719, opened 2026-09-02
fix: classify Replicate prediction failures instead of surfacing them as 500s

404oops:
Closes PUT-1608.

## What

`await this.#client.run(...)` in the Replicate image provider had no error handling. A prediction that ran and ended `failed` is thrown by the Replicate SDK as a plain `Error("Prediction failed: …")` with no HTTP status, so the driver-boundary `translateProviderError` could not classify it. Every such failure became an unhandled 500, a `critical` alarm, and a page. This was the largest single alarmed-5xx bucket on the fleet, and most occurrences were the model's content filter refusing the user's prompt.

This PR wraps the call and classifies the failure in the provider:

- **Content-filter refusals** (NSFW, "flagged as sensitive", E005, "content policy") → **400** with `code: bad_request`, `errorCode: moderation_flagged`. Same shape the chat driver already uses for moderation refusals, so clients key on one code across both.
- **Anything else** (model runtime crashes, upstream 502 pages) → **502** with `code: upstream_failed`. The alarm gate skips 5xx carrying an `upstream_*` code, so it no longer pages.
- **Status-bearing rejections** (the SDK's `ApiError`) are rethrown untouched so the boundary translator keeps mapping them to `upstream_rate_limited` / `upstream_auth_failed` / etc. as it does today.
- Upstream messages are **stripped of markup and capped at 300 chars** before they go anywhere, so an Azure Front Door HTML page can never ride into a response body or an alarm signature again.
- Both new errors carry `fields: { provider: 'replicate' }` and the original error as `cause`, matching the BytePlus/Together providers.

The txt2img reference gains an **Errors** section documenting `moderation_flagged`, `upstream_failed`, and `insufficient_funds`, with the exact body shape puter.js hands to callers.

## Reviewer notes

- **502 vs 400 for upstream faults.** The ticket asks for 502, and BytePlus video already uses `status >= 500 ? 502 : 400`. The boundary translator, chat driver, and ElevenLabs expose upstream 5xx as a 400 `upstream_provider_unavailable` instead. Going with the ticket here; either is safe for paging.
- Classification lives in the provider rather than the controller translator because it depends on Replicate-specific message wording.
- The content-filter regex is deliberately conservative (no bare `safety`, which would match parameter-validation errors like `disable_safety_checker`).

## Verification

- Five regression tests added to the provider suite: NSFW and E005 refusals → 400 `moderation_flagged` with no metering; generic crash → 502 `upstream_failed` with `cause`; HTML page → tags stripped and length bounded; status-bearing error → passed through by identity. All five fail on `main`.
- Ran a throwaway end-to-end test (not committed) that booted the real server via `setupPuterTestEnv` with a Replicate key, mocked the SDK, and POSTed to `/drivers/call` over HTTP while spying on `clients.alarm.create`: the refusal returned 400 `moderation_flagged` with no alarm, the crash returned 502 `upstream_failed` with no alarm, and a control unclassified error still returned 500 and alarmed at `critical`.
- Full `src/backend/drivers/ai-image` tree: 203 passed. `eslint`, `prettier --check`, and `npm run typecheck` clean on the source file.

### DIFF
--- a/src/backend/drivers/ai-image/providers/replicate/ReplicateImageGenerationProvider.test.ts
+++ b/src/backend/drivers/ai-image/providers/replicate/ReplicateImageGenerationProvider.test.ts
@@ -392,3 +392,94 @@ describe('ReplicateImageGenerationProvider.generate go_fast pricing', () => {
         expect(outputMp?.costOverride).toBe(Math.round(1.2 * 1_000_000));
     });
 });
+
+// -- Upstream rejection handling --
+
+describe('ReplicateImageGenerationProvider.generate upstream rejections', () => {
+    const generate = () =>
+        withTestActor(() =>
+            makeProvider().generate({
+                model: 'black-forest-labs/flux-schnell',
+                prompt: 'hi',
+            }),
+        );
+
+    it('maps an NSFW refusal to 400 moderation_flagged without metering', async () => {
+        runMock.mockRejectedValueOnce(
+            new Error(
+                'Prediction failed: Error generating image: NSFW content detected.',
+            ),
+        );
+
+        await expect(generate()).rejects.toMatchObject({
+            statusCode: 400,
+            legacyCode: 'bad_request',
+            code: 'moderation_flagged',
+            message: 'Error generating image: NSFW content detected.',
+            fields: { provider: 'replicate' },
+        });
+        expect(incrementUsageSpy).not.toHaveBeenCalled();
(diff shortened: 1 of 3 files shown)


Salazareo (approved):

Salazareo:
merged this by accident

Salazareo:
but was good imo so

PR #3720, opened 2026-09-02
🏗️ PUT-1765: a seat is created, never adopted

jfcastro92:
> Ninth in the stack: #3704 → #3705 → #3708 → #3709 → #3710 → #3712 → #3713 → #3714 → this.

**Rescoped.** This PR was "register the team plan and stamp per-account storage". That work is deleted rather than moved — a seat is an ordinary account on the common tier and the workspace owner pays for it, so there is no team tier to register, no resolver, no `plan_id`, and no `free_storage` stamping. See `TEAMS-BILLING-SPLIT.md` and PUT-1765.

What survives from it is not billing at all:

## A seat is created, never adopted

`addMember` refuses to turn an account that already has a password into a workspace seat. No service path did this — `provisionAccount` always creates — but the store permitted it, and the design rules out existing accounts joining a workspace.

Provisioning passes the guard because it admits the account *before* setting its temporary password. There is no bypass parameter.

## ⚠ An unactivated seat cannot call the API at all

Both HTTP suites now provision a real seat and authenticate as it, using the same token-minting the harness uses for `POST /login`. That surfaced something worth knowing for anyone adding to those suites: provisioning leaves `requires_email_confirmation` set and `requireVerified` rejects it with 403, so the suites activate the seat first — which is the state a member is actually in when making requests.

An earlier version of this had an `allowExisting` flag so a test could reuse an account. That was removing the rule to suit the test, and it is gone; both suites provision real seats.

## `listMembers` and `getMembership` return `u.uuid`

The billing events (#3733) need it to name an account without a second lookup.

---

## Verification

```
      Tests  all passing

$ npm run test:backend

$ npm run typecheck
```

## What went with the rescope

For the record, since it was reviewed as part of this PR before:

| deleted | |
| --- | --- |
| `TEAM_MEMBER_POLICY` | there is no team tier |
| `planIdForOrgAccount` + `registerSubscriptionResolver` | a seat resolves by the ordinary path |
| `applyPlan` / `setPlan` / `group.plan_id` | nothing reads it |
| `#stampStorage` / `free_storage` writes | a seat's ceiling is `config.storage_capacity`, like any account |

The extension points themselves — `registerPolicy`, `registerSubscriptionResolver` — stay untouched in OSS. They are how prod offers the paid tiers a seat can be upgraded to, exactly as `DefaultUserService` uses them today.

One correction worth keeping regardless of model: `SubscriptionPolicy` is `{ id, monthUsageAllowance, monthlyStorageAllowance }`, and `MeteringService` reads `monthUsageAllowance` **directly** in the enforcement path. Registering under any other name leaves it `undefined` — a silently broken allowance rather than an error.

---

Carries PUT-1705's seat-creation guard (PUT-1705 itself is closed by #3713).
Part of PUT-1765 — this is what survives the billing removal from what was
#3720. PUT-1710 and PUT-1744 are superseded.

### DIFF
--- a/config.template.jsonc
+++ b/config.template.jsonc
@@ -273,7 +273,7 @@
     //     "emailBatchSeconds": 90
     // },
 
-    // ── Teams / workspaces ──────────────────────────────────────────────
+    // ── Teams / teams ──────────────────────────────────────────────
     // Off by default, and off is the state every existing install stays in.
     // The tables ship either way and sit inert; this gates whether `/teams`
     // is registered at all, so with it off the paths do not exist rather than
--- a/src/backend/clients/database/migrations/sqlite/0077_teams.sql
+++ b/src/backend/clients/database/migrations/sqlite/0077_teams.sql
@@ -30,10 +30,10 @@ CREATE UNIQUE INDEX IF NOT EXISTS `idx_group_handle`
 -- mysql and postgres already index this column; sqlite does not.
 CREATE INDEX IF NOT EXISTS `idx_group_owner` ON `group` (`owner_user_id`);
 
--- 1 = workspace-created, 0 = the workspace owner. Decides who pays, not who may read.
+-- 1 = team-created, 0 = the team owner. Decides who pays, not who may read.
 ALTER TABLE `jct_user_group` ADD COLUMN `org_owned` INTEGER DEFAULT NULL;
 
--- Covers "list this workspace's members"; the unique pair index lands in 0072.
+-- Covers "list this team's members"; the unique pair index lands in 0072.
 CREATE INDEX IF NOT EXISTS `idx_jct_user_group_group`
     ON `jct_user_group` (`group_id`, `user_id`);
 
--- a/src/backend/clients/database/migrations/sqlite/0079_team-audit-and-group-shares.sql
+++ b/src/backend/clients/database/migrations/sqlite/0079_team-audit-and-group-shares.sql
@@ -37,7 +37,7 @@ CREATE TABLE IF NOT EXISTS `audit_team_membership` (
     FOREIGN KEY("actor_user_id") REFERENCES "user"  ("id") ON DELETE SET NULL ON UPDATE CASCADE
 );
 
--- The admin's view: one workspace, newest first.
+-- The admin's view: one team, newest first.
 CREATE INDEX IF NOT EXISTS `idx_audit_team_membership_group`
     ON `audit_team_membership` (`group_id_keep`, `id`);
 
--- a/src/backend/controllers/auth/AuthController.test.ts
+++ b/src/backend/controllers/auth/AuthController.test.ts
@@ -2108,7 +2108,7 @@ describe('AuthController grant flows', () => {
                 makeReq(
                     {
                         app_uid: app.uid,
-                        permission: `${namespace}:workspace:abc`,
+                        permission: `${namespace}:team:abc`,
                     },
                     { actor: issuerActor },
                 ),
@@ -5319,7 +5319,7 @@ describe('AuthController.handleCheckPermissions + handleListPermissions', () => 
         const permission = kvSharePermission(
             owner.uuid as string,
             'os-global',
-            'workspace:abc:',
+            'team:abc:',
         );
         await inCtx(ownerActor, () =>
             server.services.permission.grantUserUserPermission(
--- a/src/backend/controllers/auth/AuthController.ts
+++ b/src/backend/controllers/auth/AuthController.ts
@@ -4480,7 +4480,7 @@ export class AuthController extends PuterController {
         if (!existing) return null;
         // A provisioned account looks exactly like a claimable placeholder --
         // no password, unconfirmed -- but claiming it hands a stranger that
-        // workspace's membership.
+        // team's membership.
         const orgSeat = await this.stores.team.getOrgSeat(existing.id);
         if (
             existing.email_confirmed ||
--- a/src/backend/controllers/team/TeamController.http.test.ts
+++ b/src/backend/controllers/team/TeamController.http.test.ts
@@ -49,8 +49,8 @@ describe('team endpoints over HTTP', () => {
             ...(body === undefined ? {} : { body: JSON.stringify(body) }),
         });
 
-    /** A workspace owned by `user`, with one provisioned member. */
-    const makeWorkspace = async () => {
+    /** A team owned by `user`, with one provisioned member. */
+    const makeTeam = async () => {
         const res = await call('POST', '/teams', env.users.user.token, {
             name: 'Acme',
             handle: randomHandle(),
@@ -73,7 +73,7 @@ describe('team endpoints over HTTP', () => {
 
     // -- the owner-account gate --------------------------------------
 
-    it('creates a workspace and reports the caller as its owner', async () => {
+    it('creates a team and reports the caller as its owner', async () => {
         const res = await call('POST', '/teams', env.users.user.token, {
             name: 'Acme Design',
             handle: randomHandle(),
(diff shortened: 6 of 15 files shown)


src/backend/data/subPolicies/teamMemberPolicy.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    // The name metering enforces on; anything else leaves it undefined.
+    monthUsageAllowance: toMicroCents(10),
+    // Declarative here; `free_storage` is what FSService actually compares.
+    monthlyStorageAllowance: 10 * 1024 * 1024 * 1024,
+
+    // `toMicroCents` takes DOLLARS. Read by the payment integration.
+    pricePerAccount: toMicroCents(5),
+    // Well below `pricePerAccount`, or disabling would cost more than not.
+    storagePricePerGiBMonth: toMicroCents(0.02),
```
Salazareo: what does this mean?

Salazareo (approved):
I think we'll have to change those teamMemberPolicies and fold them into payment logic instead of here, but does seem ok for MVP

jfcastro92:
## Local validation

Engine: mysql 8 · `teams_enabled: true` · seat cap raised so the cap did not mask these

**Provisioning refuses to adopt an existing account**

```
POST /teams/<uid>/members {"username":"tmoutsider","email":"newaddr@example.test"}
  -> 409 {"error":"That username is taken","code":"username_already_in_use",
          "suggestions":["tmoutsider1","tmoutsider2","tmoutsider3"]}

POST /teams/<uid>/members {"username":"brandnewseat","email":"tmoutsider@example.test"}
  -> 409 {"error":"That email is already in use","code":"email_already_in_use"}

POST /teams/<uid>/members {"username":"acmeseat4","email":"acmeseat4@example.test"}
  -> 200 {"username":"acmeseat4","temporary_password":"<redacted>"}   <- control
```

**The signup-claim guard, isolated.** A provisioned seat always carries a password,
so `existing.password !== null` alone would refuse a claim and the `org_owned`
clause would never be reached. To show the clause is load-bearing, the same row was
tested twice with only `org_owned` differing:

```
password=NULL, email_confirmed=0, org_owned=1
  POST /signup {"username":"stranger2","email":"acmeseat3@example.test"}
  -> 400 {"error":"This email already exists in our database. Please use another one."}

password=NULL, email_confirmed=0, org_owned=0        <- only this changed
  POST /signup {"username":"stranger3","email":"acmeseat3@example.test"}
  -> 200 {"proceed":true,"next_step":"complete","token":"eyJhbGciOi..."}
```

The negative control also demonstrates the hazard the guard exists to stop. With
`org_owned=0` the claim succeeded and, inspecting the row afterwards:

```
id  username    email                    org_owned  team_uid
91  stranger3   acmeseat3@example.test   0          13341565-5223-4c0a-9d28-ca71c9da6682
```

The account was renamed and re-credentialed by a stranger **and stayed inside the
workspace** — exactly what the code comment predicts ("claiming it hands a stranger
that workspace's membership"). Row restored to `acmeseat3` / `org_owned=1` afterwards.

Note the guard now goes through `stores.team.getOrgSeat()` rather than a raw query
in the controller (review follow-up), which additionally scopes it to
`g.kind = 'team'`.

jfcastro92:
## Correction to the local-validation comment above

My previous comment attributed the wrong code to this PR. Correcting it rather than
leaving it to mislead a reviewer.

**What I actually exercised there belongs to #3713**, not here:

- the `username_already_in_use` / `email_already_in_use` refusals come from
  `provisionAccount` reusing AuthController's pre-existing signup validation — those
  legacy codes predate this stack entirely (`HttpError.ts:33/35`)
- the `org_owned` signup-claim A/B tests `#resolveSignupEmailClaim` in
  `AuthController`, which is #3713's change

Both are real and both passed, but neither is this PR's code. The evidence is
correctly filed on #3713.

**What this PR actually changes:**

1. `TeamStore.addMember` refuses to mark a password-bearing account as a seat:

```ts
// A seat is created, never adopted; a password means it predates us.
if (opts.orgOwned) {
    const rows = await this.clients.db.read(
        'SELECT 1 AS hit FROM `user` WHERE `id` = ? AND `password` IS NOT NULL LIMIT 1', [userId]);
    if (rows.length > 0) throw new Error('cannot make an existing account a team seat');
}
```

2. `TeamMemberRow` gains `uuid`, and both `listMembers` queries select `u.uuid`.

**The guard is not reachable through the HTTP surface today.** Both call sites are
safe by construction:

| call site | passes | reaches the guard? |
| --- | --- | --- |
| `provisionAccount` (`TeamService.ts:515`) | `orgOwned: true` | no — the user is created with `password: null` two statements earlier |
| workspace create (`TeamService.ts:272`) | `orgOwned: false` | no — guard is `orgOwned`-only |

So it is defense-in-depth against a *future* caller that tries to admit an existing
account as a seat, which is exactly what the commit message says. That is a
reasonable thing to add — I am not arguing against it.

**But it has no direct test.** Grepping the tree, `cannot make an existing account a
team seat` appears only in `TeamStore.ts` — nothing asserts it. The test changes in
this PR adjust existing suites to activate seats so they can authenticate; none
covers the new branch. Since the guard is unreachable from HTTP, a unit test on
`addMember` is the only thing that can cover it, and without one a later refactor
could drop the branch with every suite still green.

Happy to add that test here if you'd like it in this PR — it is a few lines against
`TeamStore.addMember` with a password-bearing user.

The `uuid` addition is internal: `TeamController.listMembers` still returns only
`username` / `org_owned` / `created_at`, verified live, so no response shape changed.

jfcastro92:
## Added coverage for the `addMember` guard

Following up on the correction above — the guard is unreachable from HTTP, so a unit
test on the store is the only thing that can cover it. Two added to
`TeamStore.test.ts`:

```ts
it('refuses to make an account that already has a password a seat', async () => {
    // Provisioning creates the row password-less and sets one afterwards,
    // so a password is what distinguishes an account that predates us.
    const existing = await makeUser();
    await server.stores.user.update(existing.id, { password: 'hashed' });

    await expect(
        store.addMember(team.uid, existing.id, { orgOwned: true }),
    ).rejects.toThrow(/cannot make an existing account a team seat/);
    await expect(store.isMember(team.uid, existing.id)).resolves.toBe(false);
});

it('still admits an account that has a password as the owner', async () => {
    // The guard is about seats: the owner joins their own workspace with a
    // password already set, and must not be caught by it.
    const payer = await makeUser();
    await server.stores.user.update(payer.id, { password: 'hashed' });

    await expect(
        store.addMember(team.uid, payer.id, { orgOwned: false }),
    ).resolves.toBe(true);
});
```

The existing `makeUser` helper creates users with `password: null`, which is exactly
why no prior test reached this branch — the second test is the control that keeps the
guard scoped to `orgOwned: true` rather than to "has a password".

**Falsified rather than just run.** Deleting the throw from `TeamStore.addMember`:

```
guard removed? 0
 × refuses to make an account that already has a password a seat
Tests  1 failed | 29 passed (30)
```

Exactly one failure, and the owner control still passes — so the test fails for the
right reason and is not incidentally coupled to the guard. Guard restored, suite back
to 30 passed.

Verification on the rebased stack: every commit typechecks individually
(`git rebase --exec`), full backend suite **7523 passed | 26 skipped**.

src/backend/data/subPolicies/teamMemberPolicy.ts
jfcastro92: Good question, and the honest answer is that it meant less than it looked like it meant — which is why the file is gone now.

The line you flagged was:

```ts
// Declarative here; `free_storage` is what FSService actually compares.
monthlyStorageAllowance: 10 * 1024 * 1024 * 1024,
```

There were three separate representations of "storage" and only one of them enforced anything:

| | what it was | did it gate anything? |
| --- | --- | --- |
| `monthlyStorageAllowance` on the policy | declared 10 GiB | **no** — nothing in `src/backend` read it |
| `user.free_storage` | stamped per seat at provisioning | **yes** — the real ceiling |
| held-bytes report | `SUM(size)`, billed when disabled | reporting only |

So "10 GiB included" was enforced by a column stamped from a policy field nobody read, and nothing checked the two agreed. That is what the comment was admitting, and it is not a good thing for a comment to have to admit.

And you were right in the review body — this all moved into payment logic. `teamMemberPolicy.ts` is deleted, along with `TeamBillingService`, `applyPlan`, `setPlan`, the `free_storage` stamping, the held-bytes job and `group.plan_id`. A seat is now an ordinary account on the common tier; OSS registers no team policy at all, exactly like `DefaultUserService` registers only the free and self-hosted ones and lets prod register the real tiers.

What OSS keeps is the events prod needs to do the billing: `team.account.created` / `disabled` / `enabled` / `deleted`, `team.deleted`, and `metering.credit-state`. Those are #3733 and #3724 in this stack, and `updateAddonCredit` was already per-account so no new extension points were needed.

Salazareo (approved):

jfcastro92:

jfcastro92:

| Before | After |
| --- | --- |

### Scoped, and checked

### One thing not renamed

jfcastro92:
> I think we'll have to change those teamMemberPolicies and fold them into payment logic instead of here, but does seem ok for MVP

That already happened — and your inline question on `teamMemberPolicy.ts` is now unanswerable in the good way: **the file is gone.**

Phase 3 was reworked after this review so no pricing lives in the OSS tree. `teamMemberPolicy.ts` was deleted along with `TeamBillingService.ts` and `TeamStore.getStripeCustomerId`, which took the last payment-vendor method out of the OSS teams path. What is left here is event emission only — `team.account.created/disabled/enabled/deleted` and `team.deleted` — with the pricing and the Stripe calls in the prod extension, HeyPuter/heyputer#1172.

So `pricePerAccount` and `storagePricePerGiBMonth`, which are what you were reading when you asked "what does this mean?", are no longer declared in this repo at all. They live with the payment logic, which is where you said they belonged.

⚠ Since the branch was rewritten, this PR is no longer the change you reviewed. It is now **PUT-1765: a seat is created, never adopted** — worth a fresh look rather than trusting the earlier approval.

Also note the whole stack was renamed workspace → team since your review, including the error codes and config keys. Details in the comment above.

PR #3721, opened 2026-09-02
🏗️ PUT-1711: invalidate billing caches on every membership event

jfcastro92:
> Tenth in the stack: #3704 → #3705 → #3708 → #3709 → #3710 → #3712 → #3713 → #3714 → #3720 → this. Review bottom-up.

**PUT-1711.** Small — 16 lines of implementation — but it decides how fast a billing change takes effect, so it earns the scrutiny.

## One call, not two

`getActorSubscription` caches the resolved policy per user uuid, in memory, per process. `invalidateActorSubscription(uuid)` drops the local copy and emits `outer.pubsub.metering.subscription-changed`, which sibling nodes and peer clusters consume — **and the emitting node consumes its own event**, whose handler drops the credit verdict as well:

```js
this.#dropCachedSubscription(data.userUuid);
// The allowance is half of what "has budget left" is computed
// from, so a plan change invalidates that answer too.
this.#dropCachedCredits(data.userUuid);
```

So a separate `invalidateActorCredits` call is redundant. Worth spelling out because the method *body* only touches the subscription cache — the credit drop happens via the event, which is easy to miss reading top-down.

## Every event that changes what an account resolves to

| Event | Why |
| --- | --- |
| provisioning | the seat starts resolving to the workspace's plan |
| disable / re-enable | listed by the ticket (see the note below) |
| `plan_id` change | landed in #3720; sweeps every seat |
| **workspace deletion** | **not in the ticket, and the one that matters most** |

`planIdForOrgAccount` filters `g.deleted_at IS NULL`, so the instant `softDelete` runs every seat resolves to the free tier. Without invalidation each node — and every peer cluster — kept serving the deleted workspace's plan for the full `SUBSCRIPTION_CACHE_MS` window, which is precisely the TTL wait this ticket exists to remove. Deletion now also clears `free_storage`, which otherwise kept the team ceiling permanently after the workspace was gone.

## ⚠ The uuid comes from rows already in hand, deliberately

The first cut used a small `#invalidateBillingFor(userId)` helper that looked the user up to get its uuid. That is a trap, and it is worth understanding before anyone reintroduces it:

- By the time it runs the user cache is **guaranteed empty** — `stores.user.update` refreshed it, then `invalidateById` broadcast a delete.
- So the read always misses and goes to `db.tryHardRead`, which *"prefer[s] replica when it returns rows"*.
- The `user` row always exists on the replica, so under lag it returns the **pre-`UPDATE`** row.
- `getByProperty` then writes that back to Redis for `CACHE_TTL_SECONDS` — **15 minutes**.

Concretely: a master re-enables an account, the `UPDATE` clears `suspended`, and the lookup re-caches `suspended: 1`. Any route that reads the user cache **without** forcing then rejects that member as suspended for up to 15 minutes. `userProtected` forces, so it lets them through — making the failure route-dependent and horrible to diagnose.

`enableMember` already does a forced read 25 lines earlier, with the comment *"a cached row predates this"*. The lookup was exactly the read that comment warns about.

Both call sites already have the uuid — `requireOrgAccount` returns a row carrying it, `enableMember` has it from the forced read — so the helper is gone. That also removes an `await`ed failure point sitting *after* the audit row and the suspension had already committed, where a Redis blip would 500 an operation that fully succeeded and invite a retry that duplicates the audit row.

## Testing

Four tests, one per call site. The provisioning one **asserts the call rather than the cache**, because the seat does not exist before `provisionAccount` runs and so cannot be pre-warmed — an earlier version warmed the caches *after* provisioning returned, which re-filled whatever provisioning had dropped and silently tested nothing.

The other three warm **both** caches first (`getActorSubscription` fills the subscription one, `hasAnyUsageCached` fills the credit one) and assert both are empty afterwards.

<details>

Removing each call site individually, one at a time:

```
remove provisioning -> 1 failed | 50 passed
remove disable      -> 1 failed | 50 passed
remove enable       -> 1 failed | 50 passed
remove delete       -> 1 failed | 50 passed
```

Each site fails exactly one test. Removing all four at once and seeing "3 failed" would *not* have established this — that was how the untested provisioning site initially hid.
</details>

---

## Verification

```
      Tests  51 passed (51)

$ PUTER_TEST_DB_ENGINE=postgres … src/backend/services/team/
      Tests  51 passed (51)

$ npm run test:backend
 Test Files  254 passed | 24 skipped (278)
      Tests  6868 passed | 26 skipped (6894)

$ npm run typecheck
```

## One judgement call left open

**Disable and re-enable do not actually change plan resolution.** `planIdForOrgAccount` keys only on `jct_user_group.org_owned` and `group.plan_id` — never on `suspended` — and the credit verdict is allowance plus usage. What makes a disable visible on another node is the *user-row* cache broadcast plus session revocation, not this.

So those two invalidations are semantic no-ops today. They are listed explicitly by PUT-1711, they cost one in-process map delete plus a pubsub emit, and they are correct insurance if a resolver ever becomes suspension-sensitive. Kept on that basis — happy to drop them if you would rather the call sites mean something.

---

Closes PUT-1711.


Salazareo (approved):
lgtm

jfcastro92:
Superseded. Billing moves to a prod extension, so there is no team plan for OSS to invalidate — a seat resolves to the common tier by the ordinary path, and provisioning, disable and re-enable no longer change what it resolves to.

`invalidateActorSubscription` / `invalidateActorCredits` stay in OSS untouched; prod calls them when it moves one seat to a paid tier or tops one up.

Two findings from this work are recorded on PUT-1711 because they outlive it: `invalidateActorSubscription` drops the credit verdict too (via the `subscription-changed` handler, which the emitting node also consumes), and a uuid must never be resolved with a fresh lookup on an invalidation path — `tryHardRead` prefers the replica and `getByProperty` re-caches for 15 minutes, so the lookup re-seeds the shared user cache with the pre-update row.

See TEAMS-BILLING-SPLIT.md and PUT-1765.

PR #3722, opened 2026-09-02
🏗️ PUT-1712 + PUT-1747: emit billing events and report held bytes

jfcastro92:
> Eleventh in the stack: #3704 → #3705 → #3708 → #3709 → #3710 → #3712 → #3713 → #3714 → #3720 → #3721 → this. Review bottom-up.

Covers **PUT-1712** (on-demand billing events) and **PUT-1747** (the recurring held-bytes report). Grouped because they are the two halves of one signal: what a workspace owes at each transition, and what it owes in between.

**Nothing here computes money.** The price lives in `TEAM_MEMBER_POLICY` and the charge lives outside this repo — the same place that cancels Stripe subscriptions off `user.delete`. That event is the precedent for the shape, and this follows it: emit a typed event, let an extension do the external side effect.

## The events

Each carries the master account's `stripe_customer_id` (it is the payer), the workspace uid, and the seat.

| Event | Billing meaning |
| --- | --- |
| `team.account.created` | Start the per-seat charge |
| `team.account.disabled` | Stop it; start billing the bytes it holds |
| `team.account.enabled` | Resume the seat charge; stop the byte charge |
| `team.account.deleted` | Stop both |
| `team.deleted` | Stop every seat charge — byte charges continue |
| `team.held-bytes.report` | The recurring measurement, per workspace |

Disable and re-enable carry `held_bytes` at the moment of transition, so the payment side has a figure without querying back.

## Deleting a workspace emits per seat, not in bulk

The ticket describes workspace deletion as one row. It is emitted as one `team.account.disabled` per seat **plus** a `team.deleted` marker, because the byte charge that follows is per account — a bulk event would carry an unbounded account list and still need unpacking into per-account charges on the other side. `deleteWorkspace` already disables each seat, so the per-seat events are what actually happened.

⚠ Worth stating plainly, since PUT-1740's confirmation dialog depends on it: **deleting a workspace is not a way to stop paying.** The seats stop, the bytes do not — the accounts still exist and still hold data. Deleting the accounts is what stops the byte charge.

## ⚠ `team.account.deleted` has to be captured before the delete

`jct_user_group.user_id` is `ON DELETE CASCADE` (`0015_group.sql`). By the time a listener on `user.delete` runs, the membership row is gone and **nothing can say which workspace paid for the account** — the payload is the only surviving evidence.

So `cascadeDelete` captures the seat identity before teardown and emits after, exactly as it already does for `uuid` / `stripe_customer_id` and for the same reason. There is a test asserting the membership is unreadable afterwards, so the constraint that forces this design is pinned rather than assumed.

## ⚠ `stripe_customer_id` does not exist on sqlite

It ships in `mysql_mig_1` and `postgres_mig_1` but no sqlite migration — `cascadeDelete` already wraps its read in a `try`/`catch` for this. `getStripeCustomerId` does the same and degrades to `null`, so a dev or self-hosted install on sqlite emits usable events rather than throwing inside a lifecycle operation. Production is mysql/postgres, where the value is real; the postgres suite covers the branch that actually returns one.

## The recurring report

`TeamBillingService` — separate from `TeamService` because it is scheduled rather than request-driven, the same split as `ShareNotificationService` from `ShareService`.

- **Off the hot path.** A periodic read, not a listener on `fs.write.file`. A disabled account cannot write, so its total only changes when the account is deleted, which emits its own event.
- **Period-aligned.** `observed_at` is the window start, floored to the interval — not the sample instant. Billing invoices a window; an arbitrary timestamp is not one, and flooring is also what makes two runs byte-identical.
- **One node per window.** The Redis lock key carries the period (`team:held-bytes:lock:<periodStart>`), so a second node in the same window is refused rather than racing. Each node's timer has its own phase from its own boot, so a duration-based TTL would not have deduplicated them.
- **No counter.** The authoritative `SUM(size)` is the source, as the ticket specifies.
- **Disabled seats only**, and `org_owned = 1` only.

Cadence is `team_held_bytes_report_hours`, default 24. The sweep is bounded at 50 000 accounts per run and **logs when it truncates** — a short report reads as a smaller bill, so it must never be silent.

## The `org_owned` filter is load-bearing

A master account that is itself suspended by Puter would otherwise appear in its own workspace's held-bytes report, moving its personal storage onto the workspace's bill. The master pays for its own storage under its own plan. There is a test for exactly that shape — a suspended master with ~1 MB against a disabled seat with 64 bytes, asserting only the seat is reported.

I did not write that test until falsification told me to; see below.

---

## Verification

```
 Test Files  5 passed (5)
      Tests  111 passed (111)

      Tests  92 passed (92)

$ npm run test:backend
 Test Files  269 passed | 24 skipped (293)
      Tests  7117 passed | 26 skipped (7143)

$ npm run typecheck
```

Postgres matters here beyond the usual: it is the engine where `stripe_customer_id` exists and where `suspended` is a real `boolean` rather than `tinyint(1)`, so the report's filter is built with `db.booleanValue(true)` rather than a literal `1`.

<details>
<summary>Per-site falsification — every emit and filter, one at a time</summary>

Each site disabled individually, with the suite re-run each time:

```
baseline                     0 failed | 13 passed
account.created emit         1 failed | 12 passed
disable emit                 1 failed | 12 passed
enable emit                  1 failed | 12 passed
delete: per-seat emit        1 failed | 12 passed
team.deleted emit            1 failed | 12 passed
account.deleted emit         1 failed | 12 passed
pre-delete capture           1 failed | 12 passed
period lock                  1 failed | 12 passed
period flooring              1 failed | 12 passed
suspended filter             1 failed | 12 passed
org_owned filter             1 failed | 12 passed
restored                     0 failed | 13 passed
```

Every site fails **exactly one** test: no site is untested, and no test is over-broad.

Two of these only became true because the first pass was wrong, which is the argument for doing it per-site:

- **Period flooring and the `org_owned` filter both survived their mutation.** Flooring was only incidentally covered by the two-runs-are-identical test, which catches it just by timing; and nothing at all covered `org_owned`, because no test had a suspended master. Both now have a dedicated test.
- **The harness itself produced two false negatives.** Early mutations reported "all passed" while never having applied — a shell-quoting problem, then a wrong target string. The script now refuses to report a result unless it has verified the mutation landed in the file. A falsification that silently no-ops is indistinguishable from a passing test, which makes it worse than not running one.
</details>

## Not here

| Deferred | To |
| --- | --- |
| Notify the master when a member runs out of credit | PUT-1750 |
| Hard delete on request, gated on disable | PUT-1732 |
| The charge itself | outside this repo |

`monthlyStorageAllowance` remains declarative — `user.free_storage` is what `FSService` enforces, as noted in #3721.

---

Closes PUT-1712 and PUT-1747.


Salazareo (changes requested):
might need to talk this over

jfcastro92:
Reopening on the correct base — #3721 closed, so this was stacked on a dead branch. Continued in the PR that supersedes it. Content is also rescoped: the held-bytes report is deleted rather than moved (see TEAMS-BILLING-SPLIT.md).

PR #3724, opened 2026-09-02
🏗️ PUT-1750: announce credit-state transitions so billing can act

jfcastro92:
> Eleventh in the stack: #3704 → … → #3720 → #3733 → this.

**Rescoped.** This PR was "notify the owner when a member runs out of credit", handler and all. The notification moves to a prod extension; what stays in OSS is the announcement it listens for. See `TEAMS-BILLING-SPLIT.md` and PUT-1765.

## The problem it still solves

A seat that exhausts its allowance is refused and **cannot fix it itself** — only the workspace owner can add capacity. Something has to tell the owner, or the member is stuck until they think to complain. OSS's job is to say when it happened; prod's job is to decide that becomes mail.

## Where the emit lives, and why

`MeteringService` emits `metering.credit-state` from `rememberRemainingCredits` — the single point where the credit verdict is computed from allowance, usage and addons. At 90% of the allowance, and again on exhaustion.

`assertActorHasCredits` would be the obvious place and is the wrong one: it is a pure function with no access to clients or services, so it *could* not emit. `rememberRemainingCredits` already has the numbers in hand, so the announcement costs no reads at all.

## ⚠ Transitions only, never per request

A blocked account keeps trying, and every metered call recomputes the verdict. Emitting per computation would announce hundreds of times for one member's retry loop.

The last announced state is tracked per uuid in process, under the same FIFO bound as the credit cache, and dropped in `#dropCachedCredits` — which is the choke point both the local call and the `credits-changed` pubsub handler go through. So **added capacity re-arms the alert** for the rest of the month.

There is a test that drives 50 real increments after exhaustion and asserts nothing further is announced. It uses increments rather than cached reads deliberately: a cache hit never recomputes, so a read loop would pass without exercising the dedup at all.

## What prod does with it

One notice per member, per state, per billing month. The dedup needs a **cross-region** marker — `SystemKVStore` rather than Redis, since a member blocked in three regions must produce one alert — claimed with an atomic `incr` where the caller that reads `1` owns the send. And the key must carry the **state**, or the 90% warning claims the marker and the exhaustion mail never sends.

Prod has everything it needs: `extension.on('metering.credit-state', …)`, `stores.team.getOrgSeat(userId)` to find the owner, and `updateAddonCredit(uuid, n)` for the top-up.

## ⚠ The 402 no longer claims a notification

The earlier version told the member "your workspace administrator has been notified". OSS no longer notifies anyone, so it must not say so — with prod absent it would simply be false. The message is back to the generic `No usage left for request.`, and `CreditMetering` loses the `getActorSubscription` member that existed only to decide this.

---

## Verification

```
      src/backend/services/team/CreditState.test.ts
      Tests  4 passed (4)

$ npm run test:backend

$ npm run typecheck
```

The suite resolves each seat's allowance from its own subscription rather than a constant, because a seat is on the common tier now — there is no team allowance to hard-code.

<details>

```
baseline                       0 failed | 11 passed
metering.credit-state emit     2 failed |  9 passed
credit-state transition dedup  1 failed | 10 passed
near-limit threshold           2 failed |  9 passed
restored                       0 failed | 11 passed
```
</details>

---

Part of PUT-1750; the notification half is prod's.

### DIFF
--- a/src/backend/services/metering/MeteringService.ts
+++ b/src/backend/services/metering/MeteringService.ts
@@ -146,6 +146,9 @@ export class MeteringService extends PuterService {
     static CREDIT_CACHE_MS = 15_000;
     static CREDIT_CACHE_LIMIT = 50_000;
 
+    /** Where "about to run out" starts, as a fraction of the month allowance. */
+    static NEAR_LIMIT_FRACTION = 0.9;
+
     /**
      * How long usage that isn't decided on may sit in memory before it is
      * written, and how many actor buckets are held at once. Egress and
@@ -170,6 +173,12 @@ export class MeteringService extends PuterService {
     private subscriptionCache = new Map<
         string,
         { policy: SubscriptionPolicy; expiresAt: number }
+    >();
+
+    /** Uuid → the last budget state announced, so a retry loop emits once. */
+    private creditAlertState = new Map<
+        string,
+        'ok' | 'near-limit' | 'exhausted'
     >();
 
     /** Uuid → whether the actor had budget left. See CREDIT_CACHE_MS. */
@@ -1276,6 +1285,8 @@ export class MeteringService extends PuterService {
     /** Local-only drop. The announcement path is `invalidateActorCredits`. */
     #dropCachedCredits(userUuid: string): void {
         this.creditCache.delete(userUuid);
+        // Added capacity re-arms the alert: the next exhaustion is news again.
+        this.creditAlertState.delete(userUuid);
     }
 
     async #refreshCredits(actor: Actor): Promise<void> {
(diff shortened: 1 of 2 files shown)


jfcastro92:
## Local validation

Engine: mysql 8 · events captured by a throwaway extension listening on
`metering.credit-state`.

**Making exhaustion reachable.** At the real free-tier allowance
(`toMicroCents(0.5)` = 50,000,000 microcents) against egress at
`toMicroCents(0.12 / GiB)` ≈ 0.0112 microcents/byte, exhausting an account needs
**~4.2 GiB of egress** — and the 100 MiB storage allowance makes staging that
awkward. Since the behaviour under test is the *transition detection*, not the
allowance value, `REGISTERED_USER_FREE.monthUsageAllowance` was temporarily set to
`toMicroCents(0.00002)` (2,000 microcents ≈ 175 KiB of egress) for this run and
reverted afterwards. Disclosing it because it means the thresholds below are
scaled, not production values.

**The `exhausted` transition fires with the documented payload**, on three separate
accounts driven independently:

```
20:09:40.903 state=exhausted  used=2000  allow=2000  uuid=fa0ac04f  (tmowner)
20:10:00.902 state=exhausted  used=2000  allow=2000  uuid=3f3a0293  (tmoutsider)
20:10:40.885 state=exhausted  used=2000  allow=2000  uuid=5f5df797  (credprobe1)
```

Payload shape matches `EventMap`: `user_uuid`, `state`, `allowance_used`,
`month_usage_allowance`.

**The dedup holds — this is the load-bearing property.** Each of those accounts kept
issuing requests well past the point of exhaustion, and each produced exactly **one**
event:

| account | requests driven | credit-state events |
| --- | --- | --- |
| tmowner | 15 | 1 |
| tmoutsider | 100 | 1 |
| credprobe1 | 950 | 1 |

That is the "a blocked actor retries, and every retry lands here — transitions only"
claim, verified against a live metering path rather than a mock. Without the
`creditAlertState` guard these would have been 15, 100 and 950 notifications.

---

⚠ **Not observed live: the `near-limit` transition.** In every run `allowance_used`
landed exactly on the allowance, so the account went from below 90% straight to
exhausted without a reading in the 90–100% band. Egress usage settles in aggregate
rather than per request, so the step size exceeded the 200-microcent window at this
scaled allowance; shrinking the read size to 200 bytes did not change it. The
`near-limit` branch is covered by `CreditState.test.ts`, but it was **not**
reproduced against a live server here, and I would not claim it was.

Worth noting for review: at production numbers the window between 90% and 100% is
5,000,000 microcents wide (~437 MiB of egress), so this is very likely an artefact
of the scaled-down allowance rather than a real gap. It would be reassuring to
confirm that on a deployment where the allowance is not shrunk.

Salazareo (approved):

jfcastro92:
## Review fix: near-limit ignored purchased credits

Confirmed. `#noteCreditState` tested `allowanceUsed >= monthUsageAllowance * NEAR_LIMIT_FRACTION` — the monthly allowance alone — while `remaining` right above it already summed the allowance *and* unspent purchased credits. So someone who bought their way clear of the limit got warned anyway, at 90% of a number that no longer bounded them.

The state now measures against the same capacity `remaining` does:

```ts
const capacity = (monthUsageAllowance || 0) + (addons?.purchasedCredits || 0);
const state = remaining <= 0 ? 'exhausted'
    : remaining <= capacity * (1 - MeteringService.NEAR_LIMIT_FRACTION) ? 'near-limit'
    : 'ok';
```

With no addons this is algebraically the old test — `remaining <= A*0.1` ⟺ `used >= A*0.9` — so the existing near-limit and exhausted tests are unchanged, which is the point: the fix adds a case rather than moving the threshold.

New test: a seat granted purchased credits equal to its allowance, then spending 95% of the allowance, hears nothing. Falsified against the old comparison:

```
× does not warn a member whose purchased credits carry them past the allowance
Tests  1 failed | 4 passed (5)
```

Full suite **7883 passed | 26 skipped**, typecheck clean.

jfcastro92:

jfcastro92:

| Before | After |
| --- | --- |

### Scoped, and checked

### One thing not renamed

PR #3726, opened 2026-09-02
🏗️ PUT-1719 + PUT-1725: two-workspace fixture and group permission grants

jfcastro92:
> Fourteenth in the stack, and the start of phase 4: #3704 → #3705 → #3708 → #3709 → #3710 → #3712 → #3713 → #3714 → #3720 → #3721 → #3722 → #3724 → #3725 → this. Review bottom-up.

Covers **PUT-1719** (two-workspace fixture) and **PUT-1725** (group grants). Together because the fixture has no behaviour of its own — a fixture PR with no consumer can't be judged — and the grant tests are its first real use.

## PUT-1719 — why two workspaces

Every query resolving workspace-scoped data has to take the workspace as a parameter rather than infer it. One that forgets **returns the right rows against a database holding a single workspace**. So a one-workspace fixture doesn't give weaker coverage; it gives false confidence. Two is the smallest shape that can fail.

Concretely, drop `g.uid = ?` from `listMembers` and every assertion still passes with one workspace — the rows returned are exactly that workspace's members, because there are no others. Add a second and the same query hands workspace A's master B's roster.

Provides two workspaces, each with its own master and two activated seats, plus a user in neither.

Two things worth knowing about it:

- **It runs against the real `max_workspaces_per_user: 1`.** Each workspace has its own master, so the cap never needs lifting — unlike the five existing team suites, which pass `max_workspaces_per_user: 100` because they create several workspaces per master. There's a test asserting the cap is still 1, because the moment the fixture needs it raised it stops resembling production.
- **Seats are activated before use.** Provisioning leaves `requires_email_confirmation` set and `requireVerified` rejects those outright, so an unactivated seat can't call anything — and every authorization assertion built on one would be vacuous. There's a test that a seat token actually reaches a route, for the same reason.

## PUT-1725 — the write half only

The read half already worked: `#scanUserGroup` joins `jct_user_group` itself, so a grant resolves for whoever is a member **at scan time**. That's what makes membership changes apply automatically without touching grants.

```
PermissionStore     resolveGroupId, listGroupMemberUuids,
                    upsertUserGroupPerm, deleteUserGroupPerm, auditUserGroupPerm
PermissionService   grantUserGroupPermission, revokeUserGroupPermission
```

Both rewrite the permission **first**, so `fs:/path:mode` collapses to `fs:<uuid>:mode` and a revoke names the same string the grant wrote — skipping it lets a revoke match nothing while reporting success. Both gate on `canManagePermission` and audit into the existing `audit_user_to_group_permissions`.

**The delete is scoped to the issuer.** The PK is `(user_id, group_id, permission)` where `user_id` is the *issuer*, so two issuers can hold the same permission against the same group, and one revoking must not drop the other's.

## Cache invalidation is one event, not N

`bumpCacheGenerations` takes the whole array of member uuids, does the counter increments in parallel, and emits **one** `outer.permission.generationBumped` carrying all of them. Peer regions consume it; `#applyCacheGenerationBump` is documented as never emitting, so a remote bump can't ping-pong.

Worth noting for anyone reading the existing helper: `bumpPermissionCacheForUsernames` does *not* use that batch path — it maps over usernames, looks each up, and calls the single-uid version, so it fires one event per member. This uses the batch call directly. What remains O(N) is Redis `INCR`s, not broadcasts.

Also: there is **no flat-KV equivalent** for group grants (`setFlatUserPerm` is user-to-user only), so unlike the user-to-user path there's no second write to keep consistent and no second invalidation surface. The warm-TTL and primary-vs-replica hazards in that path have no analogue here.

## Where this doesn't reach

`#scanUserGroup` opens with `if (actor.app || actor.accessToken) return;` — so **apps and access tokens do not see group grants**. `#scanUserUser` has the same guard, so this is consistent with user-to-user shares rather than a regression, but it means a file shared with a workspace is invisible to an app acting for a member. Accepted deliberately.

---

## Verification

```
$ npx vitest run --config src/backend/vitest.config.ts src/backend/services/permission/
 Test Files  3 passed (3)
      Tests  110 passed (110)

$ PUTER_TEST_DB_ENGINE=postgres … src/backend/services/permission/GroupGrants.test.ts
      Tests  13 passed (13)

$ npm run test:backend
 Test Files  276 passed | 24 skipped (300)
      Tests  7216 passed | 26 skipped (7242)

$ npm run typecheck
```

<details>

```
baseline                   0 failed | 13 passed
membership join in read    5 failed |  8 passed
grant authority check      1 failed | 12 passed
missing-group 404          1 failed | 12 passed
revoke issuer scoping      1 failed | 12 passed
grant audit                1 failed | 12 passed
revoke rewrite             1 failed | 12 passed
restored                   0 failed | 13 passed
```

**`membership join in read` is the one that justifies the whole fixture.** Replacing the `jct_user_group` join with `1 = 1` — a scan that resolves every group's grants regardless of membership — fails 5 tests. Against a single-workspace fixture it would fail none of them.

`revoke issuer scoping` initially reported `0 failed`, and the test was at fault: it used two *different* permissions, so issuer scoping never decided anything. It now has two issuers holding the same permission against the same group, which is the only shape where the `DELETE`'s issuer clause matters.
</details>

## Not here

`ACLService.setUserGroup` and the `{ team }` share recipient (PUT-1726), the share index and listing union (PUT-1727), member notifications (PUT-1761), and the consistency tests (PUT-1729) — grouped into the next PR, since a recipient without the listing work produces a share that resolves correctly and never appears in "Shared with me".

---

Closes PUT-1719 and PUT-1725.

### DIFF
--- a/src/backend/services/permission/GroupGrants.test.ts
+++ b/src/backend/services/permission/GroupGrants.test.ts
@@ -1,0 +1,362 @@
+/**
+ * Copyright (C) 2024-present Puter Technologies Inc.
+ *
+ * This file is part of Puter.
+ *
+ * Puter is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import type { Actor } from '../../core/actor';
+import {
+    setupTwoTeams,
+    type TwoTeams,
+} from '../../testFixtures/twoTeams.js';
+
+describe('group grants', () => {
+    let fx: TwoTeams;
+
+    /** A plain user actor, which is what a grant is issued as. */
+    const actorFor = async (userId: number): Promise<Actor> => {
+        const user = await fx.env.server.stores.user.getById(userId);
+        return { user } as unknown as Actor;
+    };
+
+    /** Written directly: these tests are about granting, not about writing. */
+    const makeFile = async (ownerId: number) => {
+        const uid = crypto.randomUUID();
+        const name = `f_${uid.slice(0, 8)}.txt`;
+        const owner = await fx.env.server.stores.user.getById(ownerId);
+        const path = `/${owner!.username}/${name}`;
+        await fx.env.server.clients.db.write(
+            'INSERT INTO `fsentries` (`uuid`, `name`, `path`, `user_id`, `is_dir`, `modified`) ' +
+                'VALUES (?, ?, ?, ?, ?, ?)',
+            [
+                uid,
+                name,
+                path,
+                ownerId,
+                // `is_dir` is a real boolean on postgres.
+                fx.env.server.clients.db.booleanValue(false),
+                Math.floor(Date.now() / 1000),
+            ],
+        );
+        return { path, uid };
+    };
+
+    const permissions = () => fx.env.server.services.permission;
+    const store = () => fx.env.server.stores.permission;
+
+    beforeAll(async () => {
+        fx = await setupTwoTeams();
+    }, 180_000);
+
+    afterAll(async () => {
+        await fx?.shutdown();
+    });
+
+    // -- the fixture itself -------------------------------------------
+
+    it('builds two teams, each with its own owner and two seats', async () => {
+        expect(fx.a.uid).not.toBe(fx.b.uid);
+        expect(fx.a.owner.userId).not.toBe(fx.b.owner.userId);
+        expect(fx.a.seats).toHaveLength(2);
+        expect(fx.b.seats).toHaveLength(2);
+
+        // Distinct throughout, or it cannot tell this team from any.
+        const ids = [
+            fx.a.owner.userId,
+            fx.b.owner.userId,
+            ...fx.a.seats.map((s) => s.userId),
+            ...fx.b.seats.map((s) => s.userId),
+            fx.outsider.userId,
+        ];
+        expect(new Set(ids).size).toBe(ids.length);
+    });
+
+    it('runs against the shipped one-team-per-user cap', async () => {
+        // Lifting the cap here would stop it resembling production.
+        const cfg = fx.env.server.services.team.config as {
+            max_teams_per_user?: number;
+        };
+        expect(cfg.max_teams_per_user ?? 1).toBe(1);
+    });
+
+    it('seats can actually call the API', async () => {
+        // An inert token would make every assertion below vacuous.
+        const res = await fx.call('GET', `/teams/${fx.a.uid}`, fx.a.seats[0].token);
+        expect(res.status).toBe(200);
+    });
+
+    // -- grants resolve for members, and only members -----------------
+
+    it('resolves a grant for every member of the team it was given to', async () => {
+        const file = await makeFile(fx.outsider.userId);
+        const permission = `fs:${file.uid}:read`;
+
+        await permissions().grantUserGroupPermission(
+            await actorFor(fx.outsider.userId),
+            fx.a.uid,
+            permission,
+        );
+
+        for (const seat of fx.a.seats) {
+            const rows = await store().readUserGroupPerms(seat.userId, [
+                permission,
+            ]);
+            expect(rows, `seat ${seat.username}`).toHaveLength(1);
+        }
+    });
+
+    it('does not resolve for the other team', async () => {
+        const file = await makeFile(fx.outsider.userId);
+        const permission = `fs:${file.uid}:read`;
+
+        await permissions().grantUserGroupPermission(
+            await actorFor(fx.outsider.userId),
+            fx.a.uid,
+            permission,
+        );
+
+        // Why the fixture has two: with one, an unscoped query still passes.
+        for (const seat of fx.b.seats) {
+            const rows = await store().readUserGroupPerms(seat.userId, [
+                permission,
+            ]);
+            expect(rows, `B seat ${seat.username}`).toHaveLength(0);
+        }
+    });
+
+    it('does not resolve for a user in no team', async () => {
+        const file = await makeFile(fx.a.owner.userId);
+        const permission = `fs:${file.uid}:read`;
+
+        await permissions().grantUserGroupPermission(
+            await actorFor(fx.a.owner.userId),
+            fx.a.uid,
+            permission,
+        );
+
+        const rows = await store().readUserGroupPerms(fx.outsider.userId, [
+            permission,
+        ]);
+        expect(rows).toHaveLength(0);
+    });
+
+    // -- revoke --------------------------------------------------------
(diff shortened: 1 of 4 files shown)


jfcastro92:
> [!WARNING]
> **Superseded — do not rely on the results below.**
>
> Captured at `c50c04a9e`, before a code review found 13 issues in this stack, before the fixes for them, and before three rebases (the last onto a `main` that had moved 9 commits). The behaviour described here is not the behaviour of the current head.
>
> Kept for history rather than deleted, so the comments that reference it still make sense. A fresh pass will be posted against the merged code.

---

## Local validation

Engine: mysql 8 · workspace with an owner and 4 provisioned seats, plus a
non-member account for the negative case.

**A workspace grant writes the row this PR's path reads.** After sharing a file
with the workspace, `user_to_group_permissions` holds exactly one group-issued
grant:

```
permission                                       issuer  group_uid                              kind
fs:20adac9e-5b81-4643-b4a7-1b22dfc2d626:read     87      13341565-5223-4c0a-9d28-ca71c9da6682   team
```

`issuer` is the owner (user 87), the group is the workspace, and `kind='team'`
scopes it away from the seeded system groups. That is the shape §3(b) of the design
describes — a resource grant to a group, resolved through the pre-existing
`user_to_group_permissions` → `jct_user_group` join rather than a parallel path.

**Members resolve it; a non-member does not.** Same file, same moment:

```
acmeseat1  (workspace member)      POST /stat -> 200, full fsEntry
acmeseat1                          GET /fs/read -> 200  "workspace share probe v1"
tmoutsider (authenticated, not a member)
                                   POST /stat -> 404 {"code":"subject_does_not_exist"}
```

The non-member answer is `subject_does_not_exist` rather than a permission error, so
the grant does not leak the file's existence to accounts outside the workspace.

**Revocation invalidates without a restart** — the cache-generation bump, which is
the load-bearing part of §4:

```
POST /share/revoke -> 200 {"status":"success","revoked":1}

attempt 1: stat -> 404   read -> 404
attempt 2: stat -> 404   read -> 404
attempt 3: stat -> 404   read -> 404
```

Same server process, no restart, immediately after the revoke. A stale per-actor
scan cache would have kept serving reads here.

**Baseline before sharing** was also 404, so the 200s above are attributable to the
grant rather than to ambient access.

jfcastro92:
## Review fixes

A code review across the three phase 4 PRs raised 13 issues. All 13 are fixed; two land here.

### A soft-deleted workspace kept resolving its grants

`PermissionStore.readUserGroupPerms` joined `jct_user_group` with no filter on the workspace itself. `TeamService.deleteWorkspace` only soft-deletes the group and suspends the `org_owned` seats — memberships and `user_to_group_permissions` rows survive.

So after deleting a workspace, its non-org-owned members (the owner at minimum) kept resolving every grant ever made to it. And because `#resolveTeamRecipient` → `getByUid` excludes soft-deleted rows, the uid no longer resolved — making that access **impossible to withdraw** through any endpoint.

The read now excludes deleted workspaces:

```sql
JOIN `group` g ON g.`id` = ug.group_id
WHERE ug.user_id = ? AND g.`deleted_at` IS NULL AND ...
```

Fixed at read time rather than by revoking on delete, because that is also correct for workspaces already deleted.

### Group grants no longer vouch for a revoked direct share

`#grantEvidence` recorded a group grant's issuer into the same `issuers` set as user-to-user grants, and `#reachingGrants` treats `issuers.has(issuerId)` as proof *that user-to-user row* is live. So if Alice shared a node with Bob directly, that grant was later withdrawn outside `unshare`, and Alice had also shared the node with a workspace Bob belongs to — Alice's dead share stayed listed as though it stood.

`GrantEvidence` now carries `groupIssuers` separately. The three consumers differ, which is the point:

| | question | counts group reach? |
| --- | --- | --- |
| `#reachingHolders` | does this holder reach it at all? | yes |
| `#liveGrants` | same, for the inbox | yes |
| `#reachingGrants` | whose grant backs **this row**? | **no** |

⚠ Worth a reviewer's attention: the first version of this change broke three existing tests, because moving group issuers out of `issuers` also removed them from `#liveGrants` and emptied members' inboxes. The distinction above is the fix.

### Tests

Both falsified — the fix reverted, the test confirmed failing, then restored:

```
 × stops resolving once the workspace is soft-deleted          (F9)
 × does not let a workspace grant keep a revoked direct share listed   (F12)
```

Full suite **7732 passed | 26 skipped**, typecheck clean.

Rebased onto current `main`, which had advanced 9 commits during this work — including the notification and events-surface changes (PUT-1679/1681/1684/1685).

jfcastro92:

Engine: mysql 8 · `teams_enabled: true` · fresh database restored from a pre-teams snapshot, then migrated by boot. Supersedes the earlier comment on this PR, which predates the review fixes.

Fixture: one workspace, three activated and email-confirmed seats, plus a non-member.

### Group grants resolve for every member and nobody else

Sharing a file with the workspace writes one group-issued grant, and all three seats resolve it while the non-member does not:

```
GET /share/shared-with-me
p4seat1x80144    count=1  issuer=p4own80144  mode=read
p4seat2x80144    count=1  issuer=p4own80144  mode=read
p4seat3x80144    count=1  issuer=p4own80144  mode=read
non-member       count=0
```

Before the share, all three were `0` — so the access is attributable to the grant rather than to anything ambient.

### Revocation invalidates immediately

```
POST /share/revoke -> success, revoked=1
p4seat1x80144    still-listed=0
p4seat2x80144    still-listed=0
p4seat3x80144    still-listed=0
```

Same server process, no restart. This is the cache-generation bump in §4 of the design: a stale per-actor scan reading would keep answering "allowed" here.

### The migration chain replayed against real data

The snapshot predates `0077_teams.sql`, so boot re-applied the whole chain, including the dedup that this stack's phase 1 introduced:

```
[mysql] 34 migration files applied
jct_user_group rows:  81 -> 63
duplicate (user_id, group_id) pairs:  0
```

---

⚠ **The soft-delete fix in this PR (finding 9) was not exercised live.** Deleting the workspace would have ended the fixture, and the remaining checks needed it alive. It is covered by `stops resolving once the workspace is soft-deleted`, falsified — reverting the `deleted_at` filter fails that test and only that test.

Finding 12 (group grants no longer vouching for a revoked direct share) is likewise unit-covered rather than driven here; reproducing it live needs a grant withdrawn *outside* `unshare`, which no endpoint does.

jfcastro92:
## Review fix: a group revoke told nobody

Correct, and it is an asymmetry this PR introduced. `revokeUserUserPermission` has always ended with `#announceRevoked`; the group twin I added here did not. `permission.revoked` is what `EventsService.settleRevokedGrant` stands on, so a member's FS watches outlived the unshare — the delivery re-check would deny them, but the row keeps holding an anchor slot, still bills, and still costs a filter evaluation per event.

The reason it needed thinking about rather than copying: the user-to-user path has one holder to name, and a group grant has none. Nothing on the row identifies who it reached — the membership does. So the announce fans out over `jct_user_group`:

```ts
if (revoked) {
    for (const memberId of await this.stores.permission.listGroupMemberIds(groupId)) {
        this.#announceRevoked(memberId, null, permission);
    }
}
```

Guarded on `revoked` so a no-op revoke stays silent. `listGroupMemberIds` is new on `PermissionStore` — the existing `listGroupMemberUuids` returns uuids for the cache bump, and the event carries a numeric id.

Regression test in `GroupGrants.test.ts` asserting every seat of the workspace appears in the announced holders. Falsified — dropping the fan-out fails it and only it:

```
× announces the revoke for every member, so their watches get settled
Tests  1 failed | 14 passed (15)
```

Full suite **7883 passed | 26 skipped**, typecheck clean.

jfcastro92:

Salazareo (approved):

jfcastro92:

| Before | After |
| --- | --- |

### Scoped, and checked

### One thing not renamed

PR #3729, opened 2026-09-03
🏗️ PUT-1726 + PUT-1727 + PUT-1729: share with a workspace

jfcastro92:
> Fifteenth in the stack: #3704 → … → #3725 → #3726 → this. Review bottom-up.

Covers **PUT-1726**, **PUT-1727** and **PUT-1729**. Together because a recipient without the listing work produces a share that is created correctly, resolves correctly, and never appears in "Shared with me" — a state nobody would ship, and therefore not one worth reviewing on its own.

This is the PR that makes `puter.fs.share('/Docs', { team: uid }, 'write')` actually work end to end.

## The recipient

`ShareRecipient` gains `team` (uid) and `teamHandle`, resolved **before** the email and username branches and never falling through to them.

```js
await puter.fs.share('/me/Docs', { team: '<uid>' }, 'write');            // canonical
await puter.fs.share('/me/Docs', { teamHandle: 'design-team' }, 'write'); // convenience
```

Passing both is a 400 rather than a precedence rule, so a call site always shows which was used. That is the whole reason they are separate fields: a handle is released on soft delete and can be reclaimed by an unrelated workspace, so a **scripted** share to a handle keeps working and silently starts sharing with strangers. `getByHandle` filters soft-deleted, so a released handle 404s until someone reclaims it.

There is no bare-string spelling. `toShareRecipients` reads a bare string as an email when it contains `@` and a username otherwise, so any prefix convention would change how an existing string is interpreted — a breaking change to a released SDK. There is a test that a workspace handle passed as a `username` does **not** resolve to the workspace.

## One grant, not one per member

`ACLService.setUserGroup` mirrors `setUserUser`: same read-modify-write, same one-mode-per-node rule, under a node lock keyed on the group.

The grant is a single row against the workspace, so:

- membership changes apply without touching it — **a member who joins afterwards gets access**, which is asserted;
- a share spends **one** unit of the daily quota however many members there are, and re-sharing the same workspace reports `isNew: false`.

The index row carries `holder_group_id` and leaves `holder_user_id` NULL, so `0078`'s group index constrains it rather than the user-holder one.

## ⚠ The listing is the silent failure

`#grantEvidence` gains group grants as a **third source** alongside the linked and flat reads. Without it, a team share is created correctly, resolves correctly for every member, and is then filtered out of every listing as dead — nothing errors, the share is simply absent from the one place a user would look.

`listByHolder` and `countByHolder` union the caller's workspaces into the same keyset page. `ORDER BY id` still holds, so pagination is unaffected by which kind of holder a row names. The count matters too, or `includeTotal` undercounts a member's workspace shares.

## Unsharing revokes the grant, not just the row

Deleting the index row alone would hide the share while leaving every member holding real access — worse than a listing bug, because it looks fixed. An owner may clear any issuer's grant of their node; anyone else only their own, as for user shares.

---

## Verification

```
      src/backend/services/{share,permission,acl}/ src/backend/stores/share/
 Test Files  12 passed (12)
      Tests  454 passed (454)

$ PUTER_TEST_DB_ENGINE=postgres … TeamShare.test.ts GroupGrants.test.ts
      Tests  29 passed (29)

$ npm run test:backend
      Tests  7364 passed | 26 skipped (7390)

$ npm run typecheck
```

<details>

```
baseline                       0 failed | 16 passed
#liveGrants group evidence     3 failed | 13 passed
listing union (groupIds)       3 failed | 13 passed
group index row write         12 failed |  4 passed
setUserGroup grant             4 failed | 12 passed
both-fields rejection          1 failed | 15 passed
handle resolves by handle      1 failed | 15 passed
unshare revokes the grant      1 failed | 15 passed
count unions groups            1 failed | 15 passed
restored                       0 failed | 16 passed
```

`unshare revokes the grant` initially reported **0 failed**, and the test was at fault: it asserted only that the share left the listing, which also happens when the index row is deleted. So a version that dropped the row without revoking the permission passed — the exact silent-retained-access bug worth catching. It now asserts the grant itself is gone.

</details>

## Not here

Member notifications (PUT-1761). As things stand, sharing with a workspace tells **nobody**: `notifyShared` keys off `ResolvedShare.holderId`, a single user id, and a team share has none. `holderGroupId` and `holderTeam` are populated on the resolved share so that ticket has what it needs.

The GUI (PUT-1762) is what makes this reachable without writing code.

---

Closes PUT-1726, PUT-1727 and PUT-1729.

### DIFF
--- a/src/backend/controllers/share/ShareController.ts
+++ b/src/backend/controllers/share/ShareController.ts
@@ -144,7 +144,7 @@ export class ShareController extends PuterController {
         const results: ShareOutcome[] = await Promise.all(
             settled.map(async (outcome, index) => {
                 const { recipient, item } = pairs[index];
-                const label = recipient.email ?? recipient.username ?? '';
+                const label = this.#recipientLabel(recipient);
                 if (outcome.status === 'fulfilled') {
                     // The whole share, not just an acknowledgement, so a caller
                     // needn't re-read to learn what it created.
@@ -227,7 +227,7 @@ export class ShareController extends PuterController {
         );
         const results: ShareOutcome[] = settled.map((outcome, index) => {
             const { recipient, item } = pairs[index];
-            const label = recipient.email ?? recipient.username ?? '';
+            const label = this.#recipientLabel(recipient);
             if (outcome.status === 'fulfilled') {
                 return {
                     recipient: label,
@@ -554,9 +554,13 @@ export class ShareController extends PuterController {
         const unique: T[] = [];
         const seen = new Map<string, number>();
         const indexOf = pairs.map((pair) => {
+            // Every recipient field, or two teams key alike and collapse
+            // into one -- the second reporting the first's outcome as its own.
             const key = JSON.stringify([
                 pair.recipient.email ?? null,
                 pair.recipient.username ?? null,
+                pair.recipient.team ?? null,
+                pair.recipient.teamHandle ?? null,
                 pair.item.uid ?? null,
                 pair.item.path ?? null,
             ]);
@@ -638,6 +642,17 @@ export class ShareController extends PuterController {
         return value === NO_APP ? null : value;
     }
 
+    /** Echoes back the identifier the caller named, so results are matchable. */
+    #recipientLabel(recipient: ShareRecipient): string {
+        return (
+            recipient.email ??
+            recipient.username ??
+            recipient.teamHandle ??
+            recipient.team ??
+            ''
+        );
+    }
+
     #recipients(body: Record<string, unknown>): ShareRecipient[] {
         const raw = body.recipients ?? body.recipient;
         const list = Array.isArray(raw) ? raw : [raw];
@@ -659,6 +674,24 @@ export class ShareController extends PuterController {
                     typeof rec.email === 'string' ? rec.email.trim() : '';
                 const username =
                     typeof rec.username === 'string' ? rec.username.trim() : '';
+                const team =
+                    typeof rec.team === 'string' ? rec.team.trim() : '';
+                const teamHandle =
+                    typeof rec.teamHandle === 'string'
+                        ? rec.teamHandle.trim()
+                        : '';
+                // Ambiguous rather than a precedence rule to memorise.
+                if (team && teamHandle) {
+                    throw new HttpError(
+                        400,
+                        'pass `team` or `teamHandle`, not both',
+                        { legacyCode: 'bad_request' },
+                    );
+                }
+                if (team || teamHandle) {
+                    out.push(team ? { team } : { teamHandle });
+                    continue;
+                }
                 if (email || username) {
                     out.push(email ? { email } : { username });
                 }
--- a/src/backend/controllers/share/clientShare.ts
+++ b/src/backend/controllers/share/clientShare.ts
@@ -63,6 +63,8 @@ export async function toClientShare(
         is_dir: share.isDir,
         issuer: share.issuer.username,
         holder: share.holder.username,
+        // Named, or a team share reads as a share with nobody.
+        ...(share.holderTeam ? { holder_team: share.holderTeam } : {}),
         created_at: share.createdAt,
         issued_by_app: share.issuedByApp ?? null,
         inherited_from: share.inheritedFrom ?? null,
--- a/src/backend/controllers/team/TeamController.http.test.ts
+++ b/src/backend/controllers/team/TeamController.http.test.ts
@@ -236,6 +236,115 @@ describe('team endpoints over HTTP', () => {
         );
         expect(theirs.status).toBe(404);
     });
+
+    // -- sharing with a team -------------------------------------
+
+    it('names the team it shared with instead of an empty recipient', async () => {
+        const { team } = await makeTeam();
+        const file = `/${env.users.user.username}/Documents/label-${Math.random()
+            .toString(36)
+            .slice(2, 8)}.txt`;
+        const written = await call('POST', '/fs/write', env.users.user.token, {
+            fileMetadata: { path: file, size: 3, contentType: 'text/plain' },
+            fileContent: 'abc',
+        });
+        expect(written.status).toBe(200);
+
+        const shared = await call('POST', '/share', env.users.user.token, {
+            recipients: [{ team: team.uid }],
+            items: [file],
+            mode: 'read',
+        });
+        expect(shared.status).toBe(200);
+        const body = (await shared.json()) as {
+            results: { recipient: string; status: string }[];
+        };
+
+        expect(body.results[0].status).toBe('success');
+        // Echoes the identifier the caller named; '' would leave a client with
+        // nothing to render or to match its request against.
+        expect(body.results[0].recipient).toBe(team.uid);
+    });
+
+    it('shares with two teams rather than collapsing them into one', async () => {
+        const second = await call('POST', '/teams', env.users.user.token, {
+            name: 'Second',
+            handle: randomHandle(),
+        });
+        expect(second.status).toBe(200);
+        const teamB = (await second.json()) as { uid: string };
+        const { team: teamA } = await makeTeam();
+
+        const file = `/${env.users.user.username}/Documents/two-${Math.random()
+            .toString(36)
+            .slice(2, 8)}.txt`;
+        const written = await call('POST', '/fs/write', env.users.user.token, {
+            fileMetadata: { path: file, size: 3, contentType: 'text/plain' },
+            fileContent: 'abc',
+        });
+        const fileUid = ((await written.json()) as { fsEntry: { uid: string } })
+            .fsEntry.uid;
+
+        const shared = await call('POST', '/share', env.users.user.token, {
+            recipients: [{ team: teamA.uid }, { team: teamB.uid }],
+            items: [file],
+            mode: 'read',
+        });
+        expect(shared.status).toBe(200);
+        const body = (await shared.json()) as {
+            results: { recipient: string; status: string }[];
+        };
+
+        expect(body.results).toHaveLength(2);
+        expect(body.results.every((r) => r.status === 'success')).toBe(true);
+
+        // The response shape alone proves nothing: `results` is built from the
+        // request's pairs, so a collapsed pair still reports two successes with
+        // the right labels. Only the shares that exist afterwards show it.
+        const mine = await call(
+            'GET',
+            '/share/shared-by-me?limit=100',
+            env.users.user.token,
+        );
+        expect(mine.status).toBe(200);
+        const listing = (await mine.json()) as {
+            items: { uid_entry?: string; holder_team?: { uid: string } }[];
+        };
+        const holders = listing.items
+            .filter((i) => i.uid_entry === fileUid)
+            .map((i) => i.holder_team?.uid)
+            .filter(Boolean)
+            .sort();
+        expect(holders).toEqual([teamA.uid, teamB.uid].sort());
+    });
+
+    it('names the team by handle when the caller shared by handle', async () => {
+        const handle = randomHandle();
+        const res = await call('POST', '/teams', env.users.user.token, {
+            name: 'Byhandle',
+            handle,
+        });
+        expect(res.status).toBe(200);
+
+        const file = `/${env.users.user.username}/Documents/handle-${Math.random()
+            .toString(36)
+            .slice(2, 8)}.txt`;
+        await call('POST', '/fs/write', env.users.user.token, {
+            fileMetadata: { path: file, size: 3, contentType: 'text/plain' },
+            fileContent: 'abc',
+        });
+
+        const shared = await call('POST', '/share', env.users.user.token, {
+            recipients: [{ teamHandle: handle }],
+            items: [file],
+            mode: 'read',
+        });
+        expect(shared.status).toBe(200);
+        const body = (await shared.json()) as {
+            results: { recipient: string }[];
+        };
+        expect(body.results[0].recipient).toBe(handle);
+    });
 });
 
 describe('team endpoints with teams_enabled off', () => {
--- a/src/backend/services/acl/ACLService.ts
+++ b/src/backend/services/acl/ACLService.ts
@@ -280,6 +280,136 @@ export class ACLService extends PuterService {
      *
      * Caller (controller) validates that both actors are user-type.
      */
+    // -- Group holders ---- one grant, resolved per member at scan time ----
+
+    /** The group analogue of `statUserUser`; no holder actor to validate. */
+    async statUserGroup(
+        issuer: Actor,
+        groupUid: string,
+        resource: ResourceDescriptor,
+    ): Promise<StatPermissionsResult> {
+        if (issuer.app || issuer.accessToken)
+            throw new HttpError(403, 'issuer must be a user actor', {
+                legacyCode: 'forbidden',
+            });
+
+        const out: StatPermissionsResult = {};
+        const ancestors = await resource.resolveAncestors();
+        for (const ancestor of ancestors) {
+            // Both namespaces: `manage:fs:<uid>` sits outside `fs:<uid>`.
+            const prefixes = [
+                PermissionUtil.join('fs', ancestor.uid),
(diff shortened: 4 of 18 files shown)


jfcastro92:
> [!WARNING]
> **Superseded — do not rely on the results below.**
>
> Captured at `c50c04a9e`, before a code review found 13 issues in this stack, before the fixes for them, and before three rebases (the last onto a `main` that had moved 9 commits). The behaviour described here is not the behaviour of the current head.
>
> Kept for history rather than deleted, so the comments that reference it still make sense. A fresh pass will be posted against the merged code.

---

## Local validation

Engine: mysql 8 · real file, real share, real read — end to end over HTTP.

**Setup.** Owner writes an actual file:

```
POST /fs/write {"fileMetadata":{"path":"/tmowner/Documents/probe.txt","size":24,
                "contentType":"text/plain"},"fileContent":"workspace share probe v1"}
  -> 200  uid=20adac9e-5b81-4643-b4a7-1b22dfc2d626
```

**Before sharing**, a seat cannot see it:

```
acmeseat1  POST /stat -> 404 {"code":"subject_does_not_exist"}
```

**Share with the workspace by uid:**

```
POST /share {"recipients":[{"team":"13341565-..."}],
             "items":["/tmowner/Documents/probe.txt"],"mode":"read"}
  -> 200 {"status":"success","results":[{"uid":"acf13a97-...","mode":"read",
          "path":"/tmowner/Documents/probe.txt","is_new":true,
          "uid_entry":"20adac9e-...","issuer":"tmowner","holder":null, ...}]}
```

**A seat then reads the actual content:**

```
acmeseat1  POST /stat            -> 200, full fsEntry
acmeseat1  GET  /fs/read         -> 200
           workspace share probe v1
```

**A non-member still cannot:**

```
tmoutsider POST /stat -> 404 {"code":"subject_does_not_exist"}
```

**The share row uses the `holder_group_id` column** with `holder_user_id` left NULL,
which is what makes the existing `idx_share_holder_entry_issuer` unique index
non-binding for team rows (NULLs distinct):

```
uid                                   holder_user_id  holder_group_id  team_uid
c47a95fd-309e-4484-aa0e-e82996c7c0fc  NULL            139              13341565-...
```

**Revoke takes effect immediately, no restart:**

```
POST /share/revoke -> 200 {"status":"success","revoked":1}
attempt 1: stat -> 404   read -> 404
attempt 2: stat -> 404   read -> 404
attempt 3: stat -> 404   read -> 404
```

---

**One small thing for review:** the response labels the recipient as an empty
string for team shares — `"recipient":""` in the create result above, and
`{"recipient":"","path":"...","status":"success"}` on revoke. `ShareController`
builds it as:

```ts
const label = recipient.email ?? recipient.username ?? '';
```

with no branch for `team` / `teamHandle`. Harmless to the operation, but a client
rendering "shared with X" from that field gets a blank. Cheap to add a team branch
here, or to leave deliberately if the client is expected to use the uid it sent.

jfcastro92:
## Fixed: team shares reported an empty recipient

Following up on the observation in the validation comment above.

`ShareController` built the outcome label at two sites with no team branch, so a
share to a workspace came back as `"recipient": ""` on both create and revoke:

```ts
const label = recipient.email ?? recipient.username ?? '';
```

Both now go through one helper:

```ts
/** Echoes back the identifier the caller named, so results are matchable. */
#recipientLabel(recipient: ShareRecipient): string {
    return (
        recipient.email ??
        recipient.username ??
        recipient.teamHandle ??
        recipient.team ??
        ''
    );
}
```

**Echoing the identifier the caller supplied** rather than resolving a display name:
a caller that sent `{teamHandle: 'acme'}` gets `acme` back and can match results to
its own request positionally *and* by value. Resolving to the workspace's `name`
would look nicer but breaks that matching, and `name` is not unique.

`teamHandle` is checked before `team` for the same reason — if the caller named a
handle, the handle is what they can match against.

**Two tests, both falsified.** Added to `TeamController.http.test.ts`, which already
runs with `teams_enabled: true` and has a workspace fixture (the `ShareController`
suite boots with the flag off, so it cannot create one):

```
✓ names the workspace it shared with instead of an empty recipient
✓ names the workspace by handle when the caller shared by handle
```

Reverting the helper to the old expression fails exactly those two:

```
 × names the workspace it shared with instead of an empty recipient
 × names the workspace by handle when the caller shared by handle
Tests  2 failed | 11 passed (13)
```

Full suite **7525 passed | 26 skipped**, typecheck clean.

---

**One related thing left alone deliberately.** `clientShare.ts` maps
`holder: share.holder.username`, so a team share also reports `"holder": null` —
visible in the share result quoted in the comment above. `ShareService` does resolve
`holderTeam: {uid, name, handle}`, so the information exists and could be surfaced as
e.g. `holder_team`.

I did not do that here because it **adds a field to the share response shape**, which
every share listing returns and which puter-js and the GUI both consume. That is a
wider call than fixing a blank label, and it belongs with whoever owns the client
contract. Happy to add it in this PR if you want it, or file it as a follow-up.

jfcastro92:
## Review fixes

A code review across the three phase 4 PRs raised 13 issues. All 13 are fixed; eight land here.

### The root cause behind four of them

The codebase had an invariant — **`holder_user_id IS NULL` means "pending invite"** — and this PR gave that column a second reason to be null: a workspace share, whose holder is `holder_group_id`. Every site still reading the old meaning misclassified workspace shares:

| Site | What it did |
| --- | --- |
| `revokeSharedByMe` | took the invite branch: deleted the index row, returned `revoked: 1`, **never touched the grant** — members kept access, and with the row gone it could not be listed or revoked again |
| `listSharedByMe` | `pending = !row.holder_user_id` — the owner's outbound listing showed every workspace share as a pending invite to a blank address, and checked its liveness against invite authority rather than the group grant |
| `listPendingOnFsentry` | `WHERE holder_user_id IS NULL` — the share dialog showed a phantom pending invite and never the workspace |
| `deletePendingByIssuerSubtree` | same predicate — deleted the index row while `listByFsentrySubtree` (which does the actual revoking) filters `IS NOT NULL`, so the grant was left behind, dormant until the issuer regained authority |

Each now checks both columns. `revokeSharedByMe` routes group rows to `#unshareTeam`, `#resolvedShareRow` no longer marks them pending, and the listing gained `#liveGroupGrants` — the group analogue of `#reachingGrants`, because an index row outlives the grant it records.

Three further store predicates use `holder_user_id IS NULL` (`listPendingByEmail`, `upsertPending`, `applyPending`) but are **protected** by an additional predicate — group rows carry `recipient_email = ''`, and `applyPending` is keyed by uid. `listPendingByEmail` got the guard anyway, since an empty address would otherwise match.

### Sharing with two workspaces shared with one

`#dedupePairs` keyed on `[email, username, uid, path]` — no `team`/`teamHandle`. Two workspace recipients for one item produced the identical key `[null,null,null,"/x"]`, so the second collapsed into the first. Both slots then reported the *first's* fulfilled outcome, so the response said success for a workspace that got nothing. Same collapse on `/share/revoke`, leaving real access standing while claiming it was withdrawn.

⚠ Note for review: asserting on the response cannot catch this. `results` is built from the request's pairs, so a collapsed pair still returns two entries with the right labels and `status: success`. Only the shares that exist afterwards reveal it — which is what the test checks.

### Unsharing aborted when a delegate had lost authority

`#unshareTeam` revoked each prior issuer's grant by *impersonating* them, but `revokeUserGroupPermission` gates on `canManagePermission` and throws 403. So: owner grants Bob `manage` on /Docs → Bob shares /Docs with Acme → owner revokes Bob's manage → owner unshares /Docs from Acme → the loop hits Bob first, throws, and aborts **before the owner's own grant is revoked or any row deleted**.

Now mirrors the user-to-user path: authority is the caller's throughout, with `{ issuerUserId }` naming whose grant to clear. `revokeUserGroupPermission` gained that override (the user-to-user one already had it). It also fixes the audit, which previously attributed the revoke to the impersonated issuer.

### Members got no live events, and grants outlived deleted files

- `listReaching` filters `holder_user_id IS NOT NULL` and is the sole feed for `outer.gui.item.*`. Workspace members saw a shared folder but received no create/update/move/delete events inside it — it went stale until a manual refresh. New `listGroupReachingMembers` returns one row per member, shaped as a holder row so the fan-out needs no group branch.
- `onEntryDeleted` deleted only `user_to_user_permissions` rows. Group grants on the deleted entry's uuid survived and members' cached readings were never invalidated, so a member's `fs:<uuid>:*` reading kept answering "allowed" until the TTL, with orphan rows accumulating permanently.

### Also

`toClientShare` never passed `holderTeam` through, so a workspace share reached the wire as `holder: null` with nothing naming the workspace. It now carries `holder_team`.

### Tests — all falsified

Each fix was reverted, the test confirmed failing, then restored:

```
 × lists a workspace share as a real share, not a pending invite
 × does not show a phantom invite on the item it was shared with
 × revoking from the outbound listing withdraws the grant, not just the row
 × shares with two workspaces rather than collapsing them into one
 × clears the caller own grant even when another issuer lost authority
 × counts members as reached, so they get live events on a shared folder
 × clears the group grant when the entry itself is deleted
```

⚠ Two of these were **vacuous on the first attempt** and are worth knowing about, because the same trap applies to anything testing this area:

- The revoke test probed the member's *inbox*. It passed with the fix reverted — deleting the index row also empties the listing. It now probes `readUserGroupPerms`, the grant members actually resolve.
- The phantom-invite test passed with the store guard reverted, because the `#resolvedShareRow` fix masks the symptom. It now also asserts at the store.

Full suite **7732 passed | 26 skipped**, typecheck clean. Rebased onto current `main`, which advanced 9 commits during this work.

jfcastro92:

Engine: mysql 8 · `teams_enabled: true` · fresh database restored from a pre-teams snapshot, then migrated by boot. Supersedes the earlier comment on this PR, which predates the review fixes.

Fixture: one workspace, **three provisioned seats**, all activated and email-confirmed, plus a non-member. Three is the minimum that distinguishes "told the members" from "told a member".

```
username         email_confirmed  suspended  org_owned
p4own80144             1            NULL         0
p4seat1x80144          1            NULL         1
p4seat2x80144          1            NULL         1
p4seat3x80144          1            NULL         1
```

### The migrations replayed from a pre-teams schema

The snapshot predates `0077_teams.sql`, so restoring it removed the team columns and boot had to re-add them — which re-exercised the whole chain against real duplicate data:

```
[mysql] 34 migration files applied
group.kind restored:                 yes
jct_user_group rows:  81 -> 63       (the dedup, on real rows)
duplicate (user_id, group_id) pairs:  0
```

### The listing union — every member sees it

This is what the earlier pass never covered: it tested `stat` and `fs/read`, i.e. *resolution*, and never listed. A share can resolve perfectly and still be invisible where a member would look.

Before sharing, `GET /share/shared-with-me` for each seat: `0`. After sharing with the workspace:

```
p4seat1x80144    count=1  issuer=p4own80144  mode=read
p4seat2x80144    count=1  issuer=p4own80144  mode=read
p4seat3x80144    count=1  issuer=p4own80144  mode=read
non-member       count=0
```

### The owner's outbound listing (findings 3 and 12)

```
GET /share/shared-by-me
{'pending': None, 'recipient_email': None, 'holder': None,
 'holder_team': '3ee4033b-7c0c-4c99-b66f-a8d46cdbdd29'}
rows: 1
```

`pending: None` — no longer a phantom invite to a blank address — and the workspace is named.

### The share dialog (finding 4) — and a gap the live run caught

`GET /share/shares?uid=<file>` returned **0 rows** for a live workspace share. The fix for finding 4 removed the phantom invite, but nothing put the workspace in its place: `listByFsentries` filters `holder_user_id IS NOT NULL` and the invite feed now excludes group rows, so a workspace share matched neither feed.

Before the review it showed a *wrong* row; after, it showed *none*. Both are wrong — the item's manager could not see it was shared with a workspace, so could not revoke it there. That is also PUT-1762's "Done when".

Added `listGroupOnFsentry` and wired it into `listSharesOf` with the same `#liveGroupGrants` liveness check as the outbound listing. After the fix:

```
rows: 1
{'pending': None, 'holder': None, 'holder_team': '3ee4033b-...', 'mode': 'read'}
```

Test `shows the workspace in the share dialog for that item`, falsified — dropping the group rows from the return fails it and nothing else.

**This was only found by listing live.** No unit test covered the dialog for a workspace share.

### Sharing names the workspace, not an empty string

```
POST /share {"recipients":[{"team":"<uid>"}], ...}
  -> status: success  recipient: '3ee4033b-7c0c-4c99-b66f-a8d46cdbdd29'
                      holder_team: 3ee4033b-7c0c-4c99-b66f-a8d46cdbdd29
```

Both the recipient label and `holder_team` reach the wire — neither had ever been checked live before.

### Revocation clears every member's listing, no restart

```
POST /share/revoke -> success, revoked=1
p4seat1x80144    still-listed=0
p4seat2x80144    still-listed=0
p4seat3x80144    still-listed=0
```

Same process, immediately after. A stale scan cache or a surviving index row would both show here.

---

Full suite **7733 passed | 26 skipped**, typecheck clean, on a `main` that advanced 9 commits during this work.

jfcastro92:
## SDK bridge for workspace shares

Follow-up: the backend accepted `{team}` / `{teamHandle}` and returned `holder_team`, but **puter.js mapped neither**, so the wire contract existed with nothing able to speak it. Both halves are now in `shareUtil.js`.

### Outbound — the recipient was silently dropped

`toShareRecipients` copied only `email` and `username` off an object recipient, so `puter.fs.share(path, { team: uid })` sent `{}`. The call reached the backend with an empty recipient and could report success having shared with nobody.

```js
...(record.team ? { team: String(record.team) } : {}),
...(record.teamHandle ? { teamHandle: String(record.teamHandle) } : {}),
```

Object form only — no bare-string spelling. A `#handle` or `team:` prefix would reinterpret strings that already mean an email or a username, which is breaking for a released SDK. PUT-1726 rules that out explicitly.

### Inbound — the workspace had no name

`toShare` never read `holder_team`. Since a workspace share has `holder: null` by design, a client saw a share that reached *nobody*:

```js
...(row.holder_team || row.holderTeam
    ? { holderTeam: row.holder_team ?? row.holderTeam }
    : {}),
```

### Published types

`ShareRecipient` and `Share` in `types.js` updated, since these are the SDK's public surface:

```
@typedef {string | { email?: string, username?: string, team?: string, teamHandle?: string }} ShareRecipient

@property {string | null} holder  ... Null for a workspace share, which has no
  individual holder -- see `holderTeam`.
@property {{ uid: string, name: string | null, handle: string | null }} [holderTeam]
```

`npm run check:puterjs:types` — declarations generate and type-check cleanly.

### Tests — falsified in both directions

```
✓ sends a workspace recipient through instead of dropping it
✓ sends a workspace handle under its own field
✓ publishes the workspace a share reached
```

Dropping the outbound mapping fails the first two; dropping the inbound one fails the third; neither affects the other.

```
[A] outbound removed:  2 failed | 205 passed
[B] inbound removed:   1 failed | 206 passed
    restored:          207 passed
```

Backend suite **7733 passed | 26 skipped**, typecheck clean.

---

This unblocks **PUT-1762** (share with a workspace from the share dialog), which is written against `{ team: uid }` and `holder_team` and could not have worked without these two mappings.

jfcastro92:

Salazareo (approved):
code looks good but is this broadcasting to whole team?

jfcastro92:

| Before | After |
| --- | --- |

### Scoped, and checked

### One thing not renamed

jfcastro92:
> code looks good but is this broadcasting to whole team?

Yes — a share to a team notifies every member, and that is the intent: the grant reaches all of them, so telling only some would misrepresent what happened.

It is bounded rather than unbounded. `listMemberIdsByGroupId` reads `NOTIFY_FANOUT_CAP` (500) with `LIMIT cap + 1`, so a pathological team cannot turn one share into an unbounded send. Above the cap the extra members are not notified — the grant still reaches them, only the notification is dropped.

Two things keep the volume down in practice:

- **One notice per member per share**, not per file. `ShareNotificationService` folds a multi-file share into a single notification.
- **`allowedByGroup` caches the block check per team**, so a fan-out does not re-run `isBlocked` per member per share.

⚠ Worth flagging while you are looking at it: the cap is silent. A team of 600 gets 500 notifications and no signal that 100 were skipped. That is the behaviour today; if you would rather it logged or paged, say so and I will add it.

PR #3730, opened 2026-09-03
fix: stop video generation timeouts and failures from paging (PUT-1620)

404oops:
## What

Video generation timeouts, transient poll failures, and provider-side job failures no longer surface as unhandled 500s that page on-call. They leave as honest statuses with stable codes, and a caller that disconnects mid-job stops the work before anything is metered.

**Shared poll loop** ([`polling.ts`](src/backend/drivers/ai-video/providers/polling.ts)) used by the Veo, Together and BytePlus providers:
- A job that outlives the window returns **504 `upstream_timeout`**. The window is ten minutes for every provider; Together and BytePlus move up from five.
- A transient poll failure (timeout, dropped connection, 408 / 429 / 5xx) is a missed poll, not a failed job. The eventual 504 carries the last poll error as its `cause`.
- A client disconnect ends the wait with **400 `client_aborted`** and nothing is metered. `DriverController` fires an `AbortController` when the response closes before finishing and exposes it as a typed `abortSignal` on the request `Context`.

**Failed-job classification** (`videoJobFailure`), carrying `provider` and the provider's own `upstreamCode` in the body:
- content-filter wording or code → **400 `bad_request`** with `errorCode: moderation_flagged`
- rejected input (`invalid…`, `unsupported…`, `missing…`, "not supported") → **400 `upstream_bad_request`**
- anything else → **502 `upstream_failed`**
- Veo's filtered output keeps `disallowed_value` as its code and gains the same `errorCode`, with Google's reason in the message.

**Shared upstream-error util** ([`upstreamErrors.ts`](src/backend/drivers/util/upstreamErrors.ts)): the timeout predicate, a transient-error predicate, and the sanitizer + content-filter pattern lifted out of the Replicate provider, which now imports them.

**Driver boundary and chat driver**: a status-less SDK connection timeout (`APIConnectionTimeoutError`, undici timeouts) becomes **504 `upstream_timeout`** in `translateProviderError`. The chat driver records `timedOut` per attempt: an all-timeout chain is a 504, and a timed-out attempt counts as an upstream signal so a mixed chain is `upstream_failed` rather than a paging `internal_error`. The Together chat client gets the same ten-minute request timeout as Claude; it inherited the SDK's one minute.

**Docs**: `txt2vid.md` gains an Errors section covering the codes above.

## Why

Eleven alarms in a week (PUT-1620) were "the provider is slow" or "the provider failed" dressed as our own crash. The alarm gate already skips `upstream_*` and `client_*` codes, so tagging these correctly is what stops the paging. The disconnect handling closes a billing hole: the poll used to keep running after the ALB or client dropped the connection and metered on success.

## Reviewer notes

- **Scope**: the OpenAI video provider is untouched beyond an import path. Its API shuts down on 2026-09-24; retiring it and picking a new default provider is a separate ticket.
- **Status change**: Together and BytePlus failed jobs used to be 400 `upstream_failed`; provider-side failures are now 502. puter.js never exposes the HTTP status to callers, and `txt2vid` is not marked retry-safe, so the SDK's transient auto-retry does not trigger on the new 5xx codes.
- **Pattern is heuristic**: Together's failure codes are undocumented. The input-rejection pattern was widened during testing to cover `invalidDuration` and `missingFrameImagesForImageToVideoModel`, both observed live.
- **Security**: error bodies now carry `provider` and the provider's error code string. No internal paths, secrets or other users' data. The `client_aborted` body is written to a closed socket.
- **Two pre-existing catalog bugs surfaced** during live testing, not fixed here: Together's MiniMax `video-01-director` only accepts 6-second clips (catalog pins 5, so every request fails), and Kling 2.1 standard is image-to-video only (catalog offers it for text prompts).
- **One prettier reflow** rode along in `DriverController.ts` (a pre-existing type-union formatting nit).

## Testing

- Full backend suite green (276 files); new tests for the poll loop, classifier, predicates, disconnect handling at controller and provider level, transient polls, and both chat timeout cases. Typecheck against baseline, eslint, prettier clean.
- **Docker Compose, mock provider** (BytePlus pointed at a local ModelArk mock): success, two-503-then-success, refusal → 400 `moderation_flagged`, bad parameter → 400 `upstream_bad_request`, internal failure → 502, client hang-up at 12s → polling stopped after the third poll, and a full ten-minute wait → **504 `upstream_timeout`** through Caddy after 121 polls. No alarms in the container.
- **Docker Compose, real providers**: Veo 3.1 lite rendered a 4s clip in 42s; a public-figure prompt returned 400 `disallowed_value` + `moderation_flagged` with Google's reason; Together Wan 2.7 rendered in 129s; Together MiniMax and Kling failures came back as 400 `upstream_bad_request` with their codes; Together chat completed with the new timeout. No alarms.

## Follow-ups

- ALB idle timeout in production must be at or above the ten-minute window for clients to receive a long result; the disconnect guard removes the billing hazard regardless.
- Together catalog fixes above.
- New default video provider and Sora retirement before 2026-09-24.

### DIFF
--- a/src/backend/controllers/drivers/DriverController.concurrent.test.ts
+++ b/src/backend/controllers/drivers/DriverController.concurrent.test.ts
@@ -263,19 +263,24 @@ describe('DriverController — concurrent acquire/release', () => {
     });
 
     it('does not attach release listeners when the driver declares no concurrent config', async () => {
-        // The optimisation that lets the existing test stubs in
-        // DriverController.test.ts get away without an EventEmitter-shaped
-        // `res`: skip the once() wiring entirely when there's no spec.
+        // With no spec there is nothing to release, so the gate must not
+        // wire `finish`/`close` release listeners. The client-disconnect
+        // hook is separate and always present: exactly one `close` listener.
         const driver = makeSyntheticDriver();
         (driver as { concurrent?: unknown }).concurrent = undefined;
         const controller = buildController(driver);
         const handler = captureCallHandler(controller);
 
-        // Bare object with no event-emitter surface — exposes the bug
-        // case where the gate would try to call `res.once`.
+        // Minimal `res` that only records which events get listeners.
+        const listened: string[] = [];
         const bareRes = {
             statusCode: 200,
             body: undefined as unknown,
+            writableFinished: false,
+            once(event: string) {
+                listened.push(event);
+                return this;
+            },
             status(code: number) {
                 this.statusCode = code;
                 return this;
@@ -301,5 +306,6 @@ describe('DriverController — concurrent acquire/release', () => {
             bareRes as unknown as Response,
         );
         expect(bareRes.body).toMatchObject({ success: true, result: 'pong' });
+        expect(listened).toEqual(['close']);
     });
 });
--- a/src/backend/controllers/drivers/DriverController.errors.test.ts
+++ b/src/backend/controllers/drivers/DriverController.errors.test.ts
@@ -35,9 +35,10 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Readable, Writable } from 'node:stream';
 import type { Request, RequestHandler, Response } from 'express';
+import { APIConnectionTimeoutError } from 'openai';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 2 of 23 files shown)


Salazareo (approved):
lgtm

PR #3733, opened 2026-09-03
🏗️ PUT-1712: emit workspace billing events for the payment integration

jfcastro92:
> Replaces #3722, which was based on the now-closed #3721. Stack: #3704 → #3705 → #3708 → #3709 → #3710 → #3712 → #3713 → #3714 → #3720 → this.

**PUT-1712.** The trigger for billing, and after the OSS/prod split this is the *whole* of what OSS owes it.

The charge lives outside this repo — the same place that cancels Stripe subscriptions off `user.delete`. That event is the precedent: emit a typed event, let an extension do the external side effect.

## The events

| Event | Billing meaning |
| --- | --- |
| `team.account.created` | Start charging the owner for a seat |
| `team.account.disabled` | Stop it; the account still holds its bytes |
| `team.account.enabled` | Resume it |
| `team.account.deleted` | Stop it permanently |
| `team.deleted` | Stop every seat charge for that workspace |

Each carries the workspace uid, the affected account, and the owner's `stripe_customer_id`.

## ⚠ `team.account.deleted` has to be captured before the delete

`jct_user_group.user_id` is `ON DELETE CASCADE` (`0015_group.sql`). By the time a listener on `user.delete` runs, the membership row is gone and **nothing can say which workspace paid for the account** — the payload is the only surviving evidence.

So `cascadeDelete` captures the seat identity before teardown and emits after, exactly as it already does for `uuid` / `stripe_customer_id` and for the same reason. A test asserts the membership is unreadable afterwards, so the constraint that forces this design is pinned rather than assumed.

## ⚠ `stripe_customer_id` does not exist on sqlite

It ships in `mysql_mig_1` and `postgres_mig_1` but no sqlite migration — `cascadeDelete` already wraps its read in a `try`/`catch` for this. `getStripeCustomerId` does the same and degrades to `null`, so a dev or self-hosted install emits usable events rather than throwing inside a lifecycle operation. The postgres suite covers the branch that returns a real value.

## Deleting a workspace emits per seat

One `team.account.disabled` per seat plus the workspace marker, not one bulk event: the accounts persist, suspended, holding their files and their usernames.

**Deleting a workspace is not a way to stop paying for the accounts in it.** Deleting the accounts is. PUT-1740's confirmation dialog has to say so, because the opposite is the natural assumption.

`getOrgSeat` deliberately admits soft-deleted workspaces for the same reason — their accounts still exist, so the charge is still running.

---

## Verification

```
      Tests  7 passed (7)   [TeamEvents.test.ts]

$ npm run test:backend

$ npm run typecheck
```

<details>

Each emit disabled individually:

```
baseline                       0 failed | 11 passed
team.account.created           1 failed | 10 passed
team.account.disabled          1 failed | 10 passed
team.account.enabled           1 failed | 10 passed
team.deleted                   1 failed | 10 passed
team.account.deleted           1 failed | 10 passed
pre-delete seat capture        1 failed | 10 passed
held_bytes on the payload      2 failed |  9 passed
restored                       0 failed | 11 passed
```

Every site fails exactly one test.
</details>

## What changed since the original #3722

The held-bytes report and `TeamBillingService` are **gone**, not moved. Under the rescoped model a disabled seat is an ordinary suspended account holding files — not a team-specific case — so there is nothing team-shaped to measure. Whether the owner is charged for held data is now an open pricing question; prod can compute `SUM(size)` itself if we decide we want it.

`held_bytes` still rides on the disable and enable payloads, so prod has the figure at the transition without querying back.

See `TEAMS-BILLING-SPLIT.md` and PUT-1765.

---

Closes PUT-1712. PUT-1747 is superseded.

### DIFF
--- a/src/backend/clients/event/types.ts
+++ b/src/backend/clients/event/types.ts
@@ -36,6 +36,24 @@ type GuiEvent<R = Record<string, unknown>> = {
 type FsCreateEvent = { node: FSEntry; entry: FSEntry; uid: string };
 
 /**
+ * Who pays, and for which team. Deliberately no payment identity: the
+ * payer's row is alive at emit time, so a consumer resolves it when it acts --
+ * a snapshot taken here would be stale, and two events racing on it would each
+ * create their own customer.
+ */
+export type TeamBillingContext = {
+    team_uid: string;
+    owner_user_id: number;
+};
+
+/** A team event about one seat. */
+export type TeamBillingEvent = TeamBillingContext & {
+    user_id: number;
+    user_uuid: string;
+    username: string;
+};
+
+/**
  * Extension-augmentable half of {@link EventMap}. Extensions that emit their own
  * events declare the payload here by declaration merging, so both the emitter
  * and every listener are typed against the same shape:
@@ -334,6 +352,23 @@ export type EventMap = {
         user_id: number;
         user_uuid?: string;
         stripe_customer_id?: string | null;
+    };
+
+    // ---- Team billing ---- the trigger, never the charge ----
+    'team.account.created': TeamBillingEvent;
+    /** `held_bytes` is what to bill storage on while the seat is off. */
+    'team.account.disabled': TeamBillingEvent & { held_bytes: number };
+    'team.account.enabled': TeamBillingEvent & { held_bytes: number };
+    /** Emitted pre-delete: the membership row cascades away with the user. */
+    'team.account.deleted': TeamBillingEvent;
+    /** Per-seat charges stop; byte charges do not, the accounts remain. */
+    'team.deleted': TeamBillingContext & { account_count: number };
+    /** A budget line was crossed. Emitted on transition only, never per request. */
+    'metering.credit-state': {
+        user_uuid: string;
+        state: 'near-limit' | 'exhausted';
+        allowance_used: number;
+        month_usage_allowance: number;
     };
 
     // ---- Filesystem ----
@@ -758,11 +793,10 @@ export type EventKey = keyof EventMap & string;
 // Generates a wildcard for every non-final dot-separated prefix of K.
 export type WildcardPrefixes<K extends string> =
     K extends `${infer Head}.${infer Tail}`
-        ?
-              | `${Head}.*`
-              | (Tail extends `${string}.${string}`
-                    ? `${Head}.${WildcardPrefixes<Tail>}`
-                    : never)
+        ? | `${Head}.*`
+          | (Tail extends `${string}.${string}`
+                ? `${Head}.${WildcardPrefixes<Tail>}`
+                : never)
         : never;
 
 export type ListenKey = EventKey | WildcardPrefixes<EventKey>;
--- a/src/backend/services/team/TeamEvents.test.ts
+++ b/src/backend/services/team/TeamEvents.test.ts
@@ -1,0 +1,251 @@
+/**
+ * Copyright (C) 2024-present Puter Technologies Inc.
+ *
+ * This file is part of Puter.
+ *
+ * Puter is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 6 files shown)


jfcastro92:
> [!WARNING]
> **Superseded — the payloads below are out of date.**
>
> They show `stripe_customer_id`, which this PR no longer emits. It was dropped after review: the payer's row is alive at emit time, so a value captured here would be stale, and two seats provisioned concurrently would each see `null` and each create their own customer. A consumer resolves payment identity when it acts.
>
> Kept for history. See the re-validation comment below for the current shape.

---

jfcastro92:
## Local validation (re-run after the payload change)

Replaces the comment above, which showed a `stripe_customer_id` this PR no longer emits.

Engine: mysql 8 · `teams_enabled: true` · observed through a **real consumer**, not a log probe: the prod extension from HeyPuter/heyputer#1172, loaded into the local server with only its import alias adjusted. So this is the shape an actual listener receives.

### What a consumer now gets

```
[team-seat-billing] would start charging {
  teamUid: 'f8c785d4-35e3-4e10-947c-3c4a3c2b5d6f',
  userUuid: '65636cde-7b3e-4de0-aab4-06f9aa01a6a9',
  ownerUserId: 96
}
[team-seat-billing] would pause charging  { …, heldBytes: 0 }
[team-seat-billing] would resume charging { …, heldBytes: 0 }
[team-seat-billing] workspace closed {
  team_uid: 'f8c785d4-…', owner_user_id: 96, account_count: 1
}
```

```
stripe mentions across every captured payload: 0
```

`owner_user_id` is what identifies the payer; resolving them to a payment identity is the consumer's job, at the moment it acts.

### Why the field went

Not tidiness — it was a hazard. The value was a snapshot taken at emit time, and the payer's row is alive throughout (unlike `user.delete`, whose comment records the opposite: *"the row is already gone by emit time, so identifiers ride the payload"*).

Concretely: provision two seats in quick succession for an owner who has never paid. Both events capture `null`. A consumer handling them concurrently sees `null` twice and creates **two customers for one owner**, with the second seat's charge landing on a customer the first flow never learned about. Keeping the field did not prevent that — get-or-create has to be atomic in the consumer regardless — so it bought nothing and invited the race.

`TeamStore.getStripeCustomerId` went with it, which takes the last payment-vendor method out of the OSS teams path.

### Everything else is unchanged

The five events still fire at the same transitions, `held_bytes` is still computed at emit time, and workspace closure still fans out one `pause` per seat before `team.deleted` — verified in this run.

jfcastro92:

Salazareo (approved):
looks good but I think we should cache some of these store queries

jfcastro92:

| Before | After |
| --- | --- |

### Scoped, and checked

### One thing not renamed

jfcastro92:
> looks good but I think we should cache some of these store queries

Done, in #3735 — and it now covers more than my first reply said, including the two I had originally argued for leaving alone.

**Cached, all on a 60s TTL with explicit invalidation:**

| Read | Why it matters |
| --- | --- |
| `getByUid` | every team route resolves the team through it |
| `getMembership` / `isMember` | the authorization check on every team route |
| `listMemberIdsByGroupId` | share-notification fan-out |
| `getOrgSeat` | this PR's path — `captureSeatForBilling` on every account deletion |

I initially said I would leave `isMember` and `getByUid` uncached because they gate authorization and a stale entry means a removed member keeps access. That reasoning was about the *risk*, not about whether it could be done safely — so rather than skip the hottest two reads, the invalidation is now exhaustive and each path has its own falsified test.

**Every path that can change either value busts it:**

- `addMember` / `removeMember` → the membership pair
- `update` → the team row (name, handle, `directory_enabled` all live on it)
- `softDelete` → the row *and* every membership pair over it, read while the team is still live and bounded by the fan-out cap
- `cascadeDelete` → the pair, because `jct_user_group.user_id` is `ON DELETE CASCADE` and the membership changes without passing through the store at all

That last one is the reason this needed care: a TTL alone leaves a deleted account passing `isMember` for up to a minute. An existing test caught it the moment the cache went in.

**Falsified individually** — each bust removed on its own, and only its own case fails:

| Broke | Failed |
| --- | --- |
| pair bust on add/remove | 1 |
| row + pair bust on softDelete | 1 |
| row bust on update | 3 |
| pair bust on cascade | 1 |

**Still not cached, deliberately:** `getHeldBytes`, a `SUM` over `fsentries`. It is expensive, but it is read at emit time precisely to be current — caching it would bill a stale figure.

Backend **7897 passed | 26 skipped**, typecheck clean, all 13 commits typecheck individually.

PR #3737, opened 2026-09-04
🏗️ PUT-1745: enforce the forced password change on workspace seats

jfcastro92:
Closes PUT-1745. Stacked on #3735 (phase 4).

## The gate this ticket owns

`user.requires_password_change` shipped in PUT-1699 and has been **written but never enforced**. Provisioning sets it; `reissueCredential` reads it as the "has this seat activated" test. `gates.ts` checks `requires_email_confirmation`, `requires_phone_verification` and `requires_card_verification` — there was no fourth check.

Verified live during phase 2 validation, and it is worse than "the flag does nothing":

```
POST /send-confirm-email  -> 200        (a seat can self-serve this)
POST /confirm-email       -> 200
/whoami /teams /stat /mkdir -> all 200  with requires_password_change still 1
```

The flag was also **only ever written as `1`**, never cleared, anywhere in `src/backend`. So two things followed: the temporary password the administrator generated stayed valid forever, and `reissueCredential`'s "already activated" 409 was unreachable — an administrator could mint a working credential for any seat, indefinitely.

## What this adds

**The fourth clause in `assertVerifiedAccount`, and nowhere else.** Its docblock records that WebDAV once bypassed the phone and card gates by dispatching off a single `router.use`; a second implementation repeats a documented mistake.

**The clearing path**, because the gate alone would have locked every seat out permanently. `handleChangePassword` and the `pass_recovery_token` path both clear the flag and the expiry, and record an `activate` audit row.

**`allowUnconfirmed: true` on `/user-protected/change-password`** — the obvious trap is that a gate refusing everything also refuses the endpoint that clears it.

⚠ That flag is coarser than the ticket's wording: it disables the whole `requireVerifiedAccount` gate on that route, so a pending *phone* or *card* verification is also let through, not just email. Benign for the same reason — the caller is authenticated and proving their current password — but broader than described, and it changes a shared route.

**A 24-hour expiry** on the temporary password (`user.temp_password_expires_at`, three dialects: `0080` / `mysql_mig_35` / `postgres_mig_24`), so an unused reset dies rather than becoming a standing credential. Login refuses an expired one with `temporary_password_expired`.

**`POST /teams/:uid/members/:username/password-reset`** — 20/day, audited as `reset_member_password` with no credential in the row, drops the member's sessions, leaves 2FA alone. `reissueCredential` now writes an audit row too, and its 409 is reachable.

New limit documented in `rate-limits-and-quotas.md`, same change.

## Why this is on its own

It is the only phase 5 change that touches shared platform code, and it does so twice: `gates.ts` runs on **every authenticated request for every account on the platform**, and there is a three-dialect migration. Both want isolated review and an isolated revert. The member-facing half is the PR stacked on this one.

## Verification

Typecheck clean. Full backend suite **7765 passed | 26 skipped**.

Falsified — each piece broken, the failing tests confirmed, then restored:

| Broke | Failed |
| --- | --- |
| the gate clause in `assertVerifiedAccount` | 3 — exactly the gate tests |
| flag clearing in `handleChangePassword` | 3 |
| `allowUnconfirmed` on change-password | 3 |
| login expiry check | 1 |
| expiry write in `#issueTemporaryPassword` | 1 |
| reset audit append | 2 |
| reissue audit append | 1 |
| `recordPasswordSelfChange` | 1 |
| `isTemporaryPasswordExpired` | 3 |

One test deliberately passes when its neighbour breaks: dropping the expiry *write* does not fail the HTTP login-expiry test, because that test writes the expiry itself. That split is intended — one test proves the write, the other proves enforcement — and each fails when its own code is broken.

## Decisions worth a reviewer's eye

- **Reset drops the member's sessions.** Not in the ticket; judged correct for a credential change and consistent with `#suspend`. Easy to remove.
- **`reissue` and `reset` share the `reset_member_password` audit action** (reissue carries `reason: 'reissue'`). From the member's side both mean the administrator holds a working credential.
- **`requires_password_change` is `smallint` on postgres** while the other three `requires_*` flags are `boolean` — an inconsistency inherited from PUT-1699. It is therefore *not* in `USER_BOOLEAN_COLUMNS`; writes stay `0`/`1`. Worth a follow-up to align the column type.

## Not done here

The forced-change **screen** naming who reset it and when is GUI, so phase 6. The 403 body carries no actor or timestamp today; if the screen needs them, that is a follow-up on this route.

### DIFF
--- a/src/backend/clients/database/SqliteDatabaseClient.test.ts
+++ b/src/backend/clients/database/SqliteDatabaseClient.test.ts
@@ -27,7 +27,7 @@ import { DatabaseClientFactory } from './index.js';
 import { SqliteDatabaseClient } from './SqliteDatabaseClient.js';
 
 /** Highest schema version the migration table can reach. */
-const CURRENT_SCHEMA_VERSION = 77;
+const CURRENT_SCHEMA_VERSION = 78;
 
 /**
  * These suites migrate real files on disk. Idle they finish in well under a
--- a/src/backend/clients/database/SqliteDatabaseClient.ts
+++ b/src/backend/clients/database/SqliteDatabaseClient.ts
@@ -111,6 +111,7 @@ const AVAILABLE_MIGRATIONS: [number, string[]][] = [
     [74, ['0079_team-audit-and-group-shares.sql']],
     [75, ['0080_kv-share-handles.sql']],
     [76, ['0081_event-subscriptions-indexes.sql']],
+    [77, ['0082_temp-password-expiry.sql']],
 ];
 
 export class SqliteDatabaseClient extends AbstractDatabaseClient {
--- a/src/backend/clients/database/migrations/mysql/mysql_mig_37.sql
+++ b/src/backend/clients/database/migrations/mysql/mysql_mig_37.sql
@@ -1,0 +1,21 @@
+-- Copyright (C) 2024-present Puter Technologies Inc.
+--
+-- This file is part of Puter.
+--
+-- Puter is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU Affero General Public License as published
+-- by the Free Software Foundation, either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU Affero General Public License for more details.
+--
+-- You should have received a copy of the GNU Affero General Public License
+-- along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+-- See sqlite/0082_temp-password-expiry.sql for the column rationale.
+-- No per-file applied-state tracking, so the column goes through _puter_add_col.
+
+CALL _puter_add_col('user', 'temp_password_expires_at', '`temp_password_expires_at` bigint DEFAULT NULL');
--- a/src/backend/clients/database/migrations/postgres/postgres_mig_26.sql
+++ b/src/backend/clients/database/migrations/postgres/postgres_mig_26.sql
@@ -1,0 +1,21 @@
+-- Copyright (C) 2024-present Puter Technologies Inc.
+--
+-- This file is part of Puter.
+--
+-- Puter is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU Affero General Public License as published
+-- by the Free Software Foundation, either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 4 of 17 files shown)


jfcastro92:
## Local validation

Engine: mysql 8 · `teams_enabled: true` · a real workspace, a real provisioned seat, driven over HTTP.

### The migration applies

```
[mysql] 35 migration files applied
[mysql] applied mysql_mig_35.sql (1 statements)
user.temp_password_expires_at present: 1
```

Provisioning now writes both the flag and the expiry:

```
username   req_pw  has_expiry
p5s6740      1         1
```

### The gate — the phase 2 finding, closed

That finding was: a seat signs in on its temporary password, self-serves email confirmation, and reaches everything with `requires_password_change` still `1`.

⚠ **Testing this needs care, because the email gate masks the password gate.** Straight after provisioning, `/teams` refuses with *"Please confirm your email to continue"* — that is the *third* clause, not the new one. Reading that as a pass would prove nothing. `/whoami` also stays 200 throughout, because it opts out via `allowUnconfirmed`.

So the seat's email was confirmed first, leaving the new clause as the only thing standing:

```
POST /confirm-email -> {"email_confirmed":true}

/teams   403  {"error":"Please choose your own password to continue"}
/stat    403
/mkdir   403

code: password_change_required
```

Compare the same four routes in phase 2: `/whoami /teams /stat /mkdir` were **all 200**.

Sign-in itself still succeeds, which is the design — the member is in, and can do nothing until they choose a password.

### The endpoint that clears it is reachable through the gate

The trap this ticket names: a gate refusing everything would also refuse the way out.

```
POST /user-protected/change-password -> 200  "Password successfully updated."

username   req_pw  expiry_cleared
p5s6740      0          1
```

Both the flag and the expiry clear. The seat then has the product:

```
/whoami 200   /teams 200   /stat 200   /mkdir 200
```

### The 409 that was previously unreachable

Because the flag was never cleared, `reissueCredential`'s "already activated" branch could never be taken. Now:

```
POST /teams/:uid/members/:username/activation
  -> 409 {"error":"That account is already activated","code":"conflict"}
```

### The reset route, and its audit row

```
POST /teams/:uid/members/:username/password-reset
  -> 200 {"temporary_password": "<redacted len=16>"}
```

The audit trail, read from the table:

```
id  user_id_keep  action                  reason
35      100       reset_member_password   NULL
34      100       activate                NULL     <- written when the seat chose its own password
33      100       provision               NULL
```

No credential in any row, and `activate` records that the forced change actually happened — the state transition that had no representation before.

---

Full suite **7766 passed | 26 skipped**, typecheck clean.

Not exercised here: the 24-hour expiry lapsing in real time. `isTemporaryPasswordExpired` and the login refusal are unit-covered and falsified; waiting out the window live was not practical, and moving the clock would have tested the harness rather than the code.

jfcastro92:

jfcastro92:

| Before | After |
| --- | --- |

### Scoped, and checked

### One thing not renamed

Salazareo (approved):

reynaldichernando (approved):

PR #3738, opened 2026-09-04
🏗️ PUT-1746 + PUT-1733: tell a member what was done to their account

jfcastro92:
Closes PUT-1746 and PUT-1733. Stacked on #3737.

Two halves of the same question — *what did the workspace do to my account, and how do I find out.* One is pull, the other push. Squashed because they are one change to a reviewer: same audience, same purpose, seven files, overlapping only in `TeamService`.

## The activity view (PUT-1746)

`GET /teams/:uid/audit/me` already existed, was already user-scoped, and already 404'd a non-member. What was missing is the load-bearing half.

A member is told a reset happened — but a reset only matters alongside **who signed in afterwards**. Without that row the audit says a credential was issued and never says whether it was used, which is exactly the question the member has.

`SessionStore.listSignIns` (keyset, `kind='web'` only, revoked rows included) merges their own sign-ins into the same stream, newest first, with a per-stream cursor:

```
Password reset by your workspace administrator   Tue 14:02
Sign-in from 203.0.113.7 · Chrome/macOS          Tue 14:05   <- the tell
You changed your password                        Wed 09:31
```

That middle row is what makes administrative use of a reset visible to the person who cares. PUT-1745 calls this view load-bearing for exactly that reason.

## The notices (PUT-1733)

The two things a member cannot discover for themselves: their **account** being disabled, and their **workspace** being closed.

Deliberately one each. Closing a workspace disables every account in it, so sending both would tell one person twice about one event — the workspace notice covers the disabling.

Both say plainly that **nothing was deleted**: the accounts persist, suspended, holding their files and their usernames. Deleting a workspace is not a way to stop paying for the accounts in it; deleting the accounts is. Three near-identical send helpers were collapsed into one.

## Verification

Typecheck clean. Full backend suite **7765 passed | 26 skipped**.

Falsified — each piece broken, the failing tests confirmed, then restored:

| Broke | Failed |
| --- | --- |
| `listSignIns` returns nothing | 3 |
| `listSignIns` user scoping | 7 |
| `listSignIns` `kind='web'` filter | 1 |
| merged has-more calculation | 1 |
| `ip` / `user_agent` on sign-in entries | 2 |
| disable notice | 1 |
| workspace-closed notice | 1 |
| both notices on closure (the duplicate) | 1 |
| soft delete → hard `DELETE FROM group` | 2 |

The user-scoping one matters most: seven tests fail if `listSignIns` stops filtering by user, which is the difference between a member's own activity view and a leak of everyone's.

## One breaking change to flag

**`created_at` on both audit endpoints is now unix seconds.** The two sources had to sort on one axis, and the audit rows' type already varied by dialect — string on sqlite, `Date` on mysql and postgres — so there was no consistent existing shape to preserve.

It changes `GET /teams/:uid/audit` as well as `/audit/me`, so it is wider than this ticket strictly needed. Worth a reviewer's opinion on whether that is acceptable or whether the two endpoints should diverge.

## Not done here

- **"Stop the billing"** (PUT-1733 step 3) — the mechanism is the `team.deleted` / `team.account.*` events, which live in #3733, and prod's consumer is HeyPuter/heyputer#1172. Implementing it here would collide with that PR.
- **The confirmation-dialog wording** about storage charges is GUI, so phase 6 (PUT-1740).
- **"Account about to be hard-deleted"** belongs to PUT-1732.

### DIFF
--- a/src/backend/clients/email/templates.ts
+++ b/src/backend/clients/email/templates.ts
@@ -365,6 +365,33 @@ as you. Choose your own password promptly.</li>
 <p>Puter</p>
         `,
     },
+    team_account_disabled: {
+        subject: 'Your {{team_name}} account has been disabled',
+        html: `
+<p>Hi there,</p>
+<p>{{team_name}} has disabled your Puter account <b>{{username}}</b>. You can no
+longer sign in to it.</p>
+<p>Nothing has been deleted. Your files, apps and data are as you left them, and
+{{team_name}} can re-enable the account at any time. If you believe this is a
+mistake, ask them.</p>
+<p>Sincerely,</p>
+<p>Puter</p>
+        `,
+    },
+    team_closed: {
+        subject: '{{team_name}} has been closed',
+        html: `
+<p>Hi there,</p>
+<p>{{team_name}} has closed its Puter team, and your account
+<b>{{username}}</b> has been disabled along with it. You can no longer sign in to
+it.</p>
+<p>Nothing has been deleted. Your files and data are still there; closing the
+team does not destroy the accounts it created, and each one would have to be
+deleted on its own request.</p>
+<p>Sincerely,</p>
+<p>Puter</p>
+        `,
+    },
     team_password_reset: {
         subject: 'Your {{team_name}} account password was reset',
         html: `
--- a/src/backend/controllers/team/TeamController.http.test.ts
+++ b/src/backend/controllers/team/TeamController.http.test.ts
@@ -346,6 +346,17 @@ describe('team endpoints over HTTP', () => {
         expect(body.results[0].recipient).toBe(handle);
     });
 
+    it('gives a non-member 404 on the member view, not an empty page', async () => {
+        const { team } = await makeTeam();
+
+        const res = await call(
+            'GET',
+            `/teams/${team.uid}/audit/me`,
+            env.users.other.token,
+        );
+        expect(res.status).toBe(404);
+    });
+
     // -- the forced-change gate ---------------------------------------
 
     /**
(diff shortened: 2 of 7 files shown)


jfcastro92:
## Local validation — and a bug this found

Engine: mysql 8 · real workspace, real seat, real mail to mailhog.

### The run found a real defect, on mysql only

The merged stream came back like this:

```
05:54:28  activate
05:54:27  reset_member_password
05:52:21  provision
01:56:22  sign_in     <- happened AFTER the provision above, in real time
01:54:28  sign_in
```

Every audit row sat **exactly 4 hours ahead** of every sign-in. Both sources agree in the database — `UNIX_TIMESTAMP(created_at)` on the audit row and the session's stored integer are both `1788486868` — but the API returned `1788501268` for the audit row.

The cause is the driver, not the data:

```
mysql2 returns: Date -> Fri Sep 04 2026 01:54:28 GMT-0400 (Eastern Daylight Time)
getTime()/1000 = 1788501268        (true instant: 1788486868)
```

`audit_team_membership.created_at` is a DATETIME holding UTC; mysql2 builds a `Date` from it **in local time**, so `epochSeconds`' `value instanceof Date` branch inherits a four-hour shift. `sessions.created_at` is a plain integer and is unaffected.

**This destroys exactly what the ticket is for.** Every audit row sorts above every sign-in regardless of when either happened, so the "who signed in after the reset" row can never appear in its true position — the one thing the merge exists to show.

Not visible in unit tests: on sqlite the driver returns a string, `epochSeconds` parses it as UTC, and the answer is right. It only appears on a real mysql.

### The fix

Both streams now produce the same kind of value at the source — unix seconds, selected in SQL per dialect rather than converted after the driver has already misread it:

```sql
postgres: EXTRACT(EPOCH FROM `created_at`)::bigint
mysql:    UNIX_TIMESTAMP(`created_at`)
sqlite:   CAST(strftime('%s', `created_at`) AS INTEGER)
```

After, on the same data:

```
02:00:32  sign_in
01:56:22  sign_in
01:54:28  activate
01:54:28  sign_in
01:54:27  reset_member_password
01:54:27  sign_in                  <- the tell, in its true position
01:53:46  reset_member_password
01:53:33  sign_in
01:53:19  activate
```

Interleaved correctly, and a reset is now immediately followed by the sign-in that used it.

Regression test at the store level, falsified — reverting to selecting the raw datetime fails it and nothing else:

```
× returns audit timestamps as unix seconds, not a dialect datetime
Tests  1 failed | 30 passed (31)
```

It asserts the *type*, which is the dialect-independent invariant; asserting the offset would only fail on mysql.

### The notices

```
disable a seat      -> "Your P5 account has been disabled"
close the workspace -> "P5 has been closed"
```

Counted per address, closure adds exactly one:

```
1 x P5 has been closed
1 x Your P5 account has been disabled
```

Closing a workspace disables every account in it, so a naive implementation tells one person twice about one event. It does not — the workspace notice stands in for the disable notice.

---

Full suite **7766 passed | 26 skipped**, typecheck clean.

⚠ Related, and worth a reviewer's judgement: this change makes `created_at` unix seconds on **both** audit endpoints, not just `/audit/me`. The mysql shift above means the existing `/teams/:uid/audit` was already returning a wrong instant on that engine, so this corrects it — but it is still a shape change to an endpoint this ticket did not otherwise touch.

jfcastro92:

jfcastro92:
**Terminology: "master account" is now "owner account"** throughout the workspace surface — docs, SDK JSDoc, GUI strings and backend comments. 24 files, no behavior change.

`not_the_workspace_owner` was already the error code, so the prose now matches the API rather than contradicting it.

Salazareo (approved):

jfcastro92:

| Before | After |
| --- | --- |

### Scoped, and checked

### One thing not renamed

PR #3740, opened 2026-09-04
🏗️ PUT-1732: delete a workspace seat for good, once it is disabled

jfcastro92:
Closes PUT-1732. Stacked on #3738.

## The gap this closes

Phase 5's scope named hard delete, and it was the one piece missing. `cascadeDelete` existed but appeared in the team path only as a *provisioning rollback* — there was no way for a workspace owner to remove a seat they no longer want. A disabled account persisted forever with no route out.

## The ordering is the substance

Deletion is refused unless the account is already disabled:

```
DELETE /teams/:uid/members/:username
  -> 409 {"code": "account_must_be_disabled_first"}
```

That puts a reversible step in front of the only irreversible operation in the feature, and means deletion cannot be reached by a single action on a live account. Disable first, then delete.

**The audit row is written before `cascadeDelete` runs.** The FKs are `ON DELETE SET NULL` and the `_keep` columns carry the identifiers, so the record of what was done survives the account it names. Written after, it would name nothing.

## The billing emit that is deliberately absent

`deleteMember` does **not** emit `team.account.deleted`. `UserAccountService.cascadeDelete` already captures the seat before the delete and emits it after — so a second emit here would close the storage charge twice.

That was worth checking rather than assuming: the natural-looking implementation is `captureSeatForBilling` → `cascadeDelete` → `emitSeatDeleted`, which is exactly what the phase 3 wiring already does one layer down. The test asserts the count is 1, so a future reader who adds the "missing" emit gets told.

Disabling closed the per-account charge; this closes the storage one, and it is the only thing that does.

## No restore window

Considered and rejected, for three reasons: the reversible step already exists earlier at disable; a disabled account costs only the bytes it holds, so nothing pressures a hasty delete; and a restore promise means retaining data the workspace has explicitly asked to be rid of. Delete means delete. There is no retention timer either — with no file transfer, the account still holds the member's work, so a clock would destroy real data on a schedule nobody revisited.

Published in `rate-limits-and-quotas.md` next to the workspace-deletion note, because the two are easy to confuse and only one of them frees a seat.

## Verification

Typecheck clean. Full backend suite **7887 passed | 26 skipped**.

Falsified — each piece broken, the failing test confirmed, then restored:

| Broke | Failed |
| --- | --- |
| the disable gate | 1 — `refuses to delete a live account` |
| the audit append | 1 — `keeps the audit trail attributable after the account is gone` |

Four tests total: the two above, one asserting the account is really gone and the deleted event fires exactly once, and an HTTP test covering the route wiring end to end (409 on a live account, 200 after disabling, row gone).

## Not done here

The confirmation dialog — PUT-1732 step 2 says "separately confirmed" — is GUI, so it lands in phase 6 with `TabTeams` (PUT-1740). The endpoint is unconditional today; the confirmation is a client-side gate, which is the right place for it but does mean the API will delete on a single call until that ships.

### DIFF
--- a/src/backend/controllers/team/TeamController.http.test.ts
+++ b/src/backend/controllers/team/TeamController.http.test.ts
@@ -153,6 +153,25 @@ describe('team endpoints over HTTP', () => {
         expect(res.status).toBe(404);
     });
 
+    it('deletes a seat over HTTP only once it is disabled', async () => {
+        const { team, memberUsername } = await makeTeam();
+        const path = `/teams/${team.uid}/members/${memberUsername}`;
+
+        const live = await call('DELETE', path, env.users.user.token);
+        expect(live.status).toBe(409);
+        expect(await live.json()).toMatchObject({
+            code: 'account_must_be_disabled_first',
+        });
+
+        await call('POST', `${path}/disable`, env.users.user.token);
+        const gone = await call('DELETE', path, env.users.user.token);
+
+        expect(gone.status).toBe(200);
+        expect(
+            await env.server.stores.user.getByUsername(memberUsername),
+        ).toBeFalsy();
+    });
+
     it('lists members with org_owned distinguishing the owner', async () => {
         const { team, memberUsername } = await makeTeam();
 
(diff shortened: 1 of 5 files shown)


Salazareo (approved):

jfcastro92:

| Before | After |
| --- | --- |

### Scoped, and checked

### One thing not renamed

reynaldichernando (approved):

PR #3741, opened 2026-09-04
🏗️ PUT-1735 + PUT-1737 + PUT-1738 + PUT-1739: the `puter.teams` SDK module, suite and docs

jfcastro92:
Closes PUT-1735, PUT-1737, PUT-1738 and PUT-1739. Stacked on #3740.

Four tickets, one surface: the module, the suite that exercises it, the pages that document it, and the limits those pages publish. Splitting them would mean shipping a module with no docs, then docs describing a module already merged.

## Why routes and not a driver

Workspace administration reaches the backend through `/teams` rather than a driver interface, because the gates it needs are route-level: a user actor, a verified account, and a dual-window rate limit. A driver has none of those.

## Layout

Follows `apps/` and `perms/` — one file per method, a thin `index.js`, a JSDoc-only `types.js`, and the `METHODS` rebinding so a destructured method keeps its `this`.

It deliberately does **not** follow `perms/lib/req.js`. Those endpoints resolve `{ error: true }` for backward compatibility; nothing here has callers to keep compatible, so this throws `PuterJSError` carrying the backend's own code.

## Two decisions worth a reviewer's eye

**Every method takes a `uid`, never a handle.** A handle is a mutable label that deleting the workspace releases, so a stored handle can later resolve to a *different* workspace. Accepting one would make that a silent wrong-target bug rather than a 404.

**The list methods refuse `offset`.** They offer the three forms `puter.apps.list()` does — array by default, page envelope under `cursor`/`includeTotal`, async iterator under `stream`. These routes are keyset-only, so an accepted `offset` would return page one however far you asked to skip. Refusing beats lying.

## Scope

The surface covers the routes that exist today. Usage totals and member-email correction have no backend route yet and are absent rather than shipped as methods that 404.

`deleteMember()` **is** here — the route it needs lands in #3740, directly below this. Worth flagging: this PR is unmergeable ahead of that one, by design.

## Docs

`Teams.md` plus 14 method pages, each carrying the beta banner in the same position `Events/` puts it — on every page, not only the index, since a reader arriving from search lands on a method page.

Limits published in `rate-limits-and-quotas.md` (PUT-1739), including that both buckets are per account rather than per workspace, so administering several workspaces spends one budget.

## Verification

Typecheck clean. `check:puterjs:types` — declarations generate and type-check cleanly. Backend suite **7887 passed | 26 skipped**. SDK unit tests **23 passed**, and every commit in the stack typechecks on its own rather than just the tips.

`teams.suite.ts` (PUT-1737) is registered in `suites/index.ts` and gated on the teams capability, so it skips where a deployment has teams off instead of failing.

⚠ `npm run check-translations` could not be run: the script points at `tools/check-translations.js`, which does not exist on `main` either. Pre-existing, unrelated to this PR, and worth its own ticket.


jfcastro92:
**Terminology: "master account" is now "owner account"** throughout the workspace surface — docs, SDK JSDoc, GUI strings and backend comments. 24 files, no behavior change.

`not_the_workspace_owner` was already the error code, so the prose now matches the API rather than contradicting it.

jfcastro92:

| Before | After |
| --- | --- |

### Scoped, and checked

### One thing not renamed

jfcastro92:
## Local validation — SDK, suite and docs

Engine **mysql 8**, a real local instance, current tip. Nothing stubbed but the SDK origin in the browser tests.

### The CI failure this PR had, and why it was mine

Every `teams` case in the browser runner failed with `PuterJSError: You may own 1 team`. The harness enables `teams_enabled` but never raised the cap, and the suite creates a team per case against one shared account — so every case after the first failed on `team_limit_reached` rather than on what it tests.

That is a collision between this PR and the cap in #3725, and it only appears when the two are stacked. Fixed in `harness/capabilities.ts`:

```ts
// The shipped default is 1, and the suite creates a team per case
// against one shared account, so every case after the first would
// fail on `team_limit_reached` rather than on what it tests.
max_teams_per_user: 100,
```

⚠ I had not run `npm run test:puterjs` locally before pushing — only the unit tests. That is the gap that let it through, and it is now part of the phase-6 checklist rather than something I remember to do.

**After the fix, run locally:**

```
npm run test:puterjs:node
Test Files  1 passed (1)
     Tests  488 passed | 69 skipped (557)
```

### The API surface, exercised end to end

A 40-check transcript against the running server covers every route this SDK wraps. Extract:

```
=== 1. Team lifecycle ===
  PASS  create a team — uid=ce964c0e-…
  PASS  directory is off by default
  PASS  read it back by uid
  PASS  rename it
=== 2. Seats ===
  PASS  provision a seat — credential returned once
  PASS  seat appears with org_owned
  PASS  a taken username is refused — 409 username_already_in_use
=== 3. Authority boundaries ===
  PASS  non-member gets 404, not 403 — got 404
  PASS  non-member cannot provision
  PASS  the owner is not a valid member-route target

40 passed, 0 failed
```

### Two things the run taught me about the real flows

**A provisioned seat has no `email_confirm_code`.** The administrator made the account, so no confirmation mail was ever sent. The activation order is therefore forced: sign in on the temporary password → **change password first** (the only route the gate admits) → *then* confirm the address. My first harness did it the other way and got nowhere, which is a fair proxy for what an integrator will hit.

**`/user-protected/change-password` is session-cookie only** and rejects a bearer token outright — so it is not scriptable with an API token at all. Worth knowing for anyone automating seat onboarding; the SDK method inherits this.

Neither is a defect, but neither is obvious from the docs, and both cost me a real debugging loop.

### Verification

Typecheck clean · `check:puterjs:types` clean · backend **7897 passed | 26 skipped** · GUI+SDK unit **275 passed** · all 13 commits typecheck individually.

Salazareo (approved):

reynaldichernando (approved):

PR #3742, opened 2026-09-04
🏗️ PUT-1740 + PUT-1762: the workspace surfaces in the GUI

jfcastro92:
Closes PUT-1740 and PUT-1762. Stacked on #3741.

The two places a workspace becomes visible to a person rather than an API: the Dashboard tab that administers it, and the share dialog that shares with it.

**Squashed because they are one change to a reviewer.** Both are the first GUI consumers of `puter.teams`, both add strings to the same `en.js`, and both had to answer the same question — what a workspace looks like to someone who has never seen one. Split, that answer gets reviewed twice and the same translation conflict resolved twice.

## TabTeams (PUT-1740)

Registered in `builtinTabs` after `TabFiles`, as a plain object with `html()` and `init($el_window)` like every other tab. The member table carries the operations the API allows and nothing it does not: provision, resend an activation, disable, restore, delete.

**Deleting asks for confirmation naming the account.** PUT-1732 puts the reversible step at *disable*, not in front of the request — the endpoint deletes on a single call once the account is suspended. So the client is where the confirmation belongs, and #3740 says as much in its own "not done here".

**The wording names the money.** Disabling stops the per-account charge and keeps the storage one; deleting is the only thing that ends it. This tab is the one place a person chooses between the two, and leaving them to infer the difference costs them real money — so it is stated, not implied.

Rendering and the billing arithmetic are pure functions in `teamsConsole.js` with tests beside them; the tab file is only the DOM.

## Share dialog (PUT-1762)

`UIWindowShare` gains the workspaces the caller owns as recipients.

**A workspace is one entry, not its members expanded.** Sharing with a workspace reaches every member through the group grant, so listing people would understate what the grant actually does — and would go stale the moment membership changed.

`shareWorkspaces.js` holds the resolution and deduplication, so the dialog does not grow a second source of truth about who a recipient is.

## Verification

Typecheck clean. `check:puterjs:types` clean. Backend suite **7887 passed | 26 skipped**. GUI and SDK unit tests **66 passed** across 5 files, covering the console's rendering and billing arithmetic, the tab, and the workspace resolution helper.

⚠ `npm run check-translations` could not be run — the script points at `tools/check-translations.js`, which does not exist on `main` either. Pre-existing and unrelated, but it does mean the 68 new `en.js` keys have not been machine-checked; worth an extra look at those in review.

## Not done here

Translations beyond `en.js`. Every string goes through `i18n()` with keys added to `en.js` only, which is the established pattern — other locales fall back until translated.


Salazareo (commented):
we should probably hide these till feature is ready, we could have it with a gui_params flag

jfcastro92:
## Translations, addressed

Two questions here: whether the missing checker matters, and whether the keys are actually correct. The second turned up a real bug.

### The checker is dead, not broken

`npm run check-translations` points at `tools/check-translations.js`, deleted in #2873 ("Remove gross legacy tools and docs"). The npm script was left behind. Nothing to restore — it is a stale entry on `main`, not something this PR broke.

### `en.js` only is correct, and I confirmed why

`i18n.js` resolves locale → `en` → **the raw key**:

```js
let language = translations[window.locale] ?? translations['en'];
let str = language.dictionary[key] ?? translations['en'].dictionary[key];
if ( ! str ) { str = key; }
```

So other locales fall back to English cleanly, and adding keys to `en.js` alone is the established pattern rather than an omission. What is *not* safe is a key that no dictionary has — that renders as `teams_member_state_active` to the user.

Checked by hand since the tool is gone:

```
i18n() keys called by the phase 6 GUI:  67
missing from en.js:                      0
added to en.js but never called:         0
```

(The first pass flagged 10 as unused; all 10 are reached through ternaries inside `i18n(...)` or the `AUDIT_ACTION_KEYS` lookup, which a literal-argument grep does not see. No dead keys.)

### The bug this found

`AUDIT_ACTION_KEYS` covered **4 of the 7** actions `TeamService` writes. Missing: `reset_member_password` and `activate` (added by PUT-1745) and `delete_account` (added by PUT-1732, #3740).

The fallback is `html_encode(entry.action)`, so those rows rendered the raw enum — `reset_member_password` — in the one view whose entire purpose is being readable, and on the two most consequential actions a workspace can take. Not a crash, which is why nothing caught it.

Three keys and three map entries added, plus a test over the whole vocabulary so the next action added to `TeamService` cannot silently drift out again:

```js
it.each([
    'provision', 'disable', 'enable', 'delete_team',
    'reset_member_password', 'activate', 'delete_account',
])('has a label for %s', (action) => {
    expect(auditActionKey(action)).not.toBeNull();
});
```

Falsified — dropping `delete_account` from the map fails exactly that case:

```
× has a label for delete_account
Tests  1 failed | 20 passed (21)
```

⚠ The test names the vocabulary as a literal list rather than importing it from the backend, since a GUI test reaching into `src/backend` would cross layers. It locks in today's completeness; an eighth action would still need adding in both places.

GUI and SDK tests now **74 passed**, backend **7887 passed | 26 skipped**, typecheck and `check:puterjs:types` clean.

jfcastro92:

| Before | After |
| --- | --- |

### Scoped, and checked

### One thing not renamed

jfcastro92:
> we should probably hide these till feature is ready, we could have it with a gui_params flag

Done — `gui_params.teams_ui`, off by default.

```ts
// PuterHomepageService
teams_ui: this.config.gui_params?.teams_ui === true,
```

Two entry points read it, because there are two places the feature becomes visible:

- **`TabTeams`** returns `unavailable` and hides the tab, the same state a deployment with no `/teams` route already produces. Nothing new to reason about.
- **`teams_for_sharing()`** returns `[]` before it calls the SDK, so no team recipients appear in the share dialog.

This decouples the API from the interface: `teams_enabled` can be on for the backend and the SDK while the GUI stays dark.

⚠ One thing to know: the flag is only read at those two points, so it hides the UI, it does not disable the API. Anyone who can call `puter.teams.*` still can. That is the intent — but it is a UI gate, not a kill switch, and worth saying out loud rather than leaving to be assumed.

The test for it was **vacuous on the first attempt** — it passed with the gate removed, because no SDK mock was set and the `catch` returned `[]` regardless. It now installs a `list` that would return a team and asserts it is never called:

```
× offers nothing while the teams interface is turned off
Tests  1 failed | 11 passed (12)
```

jfcastro92:
## Local validation — the GUI, driven with Playwright

A real Chromium against the running instance. Real accounts, real provisioning, real SDK (routed from `js.puter.com` to the local `dist`). **13 checks, 0 failures**, five screenshots.

```
=== 8. GUI — the admin console ===
  PASS  the Teams tab is present for an owner
  PASS  the team card names the team
  PASS  the member table lists the provisioned seat — skb667dm
  PASS  the directory panel renders
  PASS  delete is absent for a live seat — buttons=0
  PASS  and present once it is disabled — buttons=1
  PASS  the directory toggle is rendered
  PASS  turning it on asks first, naming what apps will read

=== 9. GUI — the member's side ===
  PASS  the member page actually rendered — 521 chars
  PASS  a member sees no team administration
  PASS  and no destructive controls
```

### The delete gate, observed rather than asserted

`01` shows the console for an owner: team card with handle, Directory panel, add-account form, the member table, and the Record. The seat row is **Active** and its actions are `Reissue password` / `Suspend` — no delete.

`02` is the same page after suspending: the row greys out, `State` reads **Suspended**, and `Restore` / **Delete account** appear. That is PUT-1740's "the button does not exist for a live account" holding in the browser, not just in the unit test.

### Two defects the screenshots caught that no test would have

Both were invisible to assertions and obvious on sight:

1. **`You're billed for 1 active accounts`** — the billing summary had no plural agreement, and `1` is the common case for a small team. Split into `teams_billing_summary_one`.
2. **The directory confirmation's primary button was styled `danger` (red).** Opening a directory is reversible, and the dialog itself says *"You can turn this off again at any time"* — red overstated it and put it in the same visual class as `Delete account`. Now `primary`; the `confirm()` helper takes the button kind, and delete keeps `danger`.

Fixed in this PR, GUI rebuilt, and the run repeated green.

### A vacuous test I wrote and had to fix

The member-side checks are all *absence* assertions — "sees no administration". On the first run the member page was 401ing, so every one of them passed against a blank page. They now assert the page actually rendered first:

```js
ok(memberBody.length > 200 && /Teams|Dashboard|Settings/i.test(memberBody),
    'the member page actually rendered', `${memberBody.length} chars`);
```

⚠ Same class of problem in the `teams_ui` flag check: setting `window.teams_ui = false` in an init script proves nothing, because the GUI boot assigns it from the server payload and overwrites it. The browser check now only asserts the flag *reaches* the client; the off-rendering is covered by the falsified unit test in this PR.

### Screenshots

`01-admin-teams-tab` · `02-admin-delete-gated-on-disable` · `03-admin-directory-confirm` · `04-member-view` · `05-flag-off`
01-admin-teams-tab
02-admin-delete-gated-on-disable
03-admin-directory-confirm
04-member-view



They are PNGs on my machine — GitHub's comment API cannot upload images, so say the word and I will drag them in.

### Verification

Typecheck clean · `check:puterjs:types` clean · backend **7897 passed | 26 skipped** · GUI+SDK unit **275 passed** · all 13 commits typecheck individually.

Salazareo (approved):

PR #3743, opened 2026-09-04
fix(ai): make chat fallback reach streamed Claude calls and rank Azure explicitly

404oops:
Validates and tightens the AI chat fallback chain ([PUT-1531](https://linear.app/puter/issue/PUT-1531/validate-ai-chat-fallback-logic)). The loop itself was working: capped at three attempts, never repeats a provider-and-model pair, re-runs the credit gate per attempt, stops on moderation refusals. Reviewing it turned up one real hole and two places where the intended order was not what actually ran.

## What changed

- **Claude streaming now falls back.** `ClaudeProvider` opened the upstream stream inside the populator, after the driver had already returned a chunked 200. An Anthropic overload, rate limit, or 5xx on a streamed call therefore skipped the fallback loop, never marked the route unhealthy, and reached the client as an error frame on a successful response. The provider now opens the stream and awaits its connection before returning the populator, so those failures throw from `complete()` like every other provider's do. The event iterator is created before the await so nothing the SDK emits early is dropped, and uploaded files are cleaned up on the failure path.
- **OpenAI-compatible routes no longer hard-pin a provider by default.** `/openai/v1/chat/completions` and `/openai/v1/completions` always sent `provider: 'openai-completion'` when the caller omitted one. The driver honours a pin even when that route is marked unhealthy, so those routes went OpenAI-first with Azure as fallback and ignored health marks on the first hop, while puter.js callers went Azure-first. They now forward `provider` only when the caller sent one and get the same preferred healthy route puter.js gets. A request that names no model at all stays pinned to OpenAI, so its default model does not move to Azure's.
- **Azure preference is explicit.** Azure beat OpenAI and xAI only because their cost tables are copies and Azure registered first, so a stable sort kept it ahead. A one-cent catalog drift would have flipped the served route silently. `modelRouting` now has a preferred tier for the two Azure providers, and the registration comment points at it instead of relying on order.
- **Removed Together's synthetic `model-fallback-test-1`.** It was a single-route bucket that threw a plain Error, so it produced a 500, marked its route unhealthy for ten minutes, and fired an alarm, and it was callable by any user despite being hidden from listings.
- Stale test name that called Claude the default provider now says azure-openai.

## Reviewer notes

- The Azure tier means Azure serves first even if its copied cost table ever drifts above the vendor's. I took the existing registration comment ("Azure is preferred for us") as the policy. Say so if you'd rather keep price as the deciding factor and only fix the tie-break.
- Behaviour change for OpenAI-compat callers who did not set `provider`: GPT and Grok models now go through Azure first, matching puter.js. Callers who set `provider` explicitly are unaffected.
- Those same calls now record `intended_service: ''` on `ai.prompt.validate`, `ai.prompt.complete` and `ai.prompt.cost-calculated`, where they recorded `openai-completion` — the same blank puter.js callers already produce, since the field carries the caller's pin rather than the route actually used. Nothing in this repo branches on it, but the billing and abuse extensions live in the consuming repo: anything grouping or filtering on `intended_service` will see OpenAI-compat traffic move into the unset bucket.
- Left out on purpose: GLM 5.2 sits in two buckets (`glm-5.2` on ZAI/Alibaba, `glm-5-2` on Mistral/BytePlus) so they cannot fall back to each other. Merging them would move existing callers of the dashed ids from Mistral's native response shape to ZAI's, which is a compatibility call for a maintainer. Also left as-is: Infron, Neuralwatt, and Hoonify still order by price within their tier; encoding a reliability order needs the team to state one, and the rank function is now the obvious place for it.

## Review follow-ups (second commit)

- **A stream error can reach the populator too late to be thrown.** The SDK hands an error only to an event reader already waiting on it; one that lands between the connect and the populator's first pull left the iterator reporting a plain end-of-stream, so a truncated response was metered and reported as a success. The loop now rethrows when the stream is `errored`.
- **A refused stream no longer strands its uploads.** `processPuterPathUploads` rewrites the caller's message parts in place, and the driver hands those same objects to the next route — which cannot resolve file ids we just deleted. It now returns a `restore()` that both failure paths call, so the fallback sees the original `puter_path` again.
- `/openai/v1/responses` and `/anthropic/v1/messages` keep their pins, with a comment saying why: each translates one provider's native wire shape by hand, so the preferred-route logic must not swap it out.
- `PREFERRED_PROVIDERS` is unexported and its doc now states the rank is unconditional against every other direct provider, not just the vendors Azure fronts. The duplicated hidden-model list is one constant.

## Testing

- New tests: Claude `complete()` rejects with the upstream error when a stream is refused; the OpenAI-compat routes leave `provider` unset for a named model, forward an explicit one, and stay pinned with no model; Azure wins over the vendor in both registration orders and with a higher cost table; a 400 leaves a route in rotation while a 503 marks it.
- Second commit adds: a swallowed stream failure surfaces from the populator instead of ending clean; a refused stream deletes its uploads and hands back the `puter_path`; `restore()` undoes both the file-id rewrite and the inline text error. Each was checked to fail without its fix.
- `vitest run` over `src/backend/drivers/ai-chat` and `src/backend/controllers/puterai`: 45 files, 1066 tests passing (18 integration files skipped without keys).
- `node tools/typecheck.mjs`: no new errors — 31 known baselined, matching `main` exactly.

### DIFF
--- a/src/backend/drivers/ai-chat/providers/claude/fileUpload.ts
+++ b/src/backend/drivers/ai-chat/providers/claude/fileUpload.ts
@@ -39,6 +39,12 @@ interface ContentPart {
 export interface ClaudeUploadResult {
     /** File IDs uploaded this request; caller deletes them after completion. */
     fileIds: string[];
+    /**
+     * Puts every rewritten part back to the `puter_path` it arrived as. Call it
+     * when the request fails: the driver hands the same message objects to the
+     * next fallback route, which cannot resolve our `file_id`s.
+     */
+    restore: () => void;
 }
 
 /**
@@ -66,15 +72,29 @@ export async function processPuterPathUploads(
             if (part?.puter_path) parts.push(part);
         }
     }
-    if (parts.length === 0) return { fileIds: [] };
+    if (parts.length === 0) return { fileIds: [], restore: () => {} };
 
     const fileIds: string[] = [];
+    const restores: Array<() => void> = [];
     await Promise.all(
         parts.map((part) =>
-            processPart(part, anthropic, stores, fsService, actor, fileIds),
+            processPart(
+                part,
+                anthropic,
+                stores,
+                fsService,
+                actor,
+                fileIds,
+                restores,
+            ),
         ),
     );
-    return { fileIds };
+    return {
+        fileIds,
+        restore: () => {
+            for (const undo of restores) undo();
+        },
+    };
 }
 
 async function processPart(
@@ -84,8 +104,16 @@ async function processPart(
     fsService: FSService,
     actor: Actor | undefined,
     fileIds: string[],
+    restores: Array<() => void>,
 ): Promise<void> {
     const path = part.puter_path!;
+    const original = { ...part };
+    restores.push(() => {
+        for (const key of Object.keys(part)) {
+            delete (part as Record<string, unknown>)[key];
+        }
+        Object.assign(part, original);
+    });
     delete part.puter_path;
 
     if (!actor?.user?.id) {
--- a/src/backend/drivers/ai-chat/providers/claude/fileUpload.ts
+++ b/src/backend/drivers/ai-chat/providers/claude/fileUpload.ts
@@ -107,12 +107,15 @@ async function processPart(
     restores: Array<() => void>,
 ): Promise<void> {
     const path = part.puter_path!;
-    const original = { ...part };
+    // Only the path needs to come back: `type`/`source` on success and
+    // `type`/`text` on failure are the sole fields written below, and the
+    // next provider's uploader keys off `puter_path` alone. Undoing those by
+    // name rather than snapshotting the part keeps nothing else alive.
     restores.push(() => {
-        for (const key of Object.keys(part)) {
-            delete (part as Record<string, unknown>)[key];
-        }
-        Object.assign(part, original);
+        delete part.type;
+        delete part.text;
+        delete part.source;
+        part.puter_path = path;
     });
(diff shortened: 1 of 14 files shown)


src/backend/drivers/ai-chat/providers/claude/fileUpload.ts
```diff
@@ -84,8 +104,16 @@ async function processPart(
     fsService: FSService,
     actor: Actor | undefined,
     fileIds: string[],
+    restores: Array<() => void>,
 ): Promise<void> {
     const path = part.puter_path!;
+    const original = { ...part };
```
Salazareo: what is this storing? if its file data this might lead to memory issues

Salazareo (approved):
one question there, rest looks fine

src/backend/drivers/ai-chat/providers/claude/fileUpload.ts
404oops: Not file data — the bytes go into a local `loaded.buffer` and straight to `toFile()`, and `part.source` only ever holds `{ type: 'file', file_id }`. It was a shallow copy of the content part's own fields.

That said, it did keep whatever the caller put on the part alive until the request ended, which matters if someone sends `puter_path` alongside a big inline `source`/`text` — overwriting that field used to make the old value garbage right away. Dropped the snapshot in 197eddee: the only fields this function writes are `type`/`source` (success) and `type`/`text` (failure), and the fallback uploader keys off `puter_path` alone, so restore just undoes those three by name.

PR #3745, opened 2026-09-04
🏗️ PUT-1736: a team directory apps can read, once the team opens it

jfcastro92:
Closes PUT-1736. Stacked on #3742.

## The delta is one flag, not a new disclosure

The ticket reads as "should colleagues be visible to apps," but `/teams/:uid/members` already has `requireUserActor: true` and nothing more — **any member can already enumerate their colleagues today**. The gate is the actor type, not the data.

So the only thing this adds is admitting an **app-under-user** actor to the same names. That is the whole risk, and it is what the setting governs.

## The open question, resolved

The ticket said *resolve before building*. Chosen: **a per-team setting, default off**.

Rejected alternatives, and why:

- **Per-app consent** — there is no existing consent surface to reuse, so it means building one for a single feature. It also asks members a question they cannot answer well ("should Calendar see your colleagues?") and gets clicked through.
- **Open by default** — the only option that cannot be walked back. Once apps depend on the directory, turning it off is a breaking change.

⚠ Worth a reviewer's eye: the names disclosed belong to the **members**, and the **owner** flips the switch. What makes that defensible is that the payload is strictly what colleagues already see, it is off unless deliberately turned on, and turning it on is audited. If that is not enough, the escalation is a member-level opt-out on top — deliberately not built, because a setting nobody finds is worse than no setting.

## What bounds an app

- **`directory_enabled` defaults to 0.** A team that has not opted in answers **404, not 403** — whether a team has this on is not something an app should be able to probe for either.
- **The membership tested is the person's, never the app's.** An app installed by a member of team A can never read team B.
- **The page carries `username` and `uuid` and nothing else.** No email, activation state, usage or role — those stay on the admin routes.
- **Suspended and never-activated accounts are excluded.**

## The bug this nearly shipped with

The obvious test for "never activated" is `password IS NOT NULL`. That is wrong: a provisioned seat holds its **temporary** password from birth, so the filter would have admitted exactly the accounts meant to be excluded. Activation is `requires_password_change` clearing.

I only caught it because the HTTP test asserted the seat was absent and it wasn't. There is now a test that walks a seat through provisioned → activated → suspended and asserts directory membership at each step.

## The toggle

In `TabTeams`, next to the team card. Turning it **on** asks for confirmation naming what apps will be able to read; turning it **off** does not — one grants access, the other only takes it away. A refused confirmation repaints so the checkbox does not sit checked against the truth.

Both transitions write an audit row (`directory_enabled` / `directory_disabled`), and setting it to the value it already has records nothing.

## Verification

Typecheck clean, `check:puterjs:types` clean. Backend **7892 passed | 26 skipped**, GUI and SDK **79 passed**. All 13 commits in the stack typecheck individually.

Falsified — each gate broken, the failing test confirmed, then restored:

| Broke | Failed |
| --- | --- |
| the opt-in check | 2 — the service test and the HTTP one |
| the membership requirement | 1 — an outsider reads the directory |
| the exclusion filter | 1 — suspended and unactivated appear |

## Migration

`0082_team-directory.sql` / `mysql_mig_37` / `postgres_mig_26`, adding `group.directory_enabled` with a default of 0. Idempotent on all three, since mysql has no per-file applied-state tracking.


jfcastro92:
## Local validation — the directory, API and GUI

Engine **mysql 8**, live instance. The migration applied and the column is real:

```
group.directory_enabled: tinyint(1)
[mysql] applied mysql_mig_37.sql (1 statements)
```

### Every gate, against the running server

```
=== 5. The directory (PUT-1736) ===
  PASS  closed by default, and 404 not 403 — got 404
  PASS  the owner can open it
  PASS  it lists once open — members=["dbgm3z6tn","sztrwqpz"]
  PASS  an activated seat is in it
  PASS  it carries username and uuid only
        {"username":"dbgm3z6tn","uuid":"d9dd1e68-9712-4927-b1cf-847d2bacfc48"}
  PASS  a non-member cannot read it even when open
  PASS  a suspended account drops out of it
```

The payload line is the one worth reading twice: the response carries `username` and `uuid` and nothing else — no email, no `org_owned`, no activation state. That is asserted against the shape, not eyeballed.

### The activation distinction, exercised live

The bug this PR nearly shipped with was testing "never activated" as `password IS NOT NULL`. A provisioned seat holds its **temporary** password from birth, so that filter would have admitted exactly the accounts meant to be excluded.

The live run walks a seat through all three states and checks the directory at each:

- **provisioned, not activated** → absent
- **activated** (chose its own password) → present
- **suspended** → absent again

Confirmed on real data rather than fixtures, and the API run separately shows a provisioned seat has `email_confirm_code: NULL` — the administrator made the account, so no confirmation mail exists. That is *why* activation has to be the forced-change flag clearing and cannot be inferred from the credential.

### The toggle in the console

Playwright, real browser:

```
PASS  the directory panel renders
PASS  the directory toggle is rendered
PASS  turning it on asks first, naming what apps will read
```

Screenshot `03-admin-directory-confirm` shows the dialog:

> Apps your members install will be able to read the usernames of everyone active in this team. They will not see emails, records, or suspended accounts. You can turn this off again at any time.

**One fix came out of looking at it.** The confirm button was styled `danger` (red), which reads as irreversible and put it in the same visual class as `Delete account` — while the dialog's own last sentence says it is reversible. Now `primary`.

### Verification

Typecheck clean · backend **7897 passed | 26 skipped** · GUI+SDK **275 passed** · all 13 commits typecheck individually.

Falsified earlier and still standing: the opt-in check (2 tests), the membership requirement (1), the exclusion filter (1).

Salazareo (approved):

PR #3806, opened 2026-09-05
fix(auth): honor min_pass_length 0 via nullish coalescing

Ayush7614:
What: all five min_pass_length reads used || 6, so a self-host setting of 0 (no minimum) silently fell back to 6.

Why: 0 is a meaningful config value. ?? 6 preserves it while still defaulting null/undefined, matching ShareController conventions.

How tested: npx tsc --noEmit -p tsconfig.build.json clean; npx vitest run --config src/backend/vitest.config.ts src/backend/controllers/auth/AuthController.test.ts — 318 passed. Verified || vs ?? behavior for 0 via node.

No related open/closed issues or PRs found (searched min_pass_length).

### DIFF
--- a/src/backend/controllers/auth/AuthController.ts
+++ b/src/backend/controllers/auth/AuthController.ts
@@ -401,7 +401,7 @@ export class AuthController extends PuterController {
                 legacyCode: 'password_required',
             });
         }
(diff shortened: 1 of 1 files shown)


Salazareo:
i think I won't change this for now, we probably don't want 0 length passwords

PR #3816, opened 2026-09-07
Fix/node signed batch upload

Xcoder2003X:
I Fixed `fs.upload()` for Node.js uploads containing nested folders and duplicate file names.

## Changes

- Enable signed batch-write uploads for `nodejs`.
- Prevent crashes when a response has no `content-type` header.
- Add a regression test for nested paths and duplicate file names.
- Update the upload documentation.

## Testing

```powershell
npm run test:puterjs:node -- -t "upload preserves nested paths and duplicate names in nodejs"

### DIFF
--- a/src/docs/src/FS/upload.md
+++ b/src/docs/src/FS/upload.md
@@ -71,7 +71,7 @@ When the signed batch-write endpoint is unavailable and the SDK falls back to th
 
 ## Uploading directories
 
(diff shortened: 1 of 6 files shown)


ProgrammerIn-wonderland:
Is this tested? To my recollection batch write uploads were disabled on nodejs and web/service workers due to them breaking uploads all together (and again in web workers because request latency is more important for them, especially puter workers). I understand enabling node, maybe not service/web workers though.

Salazareo:
> Is this tested? To my recollection batch write uploads were disabled on nodejs and web/service workers due to them breaking uploads all together (and again in web workers because request latency is more important for them, especially puter workers). I understand enabling node, maybe not service/web workers though.

was it broken on node? I thought that was working, was just prior broken on workers but might also be fine now

but yeah this for sure needs testing

Salazareo (approved):
did some testing and added a couple small fixes, should be good on node and workers now

PR #3828, opened 2026-09-08
feat(ai): refresh txt2vid catalogs, retire Sora, default to Veo 3.1 Lite

404oops:
## TL;DR

OpenAI shuts the Sora Videos API down on **2026-09-24**, and Sora was `puter.ai.txt2vid()`'s default model. This PR retires Sora entirely, makes **Veo 3.1 Lite on Gemini the default**, refreshes every video catalog against what the vendors serve today (Together: 9 retired models out, 18 new ones in with per-second billing; BytePlus: Seedance 2.5; Veo 3.1 Fast repriced), unifies the request options so `size`, `input_reference`, `last_frame` and `reference_images` mean the same thing on every provider, and rewrites the txt2vid docs to match. 342 backend tests pass, type check and docs build are clean.

**One known issue remains and should be decided before merge:** Veo's output URL is a Google file URL that requires the API key, so in a browser the returned element 403s and `puter_output_path` fails server-side. This predates the PR but the PR makes it the default path. Details and the proposed fix are in [Known issue](#known-issue-veo-output-url) below.

## Why now

- OpenAI announced the Sora 2 deprecation on 2026-03-24 with shutdown on 2026-09-24 and no replacement ([deprecations page](https://developers.openai.com/api/docs/deprecations)). `sora-2` was the default `txt2vid` model and `openai-video-generation` the default provider.
- The catalogs had drifted from the vendors: Together's live `/v1/models` lists 33 video models against our 23 (nine of ours no longer exist upstream), Google had repriced Veo 3.1 Fast and added reference images to Veo 3.1 Lite, and BytePlus made Seedance 2.5 generally available.
- The txt2vid documentation listed Veo models that the Gemini provider never served (Veo 2.0/3.0), had no BytePlus section, and named zero Together model ids.

### Default model and provider

- `VideoGenerationDriver`: `DEFAULT_PROVIDER` is `gemini-video-generation`; Gemini registers first. `GeminiVideoProvider.getDefaultModel()` and its unknown-model fallback both return `veo-3.1-lite-generate-preview`.
- Fixed a latent routing quirk: a request arriving under the generic `ai-video` driver name (what puter.js sends) used to fall through to whichever provider registered first rather than the default. It now lands on the default provider, with a regression test that pins `Context.driverName = 'ai-video'`.

### Sora removal

- Deleted `src/backend/drivers/ai-video/providers/openai/` (provider, catalog, tests).
- Removed `openai-video-generation` from the driver aliases and provider registration, from `config.template.jsonc`, and from `tools/lib/configMigration.mjs` (`PROVIDER_IDS` and the `openai` fan-out). Nothing reads the key any more.
- Removed Together's `openai/sora-2` and `openai/sora-2-pro` catalog rows.
- Purged Sora from the docs, both playground examples that used it, the SDK type comments and two code comments. A tree-wide search finds no remaining references.

### Catalog refresh (from live vendor listings and vendor docs)

| Provider | Source queried | Change |
|---|---|---|
| Google | `GET /v1beta/models`, [Veo docs](https://ai.google.dev/gemini-api/docs/veo), [pricing](https://ai.google.dev/gemini-api/docs/pricing) | Veo 3.1 Fast rates corrected from 15¢/35¢ to **10¢ (720p) / 12¢ (1080p) / 30¢ (4K) per second**; Veo 3.1 Lite now accepts up to 3 reference images. |
| Together AI | `GET /v1/models` (type=video), serverless catalog, Seedance 2.5 and Wan 2.7 quickstarts | Removed `google/veo-3.0` (all four variants), `kwaivgI/kling-2.0-master`, `kwaivgI/kling-1.6-pro`, `vidu/vidu-2.0`, `Wan-AI/Wan2.2-I2V-A14B`, `Wan-AI/Wan2.2-T2V-A14B`. Added `ByteDance/Seedance-2.0`, `ByteDance/Seedance-2.5`, `google/veo-3.1`, `google/veo-3.1-lite`, `Wan-AI/wan2.7-{t2v,i2v,r2v}`, `alibaba/happyhorse-1.0-{t2v,i2v,r2v}`, `alibaba/happyhorse-1.1-{t2v,i2v,r2v}`, `vidu/vidu-q3`, `vidu/vidu-q3-turbo`, `pixverse/pixverse-v5.6`, `pixverse/pixverse-v6`, `black-forest-labs/FLUX-3`. Seedance 1.0 Lite price 14 → 14.3¢. |
| BytePlus ModelArk | Create-task API reference and pricing page (rebuilt from the page source; the docs site is client-rendered) | Added `dreamina-seedance-2-5-260628` (alias `seedance-2-5`): 4–30 s, 480p/720p/1080p, 10.70 USD/M tokens at 480p/720p and 11.7 at 1080p. All existing BytePlus rates re-confirmed. |
| OpenAI | `GET /v1/models` | Only `sora-2`/`sora-2-pro` are video; both removed per above. |

### Together AI billing: per second, from the job's reported cost

The new Together models are priced per second of output, which the previous flat `per-video` billing could not express. `TogetherVideoProvider` now:

- keeps `per-video` for the older flat-priced models (all-or-nothing gate, unchanged);
- for models with `per-second` (plus optional `per-second-<tier>`) rates, estimates from the catalog rate, clamps the duration to the remaining balance through the shared `capSecondsToRemainingCredits` (same behaviour as Veo and Seedance), and sends the clamped `seconds`;
- meters the amount Together reports in `outputs.cost` on the finished job (falling back to the estimate when absent). Catalog per-second rates are the highest published rate for the model so the pre-flight estimate never undershoots;
- sends `resolution` (and `ratio` derived from width/height) for tier-sized models (Seedance 2.5, Wan 2.7) instead of width/height, and forwards `generate_audio`.

### Unified request options

- **`size`** as `WIDTHxHEIGHT` now works on every model. The driver maps it onto tier-based catalogs (Seedance, Wan 2.7) by the tier of the shorter side and fills `width`/`height` from it, so tier providers derive the aspect ratio and Together's pixel-sized models get explicit dimensions. Tier names (`720p`) still work.
- **`input_reference` / `last_frame`** now work on Together AI (mapped onto its first/last keyframes; an explicit `frame_images` still wins). They already worked on Veo and Seedance.
- **Image inputs accept a URL, a data URI or raw base64 on all providers.** Veo only takes inline bytes, so it now fetches URLs server-side through the SSRF-guarded `secureFetch`. To avoid a sideways import from the image driver, the string-level helpers (`isHttpUrl`, `toBase64DataUri`, `fetchImageAsBase64`, …) moved to `src/backend/drivers/util/imageInput.ts`; `src/backend/drivers/ai-image/inputImage.ts` re-exports them, so image providers and their tests are unchanged.
- **Not unified, on purpose:** `seed` and `generate_audio` on Veo (the Gemini API does not document them; Veo always includes audio) and `negative_prompt` on BytePlus (no such input). The docs say so.

### puter.js

- `Txt2VidOptions` gains `provider` and `generate_audio`; option comments now describe the unified semantics and the new default. No runtime change; `npm run check:puterjs:types` is clean.

### Docs

- `src/docs/src/AI/txt2vid.md` rewritten (171 → 443 lines): per-provider model tables with durations, sizes, keyframes and pricing style; a "Choosing a model" section on aliases and normalization; an "Image inputs" table; "Cost and clip length" explaining the credit clamp; Node.js/Workers return shape; expanded error table; five new examples.
- Four new playground examples (`ai-txt2vid-veo`, `-image-to-video`, `-save`, `-errors`) registered in `examples.js`; the AI overview lists them.

## Breaking and behaviour changes

- `sora-2`, `sora-2-pro`, `openai/sora-2*` and `togetherai:openai/sora-2*` now return 400 `Model not found`. The provider id `openai-video-generation` no longer exists.
- Calls without a `model` now use Veo 3.1 Lite (4 s, 1280x720, with audio) instead of Sora 2 (4 s, 720x1280). Default clip cost drops from about $0.40 to about $0.20.
- The nine retired Together ids above return `Model not found`.
- Veo 3.1 Fast is cheaper than before; the new Together models are metered per second and may be shortened to fit the balance, like Veo and Seedance already were.
- Together requests that pass `size` now carry explicit `width`/`height` (pixel models) or `resolution`/`ratio` (tier models). Requests with no `size` are unchanged.
- No rate limits, concurrency caps or quotas change, so `rate-limits-and-quotas.md` is untouched.

## Known issue: Veo output URL

A live 4-second Veo 3.1 Lite job through the local instance (~$0.20, 43 s) returned `https://generativelanguage.googleapis.com/v1beta/files/<id>:download?alt=media`. Google requires the `x-goog-api-key` header to download it; an anonymous fetch gets 403 `PERMISSION_DENIED`. No key is leaked in the URL. Consequences today: in a browser the returned `<video>` cannot play, and `puter_output_path` fails because the server-side fetch carries no key. Google keeps the file for two days.

This is pre-existing Veo behaviour, but this PR makes Veo the default path, so it should be fixed before or immediately after merge. Proposed follow-up: type the provider result (`IVideoResult` with `url`/bytes, `mime_type`, `model`, `provider`, `duration_seconds`, size and `cost`), have the Veo provider fetch the file server-side with the key, and deliver either a Puter-hosted URL or inline bytes. `toVideoElement` in puter.js already accepts an object result with `url` and `mime_type`, so the SDK needs no change for that.

## Verification

- `vitest` on `src/backend/drivers/ai-video` and `src/backend/drivers/ai-image`: 342 passed, 4 skipped (live integration tests). New tests cover the driver default and `ai-video` alias fallback, size mapping for tier and pixel models, Together per-second estimation, credit clamp, reported-cost billing, tier/ratio payloads and keyframe mapping, Veo URL inputs and failed fetches, Veo 3.1 Fast 1080p metering, Veo 3.1 Lite reference images, Seedance 2.5 aliases, reference caps and 1080p billing.
- `npm run typecheck` and `npm run check:puterjs:types`: clean. ESLint: 0 errors on changed files.
- Docs build: clean; every example fence has a playground page.
- One live Veo 3.1 Lite generation (above) exercised the default path end to end.

## What was not verified

- Together model removals rest on the live listing plus Together's docs, not on paid generations. Per-second rates for HappyHorse, Vidu Q3, PixVerse v5.6/v6 and FLUX 3 come from the listing's pricing text; durations and sizes for those are unpublished, so the docs say "provider default".
- Together lists Veo 3.1 Lite at "$0.80/s with audio", almost certainly a typo for $0.08; the catalog uses 8¢/s. Actual billing uses Together's reported cost either way.
- BytePlus changes are verified against BytePlus's own API reference and pricing page only; the local instance has no BytePlus key. The 30-image cap for Seedance 2.5 is Together's published limit for the same model; BytePlus states a 50-item total across images, video and audio.

## Reviewer callouts

- **Billing:** Together per-second metering from `outputs.cost`, Veo 3.1 Fast rate change, Seedance 2.5 rates. See `TogetherVideoProvider.generate` and the three `models.ts` files.
- **Outbound fetches:** Veo image inputs are fetched through `secureFetch` (SSRF-guarded), the same path the image drivers use. No new endpoints, permissions or config keys; two config keys removed.


Salazareo (approved):
one nit: idk if we need/want that error example in the playground, not sure what the use case for it is

PR #3846, opened 2026-09-10
feat: the team seat experience — no email required, forced password change, team label, and plan-based limits (PUT-1792)

jfcastro92:
Everything a team-provisioned account meets on its way in, and everything its
owner meets running the team. Found by testing Teams on a region and then
driving the whole flow locally against Stripe; seventeen commits, each
independently reviewable.

| | |
| -- | -- |
| Seat needs no email, and is never asked to confirm one | `b19b312` |
| Seat is prompted to choose its own password on first sign-in | `a983969` `fa36f2e` |
| Seat is emailed its credential, when an address is given | `fb74789` |
| Seat is told which team its account belongs to | `106978e` |
| Team size and allowance depend on whether the owner pays | `e886e86` `63f857c` |
| A plan is bought per account, not one tier for the team | `efe6ef0` `2083ef4` `f2c9737` `187e4c8` `b8b0650` `c2a975b` `06bdc26` `1efe086` |
| Per-account actions are icons | `bafe8af` |
| A seat's account surface is the one that is actually its own | `480d6bc` |
| Running out of seats says what raises the limit | `63f857c` |
| A team tier is named, not printed as its policy id | `de74b0c` |
| A seat is not offered a plan link it cannot use | `8d90418` |

Closes PUT-1792. Most of this is beyond that ticket — see **Scope** at the end.

---

## 1. Email was mandatory, and confirmation was demanded

`createMember` read the address with `#requireString`, and `provisionAccount` validated and de-duplicated it. An admin creating ten seats had to invent ten addresses and then live with ten uniqueness constraints — for accounts that **sign in by username**.

It also set `requires_email_confirmation: true`, reasoning that an admin-supplied address is unverified. True, but `requireVerifiedAccount` turns away on exactly

```ts
if (user?.requires_email_confirmation && !user?.email_confirmed) { ... }
```

so a fresh seat was asked to confirm an address it may not hold, and could not use the product until it did. The team creating the account is the trust anchor, not the mailbox. Now `false` either way.

`idx_user_owned_email` is partial and skips password-null rows, so omitting the address sidesteps the index rather than creating a collision surface — two emailless seats do not conflict, and there is a test for it.

## 2. Nothing asked the seat to change its password

The backend already refused every route with `password_change_required`, and `/user-protected/change-password` was already exempt — *"the one route an account owing a password change may reach"*. **The client half was missing entirely**: zero references to that code in the GUI, so a seat signed in and then failed at everything with no prompt.

Two gaps. The flag never reached the client — neither the login response nor `/whoami` carried `requires_password_change`. And no window could serve as a gate: `UIWindowChangePassword` is a settings dialog, closable and never resolving on success.

`UIWindowPasswordChangeRequired` mirrors the existing `*Required` windows, resolves only once the change lands, and `initgui` loops on it. It runs **last**, matching `assertVerifiedAccount`'s own order. It refuses a new password equal to the current one — otherwise the account stays on the credential its administrator still holds, which is the entire thing the gate exists to end.

`initgui` runs its verification gates in **two** places — the token-in-URL branch and the session-restore/login branch — and the first commit wired only one, so signing in with a password skipped the gate entirely. `fa36f2e` adds it to both and pins the invariant with a test that asserts each gate appears twice.

## 3. The credential had no delivery

`team_account_created` said the team *"will send you a temporary password separately"* and carried none. It now carries the username and temporary password, so an admin who supplies an address hands nothing over by side channel. Both credential-issuing paths pass the one they just minted — provisioning **and** re-issue; there is a test asserting the *old* password is absent from the re-issue mail.

With no address nothing is sent, which was already true (`#notifyUser` returns early) and is now covered.

## 4. A seat could not tell it was one

`whoami` now carries `team: { uid, name }`, gated on **user actors** — a seat's employer is no more an app's business than its phone number, which the same handler already withholds — and on `teams_enabled`, so a deployment without teams is byte-identical.

It rides whoami rather than a route of its own because the sidebar needs it at first paint; a `/teams/whoami` would add a request to every page load for every user, and almost none are seats. The lookup costs nothing: `getOrgSeat` is already cached, negatives included, and `team_name` comes off a join the query already made.

In the sidebar it sits under the Puter wordmark as a muted second line, hidden when collapsed. Owners see nothing — they already know, and one may own several teams, so there is no single name to show.

## 5. Team size followed nothing at all

**4 seats free, 40 paid**, decided by `subscriptionSatisfies(id, true)` — which is `!FREE_SUBSCRIPTION_IDS.has(id)`, so a plan an extension registers counts as paid without core knowing its name. An unreadable plan takes the **smaller** cap: over-provisioning a free team is the worse failure.

Two defects kept that from being true of any real deployment, both found by testing it rather than reading it:

- **`config.default.json` shipped `max_seats_per_team: 50`**, and the flat override wins over the plan — so the whole free/paid branch was dead code unless an operator deleted a key nothing told them about. A free owner got 50 seats. The defaults now carry `max_seats_per_team_free` / `_paid` and leave the flat override unset; a deployment that wants one number for everyone still sets it and still wins. A test reads `config.default.json` and fails if the flat key comes back.
- **`#ownerPays` asked metering about an `{id, uuid}` stub.** A resolver keyed on any other field missed, so the answer came down to whether something else had cached that user's plan first — the same team, at the same seat count, was refused or allowed depending on nothing. It passes the whole user row now, and a test asserts the actor carries the username.

**A seat of a team that pays for nothing gets half the free plan** (`org_seat_free`). Without it a team is a way to mint free tiers — provision four seats and each arrives with a full free allowance nobody paid for. The figures are derived from `REGISTERED_USER_FREE` rather than restated, so the two cannot drift, and the id joins `FREE_SUBSCRIPTION_IDS` because nobody paid for it either and it must not satisfy a plan gate. It is a **default** resolver, so a paid team tier — which only prod knows about, via `registerSubscriptionResolver` — still outranks it.

Running out of seats used to report the number and stop. It now says what raises the limit and offers the plan picker, and the button appears only where a billing extension put one on the page.

## 6. A plan belongs to an account, not to a team

The first pass sold one tier for the whole team. That is the wrong unit: a team buys Puter for particular people, and the people differ. Each seat now carries its own tier — one subscription, one Stripe item per tier, quantity = seats on that tier — and the accounts table has a **Plan** column with a per-row action.

The plan card came out again (`c2a975b`) once every row carried its own, and the picker moved to the billing extension (`06bdc26`) so it reuses the personal plan modal's own frame rather than a lookalike. OSS renders from a prod-served catalogue — the precedent `TabHome.js` already set — and carries no pricing.

`1efe086` is the one that made it work. The action did nothing, through three layers: `listMembers` never returned a member's `uuid`, the SDK's `toMember` dropped it, and the Plan column read `seatTiers[undefined]` so every seat rendered as *Free*. The uuid is owner-only — it is what billing keys a seat's plan on, and one member has no business identifying another.

## 7. A seat's account surface was not its own

A provisioned account could rename itself, delete itself, and see a Billing tab for a subscription it does not hold. Each is now refused server-side and dropped from the UI, keyed on one predicate (`isOrgSeat`): whoami reports a team only for an org-owned seat, and the owner joined their own team.

The Teams tab showed a seat nothing but its own audit rows, so a member could not see who else was on the team they were told they shared with. It lists them now; `listMembers` was already membership-gated and already withholds from a member what is not theirs.

The dashboard share modal had no way to reach a team, though the desktop dialog has had one since team sharing shipped and the SDK has always taken `{ team }`. Same control, same copy, same `shareTeams` helper. The access list needed a team bucket to go with it: a team share names no holder, so the aggregate dropped it and a team you had just shared with vanished from the list.

Smaller things from the same pass: the record table pages at ten rows instead of running off the bottom, and the team card no longer labels the absence of a handle that nothing here can set.

## 8. A team tier printed as its policy id

`i18n()` echoes a key it has no translation for. Personal tiers have keys;
`team-basic` does not — so a seat's dashboard read **"team-basic"** where a
personal plan reads "Basic". The offering already travels with the subscription
(prod's `subscriptions/current` now carries a seat's team tier), so the display
name was already on the wire. `i18n` stays the fallback for the tiers that do
have keys.

Found by driving a real team purchase against a local instance, not by reading.

<details>
<summary><b>A seat's dashboard, after the fix</b> (<code>~/Desktop/team-billing-fixes-20260911/07-seat-sees-its-plan-not-free.png</code>)</summary>

07-seat-sees-its-plan-not-free

</details>

## 9. A seat was offered a plan link it cannot use

The Usage card's Upgrade link relabels to **"Manage →"** for anyone on a plan —
including a seat, whose plan belongs to its team. Following it only reaches a
dialog saying so. Hidden for a seat on the same `isOrgSeat` predicate the rest of
the seat restrictions use; unchanged for everyone else, verified both ways
against a live instance (owner still sees "Upgrade →", seat sees nothing).

---

## Test bugs this surfaced

- The team suite was reading `max_seats_per_team` out of the **developer's own `config.json`**, so the cap under test was whatever that file said. It would have passed locally and failed in CI, which has no such file. The suite pins the value and the cap tests set their own.
- A new HTTP test shared a rate-limit budget with two unrelated ones and pushed them into `429`. The fixture is built through the service instead of over the wire.

## Testing

```
backend (team, auth, share, whoami)   778 passed (17 files)
gui + sdk                             564 passed (45 files)
npm run typecheck                     no new errors
```

Driven end to end against a local instance with Stripe in test mode — real checkout sessions, real webhooks, real seats:

```
free cap holds          3 seats → 4th allowed → 5th refused, 409 seat_limit_reached, limit 4
paid cap holds          max_seats_per_team_paid pinned → refused at that number, not the default
seat restrictions       no Billing tab, no Delete Account, no Change Username; note names the team
member roster           8 colleagues listed, "(you)" on self
team share              share → mode change → revoke, all three on the team's own row
per-account plan        picker opens on the seat's uuid and renders the catalogue's own copy
seat limit dialog       "Your plan covers 5 accounts…" + See plans
```

Falsified, each breaking only its own test:

```
requires_email_confirmation: true  → × never demands confirmation, with or without an address
same-password guard removed        → × refuses to reuse the password the admin handed over
resolve on a rejected response     → × stays open on a rejected change, so the gate cannot be escaped
gate on one initgui branch only    → × wires the gate into both chains
credential dropped from re-issue   → × emails the fresh credential on re-issue, not the old one
isUser gate removed                → × withholds it from an app actor
free and paid caps equal           → × lets a paid owner past four
flat cap back in config.default    → × is not overridden by the shipped defaults
id/uuid stub actor restored        → × asks about the owner with a whole user
seat resolver returning null       → × resolves half the free plan
allowance not halved               → × resolves half the free plan
uuid withheld from the owner       → × keeps the seat uuid, which the plan action is keyed on
team bucket removed from aggregate → × gives a team a row of its own
```

## Scope

Only commit 1 is PUT-1792. The rest came out of the same rounds of testing and are small enough to have travelled with it, but say the word and I will split any of them out.

Two calls I made rather than guessed, both cheap to reverse:

- **"Free org users" means seats, not owners.** The owner keeps their own plan; only `org_owned = 1` accounts get the halved policy. That also preserves the disjointness the two subscription resolvers rely on.
- **50% applies to usage only — not storage.** I originally wrote that it covered both; local testing disproved it. `ORG_SEAT_FREE` halves `monthlyStorageAllowance`, but **nothing reads that field** — real storage is `user.free_storage`, and a provisioned seat takes the ordinary default. Measured: a tierless seat gets **500 credits (half)** and **100 MB storage (full)**. Halving storage too means setting `free_storage` at provisioning, which also raises what to do about seats already created. Say the word and I'll do it here or file it.

Two things worth a reviewer's eye:

- `prodMeteringAndBilling` also registers a *default* resolver, for corporate email domains. Two defaults means `firstResolver` picks by registration order, so a seat on a trusted domain could land on `corpo` rather than `org_seat_free`. Same class of problem the team resolver already avoids by keying on `getOrgSeat`, and worth a follow-up if that overlap is real in prod.
- Nothing in the dashboard sets a team **handle**, so every team made there has none. I hid the empty label rather than inventing a way to set one; if handles should be settable, that is a separate change.

## Deleted-team edge cases (`230241d6d`)

Validating "what happens to users when a team is deleted" surfaced two gaps in `deleteMember`, both fixed here:

- **Seats of a deleted team were stranded.** Deletion suspends every provisioned seat, but all member routes resolved live teams only — the suspended accounts could never be deleted afterwards. `deleteMember` (service + route guard) now resolves the team soft-deleted or not, the same way the audit reader does. The GUI's promise is unchanged: deletion still suspends, and *recovery* still takes a support request — `enableMember` still refuses a deleted team.
- **A platform suspension could be deleted out from under Puter.** The disabled-first check accepted any suspension, so an owner could cascade-delete a seat Puter had frozen. It now requires the team's own suspension (`disabled_by_team`), mirroring `enableMember` — anything else answers 409 `conflict`.

Covered in `TeamEvents.test.ts`: deleting a suspended seat after team deletion succeeds and still emits `team.account.deleted` (the billing stop fires — `getOrgSeat` includes deleted teams), and a Puter-suspended seat is refused. The billing counterpart (ending a deleted team's paused seat subscriptions) is in heyputer#1212.

## Staged rollout by email domain (`ad3d15e7e`, `f428797a7`)

New optional `teams_allowed_email_domains` config: when set, only accounts on those domains may enter the teams surface. Enforced on the two entry routes — `POST /teams` (getting in) and `GET /teams` (what shows the tab) — answering the same 404 `not_found` a teams-off deployment gives, so the GUI needed no change and the rollout is invisible to everyone else. Members of an existing team always pass regardless of their domain (an allowed owner brought them in), so provisioned seats with outside emails keep working. Domain match is case-insensitive and exact (no subdomains); unset keeps today's behavior.

Covered in `TeamDomainGate.test.ts` (HTTP-level): allowed domain creates/lists, case-insensitive match, other domains get the teams-off 404 with `code: not_found`, and an off-domain provisioned seat still lists its team via membership.

The GUI flag follows the same gate (`f428797a7`): `gui_params.teams_ui` remains the deployment-wide switch, **or** — with it off — the shell renders `teams_ui: true` per user when they pass the allowlist (membership included, so provisioned seats see their roster). Anonymous renders hide it, and the API decides real access either way. So a staged deployment sets only `teams_allowed_email_domains` and drops `gui_params.teams_ui`; staff get the full UI, everyone else sees nothing.


Salazareo (changes requested):
mightve missed it, but think the free org seat isnt being included in the gate for subscription requirement

jfcastro92:
Good catch — you were right, not only one gate.

`org_seat_free` was in `FREE_SUBSCRIPTION_IDS`, so `requireSubscription` / `assertActorHasSubscription` did refuse it correctly. But two other surfaces decide on plan and neither consults that set:

**1. Every `bySubscription` cap.** Drivers enumerate `user_free` and `temp_free` by name, and `resolveSubscriptionLimit` falls through to the top-level `limit` — the *paid* cap — for anything unlisted. So a seat outranked an ordinary free account:

| | free user | seat (before) |
| -- | -- | -- |
| events `list` | 120/min | **240/min** |
| durable subscriptions | 100 | **500** |
| kv `list` concurrency | 3 | **5** |

Same shape in the kv, notification, subdomain and worker drivers.

**2. `subscriberOnly` models.** `ChatCompletionDriver` compared `subscription.id` against the two named ids, so a seat paying nothing could reach a paid-only model.

Fixed at the two resolvers rather than by adding `org_seat_free` to ~10 driver maps — an unlisted plan that is in the free set now takes the `user_free` entry. That covers every driver at once and any free plan added later, which is the same reason the set exists.

Guarded against over-correcting: a *paid* plan that names no cap of its own still takes the base, and an unresolved plan still takes the base. Tests for both, plus a seat held to the free cap and a seat refused a subscriber-only model.

`15527eac4`. Backend suites: 2848 passed.

jfcastro92:
Team Creation, account creation with email notification and password reset lifecycle (zenpacket):
Screenshot 2026-09-14 at 10 22 21 AM

jfcastro92:
## End-to-end purchase run, against a local instance

Teams on, Stripe in test mode, `stripe listen` forwarding to
`/marketplace/webhook`. Run with this branch checked out **and** the private
halves that go with it (heyputer#1212 for the endpoints, the `gui` branch for
the picker) — OSS alone renders the Teams tab but has no catalogue behind it,
which is exactly why the plan action stays hidden on a deploy that lacks them.

Team **SimCo**, owner `admin`, seats `sim_alice`, `sim_bob`, `sim_carol`.
Screenshots are from `~/Desktop/team-purchase-sim-20260914/`.

---

### 1. Before — every seat on Free

<details>
<summary><code>s1-team-before.png</code></summary>

s1-team-before

</details>

### 2. The tier picker

Opened from the per-account action. It reuses the personal plan modal's own
frame, and the copy comes from the offerings catalogue rather than being written
in the GUI.

<details>
<summary><code>s2-picker.png</code></summary>

s2-picker

</details>

### 3. Checkout — a real Stripe session

Card `4242`, full billing address (`automatic_tax` needs one). The mount is
capped at `80vh` and scrolls, so Subscribe stays on screen.

<details>
<summary><code>s3-card.png</code></summary>

s3-card

</details>

### 4. First seat lands on Team Basic

<details>
<summary><code>s4-alice-paid.png</code></summary>

s4-alice-paid

</details>

### 5. Second seat, different tier — **no second checkout**

`sim_bob` assigned Team Pro. **0 Stripe iframes**: the team already pays, so
this is an item change on the existing subscription, not a new sale.

<details>
<summary><code>s5-both-seats.png</code></summary>

s5-both-seats

</details>

---

### What Stripe actually holds

```
subscriptions for this team : 1                   <- not one per purchase
customer                    : cus_VF6K2gLUGRst5y  <- the OWNER, reused not recreated
items                       : team-basic qty 1, team-pro qty 1
seatmap_0                   : alice:team-basic,bob:team-pro
```

One subscription, one item per tier, quantity = seats on that tier. The owner is
the customer — a team plan is not a personal one, and `subscriptions/current`
for `admin` still reports no personal subscription.

`seatmap_0` is the seat map mirrored onto the subscription: Stripe's items carry
quantities and never *which* seat, so without it a lost kv entry could not be
rebuilt.

### What Puter holds

```json
{ "seatTiers":      { "alice": "team-basic", "bob": "team-pro" },
  "tierQuantities": { "team-basic": 1, "team-pro": 1 } }
```

### What each member sees

| account | plan | storage | credits |
| -- | -- | -- | -- |
| `sim_alice` | Team Basic | 100.10 GB | 18,000 |
| `sim_bob` | Team Pro | 1.95 TB | 190,000 |
| `sim_carol` *(no tier)* | Free | 100.00 MB | 500 |

All three report `subscribed` naming SimCo via `via_team`, and none of them sees
a Billing tab or a plan link — the team owns the plan, so those would only lead
to a refusal.

<details>
<summary><code>s6-sim_alice.png</code> — Team Basic</summary>

s6-sim_alice

</details>

<details>
<summary><code>s6-sim_bob.png</code> — Team Pro</summary>

s6-sim_bob

</details>

<details>
<summary><code>s6-sim_carol.png</code> — tierless seat</summary>

s6-sim_carol

</details>

---

### One discrepancy this run found

`sim_carol` is a seat with no tier, so it should be on `org_seat_free` — **half**
the free plan. It gets half the usage (500 credits against a free account's
1,000) but the **full** 100 MB storage, not 50 MB.

`ORG_SEAT_FREE` does halve `monthlyStorageAllowance`, but nothing reads that
field — it is declarative. Real storage is `user.free_storage`, and a
provisioned seat takes the ordinary default.

I had written in the description that the 50% covered usage *and* storage; that
was wrong and is now corrected there. Halving storage too means setting
`free_storage` at provisioning, which also raises what to do with seats already
created — happy to do it here or file it, whichever you prefer.

jfcastro92:
## Test report addendum — latest local run

The full e2e purchase report is in the comment above; this covers what changed
on this branch since then, exercised in the same live run as heyputer#1212's
report.

| check | result |
| -- | -- |
| New i18n keys render through the extensions | ✅ the cancel dialog shows the `teams_cancel_keeps_until` copy verbatim |
| Seat's dashboard after the billing reshape | ✅ plan name (`name_en`, not the policy id), plan-tier storage and credits |
| Seat restrictions | ✅ no Billing tab, no plan link, username/email/delete guarded (403s), recovery refused |
| Owner unaffected | ✅ `isOrgSeat` false → Upgrade link visible, all account actions present |

Backend suites on this branch: 778 team/auth/share/whoami tests passing;
GUI + SDK 564; typecheck clean.

Salazareo (approved):

ProgrammerIn-wonderland (approved):
LGTM

ProgrammerIn-wonderland (commented):
LGTM

src/backend/drivers/ai-chat/ChatCompletionDriver.ts
```diff
@@ -920,10 +917,8 @@ export class ChatCompletionDriver extends PuterDriver {
 
         if (model.subscriberOnly) {
             const subscription = await metering.getActorSubscription(actor);
-            const isDefaultPolicy =
-                subscription.id === DEFAULT_FREE_SUBSCRIPTION ||
-                subscription.id === DEFAULT_TEMP_SUBSCRIPTION;
-            if (isDefaultPolicy) {
+            // Every free plan, not two named ones.
+            if (FREE_SUBSCRIPTION_IDS.has(subscription.id)) {
```
Salazareo, line 921: we should use new method that checks for this. but ill fix in different PR

jfcastro92: Tracked in [PUT-1830](https://linear.app/puter/issue/PUT-1830/free-plan-checks-reach-into-free-subscription-ids-directly-instead-of) — `subscriptionSatisfies(id, true)` already owns this rule, so the driver site is a drop-in; the two limit-fallback sites are listed there too.

jfcastro92: https://linear.app/puter/issue/PUT-1830/free-plan-checks-reach-into-free-subscription-ids-directly-instead-of

Issue #3838, opened 2026-09-09
choosing which infron pricing to show


reynaldichernando:
hi team, i learned today that infron has multiple pricing for the same model

<img width="986" height="795" alt="Image" src="https://github.com/user-attachments/assets/805a68d1-b548-4168-acd7-fc5efef11b47" />

in the case of gpt 6 astra, i see that our API lists the flex pricing ($5 in, $25 out) 
<img width="423" height="369" alt="Image" src="https://github.com/user-attachments/assets/e844c8a4-d828-4f57-9d5c-d35523fbd13d" />

the flex pricing for this model seems to be for batch jobs, and it also shows zero uptime

just want to make sure if this is something expected/intentional

404oops:
any way we can see which plan was used?

PR #3852, opened 2026-09-11
fix: quote and pin Infron service tiers

404oops:
Infron sells the same model at several service tiers, and our catalog read the wrong price for it.

`min_prompt_price` / `min_completion_price` are the floor across *every* tier, so any model that also sells a flex tier was advertised at a batch-job price a default request never gets — **25 of the 286 chat models we list**. `openai/gpt-6-astra` showed $5/$25 per million against the $7.5/$37.5 it actually bills; `gpt-5.5` showed $2.5/$15 against $3.75/$22.5.

## What changed

- **Price each tier from its own `providers[]` row** instead of the catalog-wide floor, and **pin that tier on the request** (`provider: { service_tier }`), so the price quoted is the price charged rather than whatever Infron's load balancer picks.
- **Every non-default tier gets its own model id**, so a caller can opt into one by name — no new API option needed, since `model` already flows through puter.js untouched:

  ```js
  puter.ai.chat(prompt, { model: 'infron:openai/gpt-6-astra:flex' })
  ```

  The plain id stays the default (standard) tier, with `…:standard` accepted as an alias rather than a duplicate listing. The catalog goes from 286 to 343 Infron entries: +36 `:flex`, +21 `:priority`.
- A tier is only pinned when the model actually sells it; otherwise routing is left to Infron.

## Why it matters beyond display

Billing itself was never wrong — `usage_calculator` bills Infron's reported `cost`. The stale prices leaked into two places that *do* move money:

1. **The credit gate.** `#applyCreditGate` derives `maxAllowedOutputTokens` from `model.costs`. Quoting two-thirds of the real rate let a request run roughly 50% past the balance it was gated against, leaving the user negative.
2. **The cost-less fallback.** When a response carries no `cost`, `usage_calculator` prices per token from the catalog — there, understating is a straight undercharge.

## Implementation notes

- **Suffix parsing matches the catalog exactly before treating a trailing segment as a tier.** Three catalog ids carry a colon of their own (`deepseek/deepseek-v4-flash:free` and two more); a naive `split(':')` would turn those into a nonexistent model at tier `free`. Covered by a test.
- **Tier variants take the context window of their own offering**, which differs from the model-level figure for 11 of them — most sharply `google/gemma-4-31b-it`, where priority is 131070 vs 262144. Base entries keep the model-level value, so existing ids are byte-identical to before.

## Docs

Second commit, `docs: document Infron service tiers`.

- **`AI/chat.md` gains a Service tiers section** (plus a pointer from the `model` parameter): what the `:flex` / `:priority` suffixes mean, that a plain id is the default tier, that the tier named in the id is the tier billed, and that a tier can carry a smaller context window than its siblings. Written vendor-neutrally with Infron as the worked example.
- **`AI/listModels.md`** notes that a model sold at several tiers appears once per tier.
- **Drive-by correction, pre-existing:** `listModels.md` documented a `cost` object with `currency` / `input` / `output` keys. No such field is returned. Pricing comes back as `costs` alongside `costs_currency`, keyed per vendor and named by each model's own `input_cost_key` / `output_cost_key` (`prompt`/`completion` for Infron, `input_tokens`/`output_tokens` for Claude). Verified against production — of 941 models on `api.puter.com`, not one carries a `cost` key, so anyone following that page was reading `undefined`. The prose is corrected and the example replaced with a real entry.

No tests accompany the docs commit, per AGENTS.md: tests must not assert on the wording or numbers of files under `src/docs/`.

## Testing

- 22 tests in the Infron suite (9 new); 1013 pass across `src/backend/drivers/ai-chat`. Each regression test was confirmed failing before its fix.
- Verified against the live catalog through a local stack: all 343 entries match the cheapest offering of their tier, zero mismatches.
- Three real completions on the same model, one per tier, billed $6.0e-06 / $4.0e-06 / $1.7e-05 — matching direct-to-Infron calls and the advertised per-tier prices.

Refs PUT-1789, #3838

### DIFF
--- a/src/backend/drivers/ai-chat/providers/infron/InfronProvider.test.ts
+++ b/src/backend/drivers/ai-chat/providers/infron/InfronProvider.test.ts
@@ -121,6 +121,82 @@ const SAMPLE_API_MODELS = [
         min_completion_price: 10,
     },
     {
+        // Sold at several service tiers; the flex tier undercuts standard.
+        id: 'openai/gpt-6-astra',
+        display_name: 'OpenAI: GPT-6 Astra',
+        category_type: 'LLM',
+        supported_endpoint_types: ['openai'],
+        context_length: 1050000,
+        max_output_tokens: 128000,
+        min_prompt_price: 5,
+        min_completion_price: 25,
+        providers: [
+            {
+                provider_slug: 'openai/flex',
+                service_tier: 'flex',
+                prompt_price: 5,
+                completion_price: 25,
+                context_length: 400000,
+            },
+            {
+                provider_slug: 'azure',
+                service_tier: 'standard',
+                prompt_price: 7.5,
+                completion_price: 37.5,
+            },
+            {
+                provider_slug: 'openai',
+                service_tier: 'standard',
+                prompt_price: 10,
+                completion_price: 50,
+            },
+            {
+                provider_slug: 'openai/priority',
+                service_tier: 'priority',
+                prompt_price: 20,
+                completion_price: 100,
+            },
+        ],
+    },
+    {
+        // Flex is the only tier on offer — nothing to pin.
+        id: 'example/flex-only-model',
+        display_name: 'Flex Only',
+        category_type: 'LLM',
+        supported_endpoint_types: ['openai'],
+        context_length: 128000,
+        max_output_tokens: 4096,
+        min_prompt_price: 1,
+        min_completion_price: 2,
+        providers: [
+            {
+                provider_slug: 'example/flex',
+                service_tier: 'flex',
+                prompt_price: 1,
+                completion_price: 2,
+            },
+        ],
+    },
+    {
+        // Catalog ids can contain a colon of their own.
+        id: 'example/colon-model:free',
+        display_name: 'Colon Model (free)',
+        category_type: 'LLM',
+        supported_endpoint_types: ['openai'],
+        context_length: 64000,
+        max_output_tokens: 4096,
+        min_prompt_price: 0,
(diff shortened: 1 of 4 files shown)


reynaldichernando (commented):
**edit: add consideration for docs**

i think this is a good practical solution, just having the service tier as suffix
even openrouter does this with their `:nitro` models https://openrouter.ai/docs/guides/routing/model-variants/nitro
(would be equivalent to `:priority`)

although since i think there aren't that many models that use this 3 service tier variant, we should probably leave this out from the documentation
i'm afraid it will be too specific to only a few models from infron

these ones only have standard tier:
https://infron.ai/models/anthropic/claude-fable-5.1
https://infron.ai/models/anthropic/claude-fable-5
https://infron.ai/models/anthropic/claude-sonnet-5
https://infron.ai/models/z-ai/glm-5.2-fast
(and more)

---
> Some vendors sell the same model at more than one service tier, trading latency against price. Where they do, Puter exposes each tier as its own model id: append `:flex` or `:priority` to the model name.

"some vendors", "flex". "priority" are too infron specific


if we want to make it more generic we could call it something like model variant
then explain how the same model can have different behavior based on the suffix
we have these:
:free, doesnt consume credit, with limits from provider
:flex, cheaper, but slower
:priority, faster but more expensive

then in the future if any of our provider has another suffix, we can just add this here
service tier i think is too infron specific
(open to suggestion)

Salazareo (approved):
code looks good to me

ProgrammerIn-wonderland (approved):

PR #3854, opened 2026-09-11
add email composer api (puter.email.send)

ProgrammerIn-wonderland:


src/puter-js/src/modules/Email/ComposerLib.js
```diff
@@ -0,0 +1,260 @@
...
+
+    const { isMultiPart, mimeType } = determineTopLevelMimeType({ attachments, text, html });
+    const boundary1 = crypto.randomUUID();
+    if (isMultiPart) {
+        headerLines += emlHeader('Content-Type', `${mimeType}; boundary="${boundary1}"`);
+    } else {
+        headerLines += emlHeader('Content-Type', `${mimeType}`);
+    }
+
+    headerLines += '\r\n';
```
Salazareo, line 213: not a complain, just question, are empty lines important for the spec?

src/puter-js/src/modules/Email/composer-lib.js
```diff
@@ -0,0 +1,260 @@
...
+
+    headerLines += '\r\n';
+
+    // --- Body Constructing Logic ---
+    // No attachments, not multipart, easy return
+    if (!isMultiPart) {
+        return headerLines + (text || html);
+    }
+
+    const textLeaf = () => emlHeader('Content-Type', 'text/plain; charset=UTF-8') + '\r\n' + text;
```
Salazareo: nit: can these not just be consts here, or does html need to be evaluated at invocations.

Salazareo (approved):
one nit but code looks good

src/puter-js/src/modules/Email/ComposerLib.js
ProgrammerIn-wonderland: empty lines are very much for the spec
same way how in HTTP header/body is seperated by \r\n\r\n
like its the email header vs email body seperation

src/puter-js/src/modules/Email/composer-lib.js
ProgrammerIn-wonderland: you're absolutely right! thats stupid

PR #3870, opened 2026-09-15
fix: charge stat's return_shares against the share-listing budget

jfcastro92:
Closes [PUT-1597](https://linear.app/puter/issue/PUT-1597/stats-return-shares-runs-share-listing-work-under-the-stat-rate-limit). Follow-up from review on #3644.

## Problem

`return_shares` on `POST /fs/stat` and the legacy `POST /stat` runs the full `ShareService.listSharesOf` — the same work `GET /share/shares` does — but was metered only under `fs:stat` (1200/min paid) instead of `share:list` (600/min). Since the scopes are separate, the two budgets stacked: a caller could do the expensive listing at up to 2× the rate the purpose-built route allows, on top of that route's own budget.

## Fix (option 1 from the ticket)

- **`consumeRouteRateLimit(req, spec)`** (new, `rateLimit.js`): an imperative charge that resolves the bucket key and per-subscription limit exactly as `rateLimitGate` does — same spec in, same bucket out — so a handler can conditionally spend a second scope when a request flag makes the route expensive. Fails open on backend error, matching the module's policy. Reusable for any route whose cost depends on a request parameter.
- **`share/limits.ts`** (new): `SHARE_LIST_LIMIT` / `SHARE_LIMIT` move out of `ShareController.ts` into a shared module (the `fs/limits.ts` pattern), so the stat handlers charge the exact spec the `/share` routes gate on.
- **Both stat handlers**: when `return_shares` is set, charge `share:list` before the listing work; 429 `too_many_requests` (the gate's shape) when the bucket is dry. Plain `stat` is untouched, and the budgets now intersect instead of stacking.
- **Docs**: no numbers moved; `rate-limits-and-quotas.md` and `FS/stat.md` now say `returnShares` spends from the share-read bucket on top of `stat`'s own.

## Test evidence

New tests:

```
✓ src/backend/controllers/fs/FSController.test.ts > FSController.statEntry > draws return_shares from the share-listing budget 95ms
✓ src/backend/controllers/fs/LegacyFSController.test.ts > LegacyFSController.stat > draws return_shares from the share-listing budget 100ms
✓ src/backend/core/http/middleware/rateLimit.test.js > consumeRouteRateLimit > spends from the same bucket the route gate does 1ms
✓ src/backend/core/http/middleware/rateLimit.test.js > consumeRouteRateLimit > buckets per user under the 'user' strategy 0ms
✓ src/backend/core/http/middleware/rateLimit.test.js > consumeRouteRateLimit > applies bySubscription overrides via the wired metering service 0ms
✓ src/backend/core/http/middleware/rateLimit.test.js > consumeRouteRateLimit > fails open when the backend throws 0ms
```

The controller tests drain the `share:list` bucket, then assert `stat` with `return_shares` 429s while a plain `stat` on the same entry still succeeds. The middleware test proves the handler charge and the route gate resolve the identical bucket for the same spec + request.

Full affected suites (fs + share controllers + rate-limit middleware):

```
Test Files  9 passed (9)
     Tests  582 passed (582)
  Duration  4.37s
```

`npm run build:ts` is clean; the only lint errors in touched files pre-exist on `main`.


reynaldichernando (approved):
docs looks good

Salazareo (approved):
code looks good, but just make sure perf isnt affected on fs operations heavily

PR #3871, opened 2026-09-15
feat: name the issuing app in share emails, and cut the digest window to 5s

jfcastro92:
Closes [PUT-1798](https://linear.app/puter/issue/PUT-1798/sharing-email-notification-changes).

## What changed

**App attribution in the mail.** When a share is issued through an app (a user-app token), the digest line now reads "alice shared report.txt **via Mail App**" — in both the holder email and the invite email, HTML and plain text. The label prefers the app's `title`, then its `name`, then its `index_url` host, resolved once per `notifyShared` call from `actor.effectiveApp`. If the app row can't be read, the mail still goes out without the attribution. Plain shares render byte-for-byte as before (`via` is only set when an app issued the share, and a sender whose digest entries arrive through two different sources loses the attribution rather than misattributing).

**`shared_app` query param on the links.** Per-item deep links and the "Open Puter" button now carry `shared_app=<app name, or uid>` so the GUI can match the share back to the app. The button link only carries it when every entry in the digest came through the same app — a mixed digest must not attribute one sender's share to another's app. The param's length is reserved up front in the link's 2000-char budget so a long item list can't squeeze it out. GUI consumption of the param is not wired here — this ships the contract (`SHARE_DEEP_LINK_APP_PARAM = 'shared_app'` in `shareDeepLink.ts`).

**Digest window 30s → 5s.** `SHARE_EMAIL_BATCH_SECONDS` drops to 5: the window only has to fold one gesture's worth of calls, and every second there is a second the common single-share email arrives late. Docs updated (they still said 90s from an earlier value).

## Test evidence

New/extended tests:

```
✓ shareEmail.test.ts > share email > names the issuing app and stamps it on the links 162ms
✓ shareEmail.test.ts > share email > links each named file to itself, by its masked path 98ms   (now also asserts a plain share carries no shared_app / via)
✓ shareDeepLink.test.ts > sharedViewLink > carries the issuing app as its own parameter, or not at all
✓ shareDeepLink.test.ts > sharedViewLink > reserves room for the app before spending the length cap on items
✓ shareNotifyTitle.test.ts > email digests > carries the issuing app onto the line, as "via"
✓ shareNotifyTitle.test.ts > email digests > drops the app when one sender arrives through two sources
```

The end-to-end test drives a real `/share` over HTTP with a user-app token (app created + granted `fs:<uid>:read`, token minted via `getUserAppToken`), with the spy on `sendRaw` so the real templates render: it asserts the "via Mail App" wording in HTML and text, `shared_app=<name>` on both the item link and the Open Puter button, and that the invite email names the app while carrying no link param.

Full affected suites (share services + email client + share controllers):

```
Test Files  10 passed (10)
     Tests  338 passed (338)
```

`npm run build:ts` and eslint are clean on the changed files.

Note: the suite's `signUpAndConfirm` now sends a unique User-Agent per signup — the new test pushed the suite to 11 signups against the 10-per-fingerprint `/signup` limit; each signup plays a distinct person, and the per-IP backstop (50) still applies.


reynaldichernando (approved):
docs looks good

Salazareo (approved):

PR #3872, opened 2026-09-15
fix: make the block list bite on team shares, and unshare every re-share (PUT-1813)

jfcastro92:
Fixes [PUT-1813](https://linear.app/puter/issue/PUT-1813/team-shares-bypass-the-recipient-block-list-and-unshareteam-only-pages). Stacked on #3846 — merge that first; this PR is the one commit on top.

## Gap 1 — team shares bypassed the recipient block list

A sender a member blocked could still land items in that member's `shared-with-me` — and grant them real access — by sharing with any team they both belong to. The group grant is one row and cannot exclude a member, so the block is now enforced everywhere delivery is **derived per member**, with the same filter:

| Surface | Where |
| --- | --- |
| Listing + total | `listByHolder` / `countByHolder` group arm |
| fs-event fan-out | `listGroupReachingMembers` |
| Notifications | already filtered (`#unblocked`) — unchanged |

Semantics, deliberately:

- **Nothing is revoked, and the authority graph is untouched.** Delivery is suspended per pair, so unblocking restores everything without a re-share. The permission scan is deliberately blind to blocks: `canManagePermission` reads the same rows, and filtering them made a block-suspended grant look *revoked* to the revoke cascade, the unshare gate, and invite claiming — turning a reversible block into permanent deletions (caught by code review; fixed in the second commit).
- **A team share is never refused for one member's sake** — no `recipient_not_accepting_shares`, and nothing tells the sharer. The rest of the team is unaffected.
- **The blanket "block everyone" switch does not gate team shares** (membership is opt-in; blanking a member's whole workspace over an anti-spam switch would break it silently). It still quiets the notifications. Blocking the sender, or leaving the team, does gate them.
- One owner for the rule: the SQL predicate lives in `UserBlockStore.notBlockedSql` and both share-store surfaces consume it. The settings' block-all note now says team shares still arrive (blocking the sender, or leaving the team, is what stops them).

## Gap 2 — `#unshareTeam` only swept the first 200 members

The sweep listed members with `limit: MEMBER_PAGE_CAP` and never followed the cursor, so `#revokeDownstream` never ran for members 201+, leaving orphaned index rows and listing noise. It now walks the share rows in the subtree (new `listIssuerIdsBySubtree`, one recursive-CTE query over active + pending rows) and filters those issuers to members (`memberIdsAmong`) — the set that can actually need sweeping, with no cap by construction. A large team where nobody re-shared costs one query instead of one subtree walk per member.

## Docs

- `FS/share.md` and `Teams.md`: the blocked-member exception stated where team sharing is documented.
- Dashboard `blocked_senders_note` wording updated to match the new behavior.

## Tests

7 new cases in `TeamShare.test.ts`: blocked member gets no listing entry, no count movement, no grant, no fan-out row while the rest of the team does; block-then-unblock suspends and restores; only the blocked pair is suspended; member re-shares and unclaimed invites are swept on team unshare; subtree issuer discovery + member filtering. One fixture fix: `makeNestedFile` now sets `parent_id`, since the subtree CTE walks parent linkage like production rows.

```
Test Files  1 passed (1)
     Tests  35 passed (35)
```

Full backend suite:

```
Test Files  319 passed | 24 skipped (343)
     Tests  8325 passed | 26 skipped (8351)
  Duration  176.06s
```

`npm run typecheck` passes (no new errors), eslint clean.

Not reproduced live: the literal 201st-member case (needs 200+ accounts). The new path has no membership cap by construction, which the store-level test pins.

## Review fixes (`e6b2601a8`)

`/code-review` on the first commit confirmed the block filter inside `readUserGroupPerms` leaked into authority checks, with three permanent-loss scenarios (cascade-deleted re-shares, un-withdrawable re-shares, hard-deleted invites). Second commit:

- **Scan unfiltered again** — blocks act only on delivery surfaces; the grant and everything resting on it survive the block. New test pins that a blocked member can still withdraw what they re-shared.
- **Team-held re-shares swept**: `#revokeDownstream` now also revokes group-held rows the revoked issuer made in the subtree (user path included) — previously a dormant grant that sprang back if the issuer requalified. Tested via a member re-sharing team A's file into team B.
- **Sweep skipped while the team is still granted by another issuer** — members' re-shares rest on that surviving authority; revoking them would orphan live access. Tested with two issuers on one team.
- Consolidated the block predicate into `UserBlockStore.notBlockedSql`, deduped the recursive subtree CTE, hoisted the `seen` set across the sweep, fixed the false `blocked_all_note` copy, documented the block-all/team exception in `share.md`, dropped ticket tags from test comments.

Known limitations, deliberate: a *departed* member's re-shares still escape the team-unshare sweep (no departure path exists yet that would re-walk issued grants), and a block doesn't push `item.removed` to the blocker's open windows (refresh reconciles). Both are pre-existing classes flagged by the review, worth their own change if we want them.

Re-verified: TeamShare suite 38/38, full backend 8330 passed / 0 failed, typecheck and lint clean.

## Review fixes (`27e858d76`)

A second `/code-review` pass flagged two more items, fixed here:

- **`deleteActiveGroup` gated on authority**, matching the user path: it ran even when every `canManagePermission` check was denied, deleting the share index row while the team's permission rows survived — live access nothing lists, the exact class this PR removes. A denied node now keeps its index row.
- **`memberIdsAmong` qualifies `deleted_at`**: `#live()` now takes a table alias, so the join stops relying on `jct_user_group` never growing a `deleted_at` column.

Two more review findings recorded as known limitations rather than fixed here: the group-path revocation in `#revokeGroupSharesBy` does not recurse into members of the *receiving* team the way the user path recurses through holders (one extra level of indirection escapes the sweep), and the block-list copy change only landed in `en.js` — other locales need their usual i18n follow-up.

Base retargeted to `main` (#3846 merged), so the diff now shows only this PR's commits. Re-verified after the fixes: TeamShare + TeamStore suites 78/78, typecheck and lint clean.


src/docs/src/Teams.md
reynaldichernando, line 1: for this overview page, please follow the structure of

- intro
- features
- functions
- examples

reference: https://docs.puter.com/AI/

the aim of this page is to give developers a top level overview of what the API is, details should ideally be kept in each methods page

reynaldichernando (changes requested):

Salazareo (changes requested):
team shares bypassed the recipient block list

I think this is intentional, like I don't want my own team to block my shares

src/docs/src/Teams.md
jfcastro92: Restructured to intro → features → functions → examples, following the AI page.

Details moved off the overview and onto the method pages, so each is self-contained: the pagination forms, the object shapes (`Team`, `TeamDirectoryEntry`) and the error tables now live in [`list`](/Teams/list/) and [`listDirectory`](/Teams/listDirectory/) rather than being linked back to anchors on the overview. Those cross-links are gone — the overview no longer has the `#pagination` / `#team` anchors they pointed at.

What stayed on the overview is top-level only: what a team is, the four capabilities (team context, colleague lookup, sharing with a team, stable identifiers), the two functions, and three runnable examples. I also kept the "deliberately absent" note, because a team *not* being able to restrict how its members share is the assumption most people bring the other way round — happy to move or cut it if you'd rather the overview carry no caveats.

`node build.js` is clean (the four warnings it prints are pre-existing KV/Auth playground links).

jfcastro92:
Both reviews addressed, and `main` merged in — the PR is mergeable again.

## @Salazareo — you're right, and I've reversed it

> team shares bypassed the recipient block list — I think this is intentional, like I don't want my own team to block my shares

Agreed, and that's also the second of the two directions your own ticket offered ("or state the limitation in the block API"). Dropped the enforcement in `3f33593`:

- `listByHolder`'s group arm, `countByHolder` and `listGroupReachingMembers` no longer filter team rows by the recipient's block list, and the shared SQL fragment that did it is gone.
- The tests now pin the intentional behaviour instead: a member who blocked the sharer **still** gets the listing row, **still** has the grant behind it, and **stays in** the live-event fan-out.

One thing left deliberately as it was, which I'd like your read on: **the notification is still suppressed** for a blocked sender, including for team shares. That was true before this branch — `ShareNotificationService` has always filtered the telling. So a block now stops the interruption without pretending to stop the access, which I think is the defensible line, and it's what the docs and settings copy now say. If you'd rather a block not affect team-share notifications either, that's a one-line change to `#unblocked` and I'll make it.

**The unshare half of the ticket stands** — that was never contested: `#unshareTeam` still swept by a capped 200-member page, so re-shares by members past the cap were never withdrawn.

## Merge resolution

Two conflicts, both places where `main` grew a sibling of something here. Worth flagging the second, since I took main's logic over my own:

- `listAllByFsentrySubtree` (main) and `listGroupSharesBySubtree` (here) answer different questions — both kept, now sharing this branch's subtree-CTE helper instead of repeating the recursive query.
- `#unshareTeam`'s sweep now carries **all three** guards. Main's `#managedFromOutside` is the more precise per-member test (it spares a member whose `manage` comes from a personal grant, another team, or a folder above), so it's kept — but it was still walking the capped member page, and it can't see a grant to *this* team from an issuer the call isn't withdrawing. So main's check now runs inside this branch's row-driven candidate list, behind its last-grant guard.

Verified after the merge: share, share-store, user-block and team suites — **467 passed**; `npm run typecheck` clean; docs build clean.

Salazareo (approved):

reynaldichernando (approved):

PR #3887, opened 2026-09-16
fix(ai): harden chat/video model maps and txt2vid argument guards

404oops:
Part of the AI driver maintenance work, split out for its own review. Two changes:

**Chat and video drivers**

The provider/model lookup maps were plain objects, so a `model` or `provider` value like `__proto__` or `constructor` resolved against `Object.prototype` and blew up as a 500. Both maps are now null-prototype objects, and those names are treated like any other unknown name: chat returns 400 `Model not found`, video falls back to the configured default provider. Tests cover both drivers.

**`puter.ai.txt2vid()`**

- Rejects blank or non-string prompts up front with `prompt_required` instead of a raw `TypeError`
- No longer throws on `null` in either argument slot
- Copies the caller's options before resolving the `duration` alias and output path, so frozen options work and the caller's object is never mutated

Both call forms are covered by suite tests; `txt2vid.md` and the `Txt2VidOptions.prompt` JSDoc document the behavior.

## Backward compatibility

Call signatures, option names, response fields, and error codes are unchanged. One observable nuance: callers that read `options.seconds` or `options.puter_output_path` off their own object after a `txt2vid` call will no longer see those writes, since the options are now copied first.

## Verification

- `npx vitest run --config src/backend/vitest.config.ts src/backend/drivers/ai-chat/ChatCompletionDriver.test.ts src/backend/drivers/ai-video/VideoGenerationDriver.test.ts`
- `npx vitest run src/puter-js/src/modules/ai/ai.test.js` (61 passed)
- `npm run check:puterjs:types` and eslint clean


reynaldichernando (approved):
docs looks good

Salazareo (approved):

PR #3889, opened 2026-09-16
fix(ai-image): refresh model catalogs, unify txt2img requests, expand docs

404oops:
Image-generation maintenance for the standard AI checklist: refresh the model catalogs, normalize how `puter.ai.txt2img()` requests are built, and expand the docs. Independent of #3887, which carries the chat/video pieces split off this tree — the two touch disjoint areas.

### Model catalogs

Every provider's model list was checked against what the vendor actually offers (re-probed on 2026-09-16), and for paid models a live generation sweep across the catalog (114 of 117 succeeded; 3 recorded failures are surfaced when requested):

- **Gemini**: use the current stable ids (`gemini-3-pro-image`, `gemini-3.1-flash-image`); retired `-preview` spellings kept as aliases; `gemini-3.1-flash-image` is now the default. `gemini-2.5-flash-image` stays available until Google's 2026-10-02 shutdown. Token tables and prices updated.
- **OpenAI**: `gpt-image-1` and related ids are deprecated upstream, so they're hidden from listing but still usable until their shutdown dates; `gpt-image-2` is the new default.
- **xAI**: adds `grok-imagine-image-2.0` with its `low`/`medium` quality tiers and separate `resolution` option; edit requests now accept up to 5 reference images (previously truncated silently).
- **BytePlus**: adds the 3K/4K tiers, per-model size limits, and resolution tables per aspect ratio.
- **Cloudflare**: adds SDXL Lightning, SDXL Base 1.0 and SD 1.5 Inpainting (with mask support); per-model step/size bounds; FLUX.2 uploads send real bytes.
- **Replicate**: adds 88 models driven by their API schemas, version-pinned predictions, per-tier/second billing with price sources, and retired ids rejected by name.
- **Together**: all image routes are disabled (`thirdPartySharing` data policy) because Together rejects all image models unless the org opts into third-party data sharing; requests fail with `bad_request` before any upstream call and the models are hidden from discovery. Retired ids are blocked so they can't silently route to Cloudflare.

### Request normalization

What you pass is now what each provider interprets by the same rules: the driver turns `ratio` / `width`+`height` / `aspect_ratio` into one `imageSize`, validates `prompt`, `quality` and `resolution` once, accepts short provider hints (`'openai'`, `'gemini'`, …) as well as full driver ids, and hands every provider an immutable copy of your options. Shared helpers standardize prompt checks, aspect snapping, and content-sniffed data URIs so the returned image MIME type is correct on every provider. Replicate predictions are created once, cancelled on abort or deadline, and input fan-out is bounded by count, bytes, concurrency, and time.

About `normalize`: the image response shape is already uniform across browsers and Node (a single image element, or `{ src }`), so there is no vendor-specific shape to convert. The SDK doesn't send `normalize` for `txt2img()`, and the docs say so.

### SDK (backward compatible)

`txt2img()` keeps all three call forms. It now copies your options instead of mutating them (frozen objects work), and rejects missing, blank, or non-string prompts with `prompt_required` in every form. No option, response field, or error code was renamed.

### Docs

`txt2img.md` rewritten per provider with defaults, discovery, availability notes, and output-size rules; new catalog and pricing page at `AI/image-models.md`; image-generation limits added to `rate-limits-and-quotas.md`.

## Security notes

No auth or permission changes. Caller-supplied image URLs stay behind `secureFetch`; Replicate fetches are capped at 30 MB, 30 s, and 10 references. Error messages name the model and provider only.

## Verification

- `npx vitest run --config src/backend/vitest.config.ts src/backend/drivers/ai-image src/backend/drivers/util/imageInput.test.ts` (597 passed, 3 skipped)
- `npx vitest run src/puter-js/src/modules/ai/ai.test.js` (62 passed)
- `npm run build:workerLib && npm run test:puterjs:node` (507 passed, 67 skipped)
- `npm run typecheck` clean; `npm run check:puterjs:types` clean; eslint clean on touched sources
- Vendor listings re-probed 2026-09-16 with configured keys. Deliberate omissions: `chatgpt-image-latest`, Cloudflare `dreamshaper-8-lcm`, BytePlus `seedream-4-0-20260415` (no published price).


src/docs/src/AI/image-models.md
reynaldichernando, line 1: we have an always up to date SSOT of puter AI models here https://developer.puter.com/ai/models/
(which includes the pricing)
please link to our model directory instead of creating a new static docs

reynaldichernando (changes requested):

reynaldichernando (approved):
lgtm

ProgrammerIn-wonderland (approved):

PR #3890, opened 2026-09-16
feat: tell the plan picker when a seat's plan is on the way out

jfcastro92:
Console half of the PUT-1819 follow-up (see heyputer PR "name each seat's subscription status on the team billing view"): the Teams tab reads the new `seatStatuses` off the team billing view and passes the seat's status into the `team-plan-purchase` event, so the picker can offer keeping a cancel-pending plan. Adds the two strings that flow renders (`teams_keep_plan`, `teams_plan_ending`).

Degrades cleanly against an older API: no `seatStatuses` in the payload means no status is passed and the picker behaves exactly as today.


jfcastro92:
## Live e2e on a local instance (real Stripe test mode)

Ran this branch as `packages/puter` in a local heyputer instance, with [heyputer#1229](https://github.com/HeyPuter/heyputer/pull/1229) (the `seatStatuses` field) and [gui#196](https://github.com/HeyPuter/gui/pull/196) (the picker button), Playwright driving the real GUI and a real Stripe test-mode subscription behind it.

The console half is what carries the status from the wire to the picker, so the evidence that it works is the picker rendering the state:

```
current-plan card hint  : Ends with the current period
current-plan card button: Keep Team Basic
```

…which only appears because `loadPlan` picked up `seatStatuses` and `changeSeatPlan` passed `currentStatus` through. Read back out of the live DOM.

| Teams tab, seat on Team Basic | Picker for a cancel-pending seat |
| --- | --- |
| 
01-teams-tab-paid
 | 
05-picker-cancel-pending-KEEP-BUTTON
 |

Sequence: buy Team Basic → Switch to Free → seat reads `cancel_pending` on the wire → reopen picker → **Keep Team Basic** → `active` again, same Stripe subscription (`cancel_at_period_end` cleared, no duplicate).

Also confirms the degradation path works in the other direction: the instance ran the old backend first, `seatStatuses` was absent, `currentStatus` came through `null`, and the picker behaved exactly as it does today — no button, no error.

`npm run typecheck` clean; the Teams suites pass unchanged.

Salazareo (approved):

PR #3892, opened 2026-09-16
fix: stop disclosing invite addresses to apps, tokens and delegates

jfcastro92:
Closes [PUT-1806](https://linear.app/puter/issue/PUT-1806/invite-email-addresses-are-disclosed-to-apps-manage-delegates-and).

## The disclosure

`clientShare.ts` states the wire invariant — "only ever the username … never an email the caller didn't already supply" — but nothing enforced it. `listSharesOf` gated on `#assertCanManage`, whose `mode` defaults to `'see'`, so **any credential that could see the node** got every unclaimed invite's raw email address, on that node *and its ancestors*. All three routes in the ticket reproduce:

| Route in | Credential | Got |
| --- | --- | --- |
| `GET /share/shares?uid=F` | app granted `fs:F:read` by the picker | every invitee address on F and above |
| `POST /fs/stat {return_shares}` | `list`-scoped access token (`allowAccessToken: true`) | same listing |
| `getShares()` | `manage` delegate | the owner's invitee addresses |

## The fix

Both bounds from the ticket's direction:

- **An invite's address goes only to the item's owner and to whoever sent it**, and never to an app or token acting for either. The boundary matches the one `unshare()` already enforces — "anyone else can withdraw the shares *they* granted" — so a delegate loses nothing they could act on. The row is still listed (`pending: true`, `holder: null`); only `recipientEmail` is withheld, and `toClientShare` now omits the key rather than emitting `undefined`.
- **An app or token must hold `manage` reach of its own** to read the listing at all. `getShares()` is documented as "who can reach a file you own, or one you have `manage` access to"; the credential check now matches that. An app that genuinely holds `manage` still works (proven below), and `tryListSharesOf` turns the refusal into an empty `shares` array, so `stat()` itself keeps working for a scoped token.

Refusals reuse the ACL's own safe error, so a credential still can't distinguish "exists but you lack manage" from "doesn't exist". For a plain session `#hasOwnReach` short-circuits, so the common path costs nothing.

GUI: an invite whose address is withheld would have rendered a blank row with a revoke button the server refuses, so it now names the row (`share_invited_someone`) and drops the button.

## Test evidence

Service level:

```
✓ who may read an invite address > tells the owner who they invited
✓ who may read an invite address > withholds it from a manage delegate who did not send it
✓ who may read an invite address > tells a delegate the address of an invite they sent
✓ who may read an invite address > refuses the listing to an app holding only the file
✓ who may read an invite address > still answers an app that holds manage on the item
```

Over real HTTP, against the routes named in the ticket — an app token on `/share/shares`, and a genuinely minted `fs:<uid>:list` access token on `/fs/stat`:

```
✓ an invited address over the wire > reaches the owner but not an app the file was handed to
✓ an invited address over the wire > refuses a list-scoped access token on the `return_shares` surface
✓ an invited address over the wire > answers an app that was handed the item to manage
```

**These fail against the unfixed code** — I stashed the two source files and re-ran: 4/6 service tests and 3/3 HTTP tests go red, the two that stay green being the owner-still-sees-it controls.

Full backend suite: `8536 passed | 26 skipped`, 0 failed. Typecheck and eslint clean. The one GUI suite failure (`pdfThumbnails/browser.test.js`, missing local `pdfjs-dist`) reproduces with my changes stashed and is unrelated.

## Notes

- `listSharedByMe` is untouched: `listOutbound` scopes it to the caller's own issued rows, so the caller is the issuer of every address it returns.
- Docs updated in `FS/getShares.md` and `FS/stat.md`.
- Only `en.js` has the new i18n key, per the usual separate-locale follow-up.

## Review fixes (`461cb7d0a`)

`/code-review` found two gaps in the first commit, both fixed here:

- **A full-access token was denied the address, while `GET /share/shared-by-me` still handed it the same rows.** The clause bought no privacy against that credential class and broke the contract this PR documents, costing an API client the address `unshare(path, email)` takes. The boundary is now `isAccountContext` — the concept the rest of the codebase already uses for "holds the account's own reach": a plain session or a full-access token, never a scoped one. Scoped tokens, the ticket's actual attack, stay refused. Pinned by a test that mints a real full-access token and asserts both routes agree.
- **The Dashboard share modal dropped a withheld invite entirely.** `shareAggregate.identify()` keys a pending row on `recipientEmail` and discards a row with no name, so after the first commit a `manage` delegate saw no sign an invite was outstanding and `accessCount` under-reported who reaches the item. The invite is now kept, keyed on the share uid (so two withheld invites don't collapse into one row), labelled, and rendered without the mode/revoke controls — every one of those addresses a person by name, which is exactly what was withheld. `UIWindowShare.js` got this in the first commit; this is the second dialog.

Both regressions are covered by tests that fail against the previous commit.

**Release note, deliberate:** an app holding `write` on a user-granted file can still create shares on it but can no longer read them back through `getShares()` / `stat(returnShares)` — `POST /share` is gated on the mode being granted, the listing now on `manage` reach. Its remaining path is the app-scoped `listSharedByMe`, which has a different (flat, paginated) shape. That is the intended tightening, but it is an unversioned break worth calling out in the release notes.

Re-verified: full backend suite `8537 passed | 26 skipped`, 0 failed; GUI `664 passed` (the one failing file, `pdfThumbnails/browser.test.js`, is a missing local `pdfjs-dist` and fails identically on a clean tree); typecheck and eslint clean.


reynaldichernando (approved):
docs looks good

reynaldichernando (approved):
docs looks good

Salazareo (approved):

jfcastro92:
Merged `main` in — mergeable again. Worth reading before re-reviewing, because **`main` grew its own answer to half of what this PR was for, and I deferred to it.**

## The two textual conflicts

Both additive on each side, both kept: `isAccountContext` (this branch) and `isPlainUserActor` (main) are each used, and the `stat()` note keeps both sentences — this branch's on who may read an invite address, main's on the share-read limit `returnShares` spends.

## The part that isn't just a merge

`listSharesOf` on `main` now bounds an app to **the rows it issued itself**. This branch had taken a different route: require the *credential* to hold `manage`, and refuse it otherwise.

Main's is the better mechanism, so the gate is gone:

- it **answers** the app instead of turning it away, and
- it hides other issuers' rows **outright** rather than redacting one field on them.

Two tests that asserted my refusal went with it. In their place, two that hold main's line — an app sees none of the invites it did not send, **with or without `manage`**:

```
✓ shows an app none of the invites it did not send
✓ shows an app with manage no more than one without
```

## What this PR still fixes

Main's row filter only applies when an app is acting (`issuedHere` is a no-op otherwise), so a plain **manage delegate still reads the owner's invite addresses**. That is the gap `#maySeeInviteAddress` closes, and its tests pass unchanged:

```
✓ tells the owner who they invited
✓ withholds it from a manage delegate who did not send it
✓ tells a delegate the address of an invite they sent
```

I also corrected the `stat()` doc line, which described the gate I removed rather than main's behaviour.

Verified after the merge: share, share-store and fs suites — **623 passed**; `npm run typecheck` clean; docs build clean.

jfcastro92:
CI was right, and it caught a real hole rather than just stale assertions. Fixed in `72203152`.

## What failed

Three HTTP-level tests in `ShareController.http.test.ts` — the same assertions I had already updated at the service level, in a file I did not run locally. My miss.

## But one of them was pointing at something real

Dropping this branch's `manage` gate in favour of main's row filter lost a case **main's filter does not cover**. Main bounds an *app* to the rows it issued; a **scoped access token is not an app**, so it fell through unbounded — `/fs/stat` with `return_shares` would hand a `fs:<uid>:list` token the owner's entire share list. Addresses stayed withheld, but who else can reach a file is no more a narrow token's business than the addresses are.

So rather than restore my gate, I generalised main's mechanism:

| Credential | Sees |
| --- | --- |
| Session, or full-access token | everything — it holds the account's reach |
| App | the rows it issued (unchanged from main) |
| **Scoped token** | **nothing — it issues nothing under its own name** |

Filtering a token by a null app would have been *worse* than not filtering, since a null `issuedByApp` matches the owner's own rows. Hence the explicit empty.

## One bug I introduced and caught on the way

My first version keyed on `isAccountContext()`, which requires `effectiveApp === null`. Test actors built as `{ user }` leave it `undefined`, so **owners** were being classed as bounded credentials and the share dialog listed nothing — four more failures, in `TeamShare` and `LegacyFSController`. Narrowed to an explicit app-or-scoped-token check, which does not depend on `effectiveApp` being set. Worth knowing because any production path constructing an actor without `effectiveApp: null` would have hit the same thing.

## Verification

Full backend suite this time, not the slice I checked before:

```
Test Files  327 passed | 24 skipped (351)
     Tests  8637 passed | 26 skipped (8663)
```

`npm run typecheck` clean.

(The three test names also got honest: one had always promised a refusal its body never actually asserted.)

PR #3893, opened 2026-09-17
Add auth message for external apps using puterjs

reynaldichernando:
- add the message "`<website>` is powered by Puter for its AI and cloud features." on auth screens
- minor styling and mobile responsiveness fix

this is intended to give context to users coming from a website using puterjs, on why they have to sign up/login to puter
sign up
image
sign in
image

only affects 3rd party website, direct access from puter.com does not show this message
image

this is a part of improving adoption for puterjs
devs have always had an issue of integrating puterjs because the need of a 2nd login
and this is aimed to improve that
also, just for any website where users see this auth screen

(answering why do i have to login to another platform)


Salazareo (approved):

reynaldichernando:
merging this since its tested on staging

PR #3916, opened 2026-09-21
fix: block ordinary apps from launching godmode apps via the SDK

jfcastro92:


jfcastro92:
Closed to be restructured in another PR

Issue #1748, opened 2025-10-15
Windows Error


vermashobhit1994:
running below command
>npm start

how to solve it?

> puter.com@2.5.1 start
> npm run build:ts && node ./tools/run-selfhosted.js

> puter.com@2.5.1 build:ts
> tsc

[BOOT/INFO] configuration `F:\\artificial-intelligence\\mlproject\\agentic_AI\\my-node-server\\puter\\volatile\\config`
[BOOT/INFO] working directory `F:\\artificial-intelligence\\mlproject\\agentic_AI\\my-node-server\\puter\\volatile\\runtime`
WORKERS ERROR: Preamble has not been built! Workers will not have access to puter.js
To fix this cd into src/backend/src/worker and run npm run build
node:internal/child_process:420
    throw new ErrnoException(err, 'spawn');
          ^

Error: spawn EINVAL
    at ChildProcess.spawn (node:internal/child_process:420:11)
    at spawn (node:child_process:786:9)
    at Kernel.run_npm_install (F:\artificial-intelligence\mlproject\agentic_AI\my-node-server\puter\src\backend\src\Kernel.js:581:22)
    at Kernel.install_extern_mod_ (F:\artificial-intelligence\mlproject\agentic_AI\my-node-server\puter\src\backend\src\Kernel.js:422:20)
    at async Promise.all (index 3)
    at async Kernel.install_extern_mods_ (F:\artificial-intelligence\mlproject\agentic_AI\my-node-server\puter\src\backend\src\Kernel.js:285:30)
    at async Kernel._install_modules (F:\artificial-intelligence\mlproject\agentic_AI\my-node-server\puter\src\backend\src\Kernel.js:168:9)
    at async F:\artificial-intelligence\mlproject\agentic_AI\my-node-server\puter\src\backend\src\Kernel.js:139:13
    at async F:\artificial-intelligence\mlproject\agentic_AI\my-node-server\puter\src\backend\src\util\context.js:197:20
    at async Context.arun (F:\artificial-intelligence\mlproject\agentic_AI\my-node-server\puter\src\backend\src\util\context.js:195:16) {
  errno: -4071,
  code: 'EINVAL',
  syscall: 'spawn'
}

Node.js v22.18.0

ProgrammerIn-wonderland:
[...]

Issue #294, opened 2024-04-17
Unable to Run on Windows 11 with Node 21


rajivraghu:
```
PS C:\POC\PuterOs\puter> npm install
npm WARN deprecated phin@2.9.3: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
npm WARN deprecated xterm-addon-fit@0.7.0: This package is now deprecated. Move to @xterm/addon-fit instead.
npm WARN deprecated querystring@0.2.0: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
npm WARN deprecated sinon@15.2.0: 16.1.1
npm WARN deprecated xterm@5.3.0: This package is now deprecated. Move to @xterm/xterm instead.

> puter.com@2.1.0 postinstall
> cd packages/phoenix && cd packages/contextlink && npm install && cd - && cd packages/strataparse && npm install && cd - && cd packages/pty && npm install


added 77 packages, changed 1 package, and audited 82 packages in 2s

21 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
The system cannot find the path specified.
npm ERR! code 1
npm ERR! path C:\POC\PuterOs\puter
npm ERR! command failed
npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c cd packages/phoenix && cd packages/contextlink && npm install && cd - && cd packages/strataparse && npm install && cd - && cd packages/pty && npm install

```
After this npm start , This will  not start properly..

![image](https://github.com/HeyPuter/puter/assets/5885193/07232ae9-2733-499f-a446-01facd8f076c)

Blank UI

half-real-SCRACX:
I have the same error! I haven't figured the solution out yet, but I'll try looking into it and posting it here.

half-real-SCRACX:
This is related to the problem in #208, which has a pending pull request (#215) to fix the issue.
But I don't think that this also fixes
```
npm ERR! code 1
npm ERR! path path\to\puter\repo
npm ERR! command failed
npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c cd packages/phoenix && cd packages/contextlink && npm install && cd - && cd packages/strataparse && npm install && cd - && cd packages/pty && npm install
```

KernelDeimos:
[...]

Issue #3910, opened 2026-09-21
apps repeatedly ask permission when it wants to access files


reynaldichernando:
hi team i keep getting this popup when i want to access my file (in this case, word processor) even though i've given it access multiple times

ideally it should just remember that i already given it access

<img width="429" height="444" alt="Image" src="https://github.com/user-attachments/assets/3ecaaaf8-0cdc-42fc-b2b4-310c8fb8a3be" />

Salazareo:
@reynaldichernando whats your browser?

reynaldichernando:
i’m using chrome

Issue #3908, opened 2026-09-20
new addition for puter.js


thecodegangster:
with the direction that puter.js is going i believe that it should contain an ad  (maybe payment too) api that would allow puter devs to monetize their apps

jelveh:
Hey @thecodegangster. We always wanted to have something like that in Puter.js, i think it's a bit too early though. For the time being I'll close this but we will announce it once we ship something in that direction.

PR #3923, opened 2026-09-22
fix(tools): spawn npm.cmd through a shell on Windows

### LINKED ISSUE #3911: npm start fails on Windows with spawn EINVAL — regression of #1748 after the tools/ restructure
npm start fails immediately on a fresh clone on Windows 11, so the backend never starts.

Env. -  Windows 11 (10.0.26200), Node v24.19.0, `main` at 24fb679dd.

## This is what I found - 

In tools/start.mjs, npm is resolved to `npm.cmd` on Windows and then spawned without a shell:

const npm = process.platform === 'win32' ? 'npm.cmd' : 'npm';

This causes `spawn EINVAL` on current Node versions when trying to execute the `.cmd` file.

I also found a similar `npm.cmd` spawn pattern in `tools/extensionSetup.mjs`.

I tested a local fix for both call sites and was able to get Puter running natively on Windows.

### Steps to reproduce

git clone https://github.com/HeyPuter/puter.git
cd puter
npm install
npm start

### Actual result

> puter.com@26.07 start
> node ./tools/start.mjs

spawn EINVAL

rasadregmi:
`npm start` fails on Windows 11 with `spawn EINVAL` on current Node versions, so the backend never boots. `tools/start.mjs` and `tools/extensionSetup.mjs` resolve `npm` to the `npm.cmd` batch shim on win32 and spawn it without a shell; Node >= 18.20 refuses to execute `.cmd` shims without one (CVE-2024-27980 hardening) and throws `EINVAL` instead of running it.

This is the third appearance of the same bug #294 and #1748 both reported it, #2269 fixed it, and the win32 handling was lost when the 26.07 restructure replaced `Kernel.js` / `DevWatcherService.js` with `tools/`.

Adds `tools/npmSpawn.mjs` with platform-aware `npmCommand()` and `npmSpawnOptions()` (merging `shell: true` on win32 only) and uses it in both scripts, so the shell option only ever reaches Windows spawns. Covers it with a regression test registered in the backend vitest suite. POSIX behavior is unchanged.

Testing:
- `npx vitest run --config src/backend/vitest.config.ts tools/npmSpawn.test.mjs` 6/6 passed; the same run fails when the shell option is removed, confirming the test catches the regression.
- `npx vitest run --config src/backend/vitest.config.ts tools/npmSpawn.test.mjs extensions/devWatcher.test.ts` — 22/22 passed.
- `npm run typecheck` — no new errors.
- `node ./tools/extensionSetup.mjs` — exit 0.
- `npm start` (Linux) — reaches "PuterServer has fully booted."

Closes #3911


rasadregmi:
**CI failure is an upstream flake, not caused by this change, please re-run the failed job.**

The only failing tests are in `src/backend/services/share/ShareService.trash.test.ts` (ShareService trash + a couple of ShareService trash tests), failing inside `FSService.move` -> #assertNotUserRoot with `Cannot move a home directory` (a guard added upstream in `HeyPuter/puter#3920`). My change only touches the npm-spawn helpers in `tools/`, a mocked-spawn unit test, and a vitest include glob — there is no shared code path with ShareService.trash tests, and those tests are untouched by this PR.

For reference, HeyPuter main (commit 2690f4934, which this branch is based on) and HeyPuter's own merged PRs #3913/#3914/#3915 all passed this exact backend suite, so the suite is green upstream. The failure here stacks through the ShareService trash suite, which upstream has flagged as flaky (#3908, #3910, #3922 touched the same area).

Please re-run the failed `test (pr)` / coverage job when you get a chance, it is almost certainly a timing-dependent trout in the ShareService suite rather than anything in this PR.

PR #3927, opened 2026-09-22
fix(ai): classify upstream credit exhaustion, contain the route, redact vendor text (PUT-1870)

404oops:
## What

- classify HTTP 402 and provider credit/quota/balance signals as `upstream_credits_exhausted` before rate-limit and auth handling
- mark exhausted chat routes unhealthy so later requests skip them during the health TTL
- strip HTML, vendor URLs, and request identifiers from attempt details, HTTP errors, logs, alarms, and stream error frames
- force `upstream_credits_exhausted` alarms at warning severity
- mirror the classification and sanitization in the `/drivers/call` fallback translator

## Why

Drained vendor accounts were being reported as bad credentials, bad requests, or silent free-model rate limits. Those classifications retried dead routes, hid an actionable billing condition from on-call, and exposed raw vendor text to callers.

Existing callers are unaffected except that upstream credit exhaustion now has the stable `upstream_credits_exhausted` code and generic 503 message. Existing attempt fields remain available, with their error text sanitized.

Linear: https://linear.app/puter/issue/PUT-1870

## Security and privacy

Provider URLs, request ids, and HTML are removed before attempt details can reach response fields, log lines, alarm details, or stream frames. The controller translation path also avoids retaining the raw vendor exception on the new credit error.

## Tests

- `npm run typecheck` — passed; no new errors
- requested focused Vitest scope — 76 files passed, 1,670 tests passed, 20 skipped
- targeted changed suites — 5 files passed, 109 tests passed
- AI video/image regression suites — 17 files passed, 749 tests passed, 4 skipped
- production TypeScript files checked with ESLint — no errors
- `npm run test:backend` — 332 files passed; 9,104 tests passed, 26 skipped; 6 failures are an existing `origin/main` regression in `ShareService.test.ts` and `ShareService.trash.test.ts`, reproducible in isolation, where the new home-directory guard rejects those fixtures' rename/move operations

## Follow-up

`VoiceChangerDriver.ts` and `XAISpeechToTextProvider.ts` retain their independent 401/403-to-auth mappings and should adopt the shared predicate separately.


Salazareo (approved):
@ProgrammerIn-wonderland plz take a look too. thnx

ProgrammerIn-wonderland (approved):

